PR #19, opened 2020-06-05
Update README.md

hannesmuehleisen:
we now have maven


mrigger:
Thanks for fixing the outdated README!

Issue #3, opened 2020-05-26
Shortly run SQLancer on the individual DBMS as part of the CI


mrigger:
It would be desirable to add smoke testing to the CI, to verify that the testing implementation for one of the DBMS is not obviously broken.

mrigger:
This seems to be more tricky than initially thought. For DuckDB and SQLite, running SQLancer might cause the JVM to crash when a memory error is triggered (see https://github.com/sqlancer/sqlancer/issues/1). Also for the other DBMS, it might be that bugs are still triggered on their latest version, either since a new bug was introduced or because an old one has not yet been fixed. An option would be to disable checking whether SQLancer finds a bug, but then we would also not detect any regression bugs in SQLancer.

I think the optimal setup would be if the DBMS under test would, for each commit to it, pull the latest (release) version of SQLancer and run it for a while (e.g., 10 minutes), to prevent that any new bugs are introduced to the DBMS. SQLancer could, in turn, test the latest (release) version of the DBMS to check that SQLancer is not obviously broken.

mrigger:
Closing this, since SQLancer now runs tests for the DBMS it supports.

PR #45, opened 2020-06-23
Test DuckDB as part of the SQLancer CI

mrigger:
Partly addresses https://github.com/sqlancer/sqlancer/issues/3. DuckDB will also run SQLancer as part of the CI, see https://github.com/cwida/duckdb/pull/708.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ The following commands clone SQLancer, create a JAR, and start SQLancer to fuzz 
 ```
 git clone https://github.com/sqlancer/sqlancer
 cd sqlancer
-mvn package
+mvn package -DskipTests
 cd target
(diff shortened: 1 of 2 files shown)


mrigger:
It seems that the JVM crashed (see https://travis-ci.com/github/sqlancer/sqlancer/builds/172790285):
```
[2020/06/24 07:22:25] Executed 201454 queries (313 queries/s; 3.00/s dbs, successful statements: 48%). Threads shut down: 0.
java: ../nptl/pthread_mutex_lock.c:81: __pthread_mutex_lock: Assertion `mutex->__data.__owner == 0' failed.
Aborted (core dumped)
```
@hannesmuehleisen I assume that this is due to DuckDB or its JDBC driver. What do you think?

hannesmuehleisen:
Yes that crash is similar to the ones we have seen. On it.

PR #101, opened 2020-07-13
Fix the thread exiting logic

mrigger:


mrigger:
@qoega, it seems that the ClickHouse CI currently fails on this PR, as well as on #102 and #103. Is it okay to remove the ClickHouse Travis CI tests until you have time to have a look into this, which is when we could re-add them?

qoega:
Ok.
I guess I know the reason and it can be fixed by filtering out couple of Table engines. I think I'll look on TableGenerator, as it still too simple.

mrigger:
Sounds good, thanks @qoega!

Another option might be to add a class like https://github.com/sqlancer/sqlancer/blob/master/src/sqlancer/mysql/MySQLBugs.java, which helps to add "expected errors messages" for unfixed bugs. Once the bug is fixed, the conditionally included error messages can be easily identified and removed.

PR #104, opened 2020-07-13
standardizing logs & current logs across test oracles & within the PostgreSQL implementation

nukoyluoglu:
Proposed changes:

General:
- Updating the `state.queryString` variable inside ComparatorHelper when the test fails for the other TLP Oracles (in the current scenario, the `state.queryString` is still set to the left-over value from the last TLP Aggregate Oracle testing, so the query printed at the bottom of the log file with preceding "--" is the wrong query)
- Not adding the query string to `state.statements` in `ComparatorHelper` (since `state.queryString` will now be set to the query string, adding the query string to `state.statements` causes the query to be printed twice at the bottom of the log file)
- Standardizing the format of the `state.queryString` and AssertionError messages being printed in `ComparatorHelper`

Postgres:
- Including SELECT queries generated by the Postgres TLP Aggregate Oracle in the "-cur.log" files if the `logEachSelect()` option is on
- Setting `state.queryString` to "null" in the beginning of each test for all oracles in the Postgres implementation, in order to clear the value from the previous test (in the current scenario, if there is an exception in the middle of the test before the results are compared, the `state.queryString` from the previous test is printed at the bottom of the log file)
- Standardizing the format of the `state.queryString` and AssertionError messages being printed

### DIFF
--- a/src/sqlancer/ComparatorHelper.java
+++ b/src/sqlancer/ComparatorHelper.java
@@ -85,7 +85,7 @@ public final class ComparatorHelper {
     public static void assumeResultSetsAreEqual(List<String> resultSet, List<String> secondResultSet,
             String originalQueryString, List<String> combinedString, GlobalState<?, ?> state) {
         if (resultSet.size() != secondResultSet.size()) {
-            String queryFormatString = "%s; -- cardinality: %d";
+            String queryFormatString = "-- %s;\n-- cardinality: %d";
             String firstQueryString = String.format(queryFormatString, originalQueryString, resultSet.size());
             String secondQueryString = String.format(queryFormatString,
                     combinedString.stream().collect(Collectors.joining(";")), secondResultSet.size());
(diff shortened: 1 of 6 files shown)


mrigger:
Thanks a lot for your contribution! The changes look good to me. It seems that the CI currently fails due to an formatting issue. You can use `mvn formatter:format` to format the sources.

Sorry for the messy logging logic. The logging infrastructure needs a complete overhaul, and I hope I'll be able to invest some time to improve this soon.

nukoyluoglu:
@mrigger 
Also, here is a bug I reported in PostgreSQL that I found using SQLancer: 

https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=f1fcf2d3b2e00b3d7ad3e0d2b1d26b77f5a48413

mrigger:
Thanks a lot for analyzing and reporting the bug, @nukoyluoglu, as well as for mentioning SQLancer in the report. I really appreciate this!

Btw, I've noticed that you have been using and extending SQLancer to test Citus. I'd be happy to learn more about your experience and plans in this regard! Could I state at https://www.manuelrigger.at/dbms-bugs/ that citusdata has been extending and using SQLancer to test Citus?

mrigger:
Could you please squash the merges and format fixes (i.e., the last five commits based on the current status) into one commit, when the CI succeeds, so that the commit history is better readable?

mrigger:
I'm really sorry about the additional effort, but I think something went wrong with the squashing. The changes now seem to affect 50 different files. I typically use the following workflow:
```
git checkout master
git pull
git checkout <branch>
git rebase master
```
This way, the changes are applied on the latest trunk. In your case, it might help, so that unrelated files aren't updated.

PR #119, opened 2020-07-22
prevent SQL injection with prepared & filled-in statements

nukoyluoglu:
This code allows the option to pass in a template SQL statement and variables to fill in, which prevents [SQL injection](https://www.w3schools.com/sql/sql_injection.asp). It also diversifies the query execution options across classes by including the executeAndGet function in more places.

### DIFF
--- a/src/sqlancer/GlobalState.java
+++ b/src/sqlancer/GlobalState.java
@@ -1,6 +1,7 @@
 package sqlancer;
 
 import java.sql.Connection;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 
 import sqlancer.Main.QueryManager;
@@ -89,7 +90,7 @@ public abstract class GlobalState<O, S> {
         this.databaseName = databaseName;
     }
 
-    public boolean executeStatement(Query q) throws SQLException {
+    public ExecutionTimer executePrologue(Query q) throws SQLException {
         boolean logExecutionTime = getOptions().logExecutionTime();
         ExecutionTimer timer = null;
         if (logExecutionTime) {
@@ -105,7 +106,11 @@ public abstract class GlobalState<O, S> {
                 getLogger().writeCurrent(q.getQueryString());
             }
         }
-        boolean success = manager.execute(q);
+        return timer;
+    }
+
+    public void executeEpilogue(Query q, boolean success, ExecutionTimer timer) throws SQLException {
+        boolean logExecutionTime = getOptions().logExecutionTime();
         if (success && getOptions().printSucceedingStatements()) {
             System.out.println(q.getQueryString());
         }
@@ -115,7 +120,21 @@ public abstract class GlobalState<O, S> {
         if (q.couldAffectSchema()) {
             updateSchema();
         }
+    }
+
+    public boolean executeStatement(Query q, String... fills) throws SQLException {
+        ExecutionTimer timer = executePrologue(q);
+        boolean success = manager.execute(q, fills);
+        executeEpilogue(q, success, timer);
         return success;
+    }
+
+    public ResultSet executeStatementAndGet(Query q, String... fills) throws SQLException {
+        ExecutionTimer timer = executePrologue(q);
+        ResultSet result = manager.executeAndGet(q, fills);
+        boolean success = result != null;
+        executeEpilogue(q, success, timer);
+        return result;
     }
 
     public S getSchema() {
--- a/src/sqlancer/Main.java
+++ b/src/sqlancer/Main.java
@@ -8,6 +8,7 @@ import java.io.StringWriter;
 import java.io.Writer;
 import java.nio.file.Files;
 import java.sql.Connection;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 import java.text.DateFormat;
@@ -253,11 +254,20 @@ public final class Main {
             this.globalState = globalState;
         }
 
-        public boolean execute(Query q) throws SQLException {
+        public boolean execute(Query q, String... fills) throws SQLException {
             globalState.getState().logStatement(q);
-            boolean success = q.execute(globalState);
+            boolean success;
+            success = q.execute(globalState, fills);
             Main.nrSuccessfulActions.addAndGet(1);
             return success;
+        }
+
+        public ResultSet executeAndGet(Query q, String... fills) throws SQLException {
+            globalState.getState().logStatement(q);
+            ResultSet result;
+            result = q.executeAndGet(globalState, fills);
+            Main.nrSuccessfulActions.addAndGet(1);
+            return result;
         }
 
         public void incrementSelectQueryCount() {
--- a/src/sqlancer/QueryAdapter.java
+++ b/src/sqlancer/QueryAdapter.java
@@ -1,5 +1,6 @@
 package sqlancer;
 
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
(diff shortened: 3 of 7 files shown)


src/sqlancer/GlobalState.java
```diff
@@ -118,6 +120,93 @@ public boolean executeStatement(Query q) throws SQLException {
         return success;
     }
 
+    public ResultSet executeStatementAndGet(Query q) throws SQLException {
```
mrigger, line 123: It seems that much code is duplicated here with respect to the `executeStatement` method. Could you try to factor out common functionality?

src/sqlancer/QueryAdapter.java
```diff
@@ -117,4 +119,48 @@ public boolean couldAffectSchema() {
...
+            for (int i = 1; i < fills.size() + 1; i++) {
+                s.setString(i, fills.get(i - 1));
+            }
+            result = s.executeQuery();
+            Main.nrSuccessfulActions.addAndGet(1);
+            return result;
+        } catch (Exception e) {
+            s.close();
+            boolean isExcluded = false;
+            Main.nrUnsuccessfulActions.addAndGet(1);
```
mrigger, line 152: Please use the `checkException` method here.

mrigger:
If SQL injection is a concern, we might want to consider enabling the `SQL_NONCONSTANT_STRING_PASSED_TO_EXECUTE` SpotBugs check:

https://github.com/sqlancer/sqlancer/blob/f47d180246a7e86bb02f76cf8ceba698dbf6a668/configs/spotbugs-exclude.xml#L8

Currently, it fails with some errors:
```
[ERROR] High: sqlancer.mariadb.MariaDBSchema.fromConnection(Connection, String) passes a nonconstant String to an execute or addBatch method on an SQL statement [sqlancer.mariadb.MariaDBSchema] At MariaDBSchema.java:[line 294] SQL_NONCONSTANT_STRING_PASSED_TO_EXECUTE
[ERROR] High: sqlancer.mariadb.MariaDBSchema.getTableColumns(Connection, String, String) passes a nonconstant String to an execute or addBatch method on an SQL statement [sqlancer.mariadb.MariaDBSchema] At MariaDBSchema.java:[line 339] SQL_NONCONSTANT_STRING_PASSED_TO_EXECUTE
[ERROR] High: sqlancer.mysql.MySQLSchema.fromConnection(Connection, String) passes a nonconstant String to an execute or addBatch method on an SQL statement [sqlancer.mysql.MySQLSchema] At MySQLSchema.java:[line 290] SQL_NONCONSTANT_STRING_PASSED_TO_EXECUTE
[ERROR] High: sqlancer.mysql.MySQLSchema.getTableColumns(Connection, String, String) passes a nonconstant String to an execute or addBatch method on an SQL statement [sqlancer.mysql.MySQLSchema] At MySQLSchema.java:[line 335] SQL_NONCONSTANT_STRING_PASSED_TO_EXECUTE
[ERROR] High: sqlancer.postgres.PostgresSchema.getTableColumns(Connection, String) passes a nonconstant String to an execute or addBatch method on an SQL statement [sqlancer.postgres.PostgresSchema] At PostgresSchema.java:[line 339] SQL_NONCONSTANT_STRING_PASSED_TO_EXECUTE
```

nukoyluoglu:
@mrigger Could you relay me the information about why the TiDB test failed in CI?

src/sqlancer/GlobalState.java
```diff
@@ -89,8 +91,7 @@ public void setDatabaseName(String databaseName) {
         this.databaseName = databaseName;
     }
 
-    public boolean executeStatement(Query q) throws SQLException {
-        boolean logExecutionTime = getOptions().logExecutionTime();
+    public ExecutionTimer logExecution(Query q, boolean logExecutionTime) throws SQLException {
```
mrigger, line 94: Do we need to pass `logExecutionTime` here? I think the method can just use `getOptions()` instead, no?

src/sqlancer/GlobalState.java
```diff
@@ -118,6 +125,54 @@ public boolean executeStatement(Query q) throws SQLException {
...
+            updateSchema();
+        }
+        return result;
+    }
+
+    public boolean fillAndExecuteStatement(Query q, String template, List<String> fills) throws SQLException {
+        boolean logExecutionTime = getOptions().logExecutionTime();
+        ExecutionTimer timer = logExecution(q, logExecutionTime);
+        boolean success = manager.fillAndExecute(q, template, fills);
+        if (success && getOptions().printSucceedingStatements()) {
```
mrigger, line 148: Could we add a method for the common code that is executed after a statement or query is executed? We could name it `executeQueryEpilogue` or so.

src/sqlancer/GlobalState.java
```diff
@@ -118,6 +125,54 @@ public boolean executeStatement(Query q) throws SQLException {
...
+        if (logExecutionTime) {
+            getLogger().writeCurrent(" -- " + timer.end().asString());
+        }
+        if (q.couldAffectSchema()) {
+            updateSchema();
+        }
+        return result;
+    }
+
+    public boolean fillAndExecuteStatement(Query q, String template, List<String> fills) throws SQLException {
```
mrigger, line 144: I think it would be cleaner to just extend the current `executeStatement` method. Rather than passing `List<String> fills`, you can specify `String... fills` as a signature, to specify variable-length arguments. This way, existing code will work without modification, and users of that method can optionally specify the additional fills when needed. Using varargs might also make it easier to use the method, since no intermediate list needs to be executed. For example, the `executeStatement` method could be called like this: `executeStatement(q, templateString, fill1, fill2, fill3)`.

src/sqlancer/Main.java
```diff
@@ -254,30 +254,36 @@ private void printState(FileWriter writer, StateToReproduce state) {
...
+            boolean success;
+            if (templateAndFills.length < 2) {
+                success = q.execute(globalState);
+            } else {
+                String template = templateAndFills[0];
+                List<String> fills = new ArrayList<>();
+                for (int i = 1; i < templateAndFills.length; i++) {
+                    fills.add(templateAndFills[i]);
+                }
+                success = q.fillAndExecute(globalState, template, fills);
```
nukoyluoglu, line 268: Is this what you had in mind? I couldn't remove the fillAndExecute method from all classes, since it prepares the SQL statements in a different way than the normal queries. That's why, I turned the execution methods in the GlobalState and Main classes into a generic mode, but there is a logic inside the execution methods in the Main class that calls different methods from the Query class.

src/sqlancer/Main.java
```diff
@@ -254,30 +254,36 @@ private void printState(FileWriter writer, StateToReproduce state) {
...
+            }
             Main.nrSuccessfulActions.addAndGet(1);
             return success;
         }
 
-        public ResultSet fillAndExecuteAndGet(Query q, String template, List<String> fills) throws SQLException {
+        public ResultSet executeAndGet(Query q, String... templateAndFills) throws SQLException {
```
nukoyluoglu, line 274: Is this what you had in mind? I could not remove the fill-in-specific methods completely since they use different methods to prepare and execute SQL statements. Instead, I turned the methods in the GlobalState and Main classes into a generic format, and altered the method in the Main class to include a logic to decide which type of method from the Query class to call.

mrigger: It's almost what I had in mind: I would still keep the `String template` parameter, since this is a necessary argument, and have the varargs as an additional parameter. Thanks for incorporating the feedback!

nukoyluoglu: Of course! String template is not a necessary argument for the usual query execution, and introducing that could require going into all DBMS implementations and changing that. Currently, `globalStatement.executeStatement(query)` is a valid call, and does not require the query string additionally. Would you like to change the code in each DBMS, then?

mrigger: I looked at the code in greater detail, and I'm not sure if I fully understand. Is the query string in `Query q` left empty when the fills are used? If yes, I think that's undesirable. In this case, I think it would make sense to use the query string to store the template.

nukoyluoglu: Just to clarify, then we wouldn't be passing a separate templateString at all which you asked beforehand. We would be using the String as part of the query.

mrigger: Sorry, I think I initially misunderstood how the method should be used.

nukoyluoglu:
Could you take a look at why the SQLite test failed?

mrigger:
Locally, I could not reproduce this based on the latest SQLite version. I just merged a version bump of the SQLite JDBC driver (see https://github.com/sqlancer/sqlancer/pull/135), which might fix this issue. Could you rebase based on the latest trunk version? If this does not solve the problem, I can try and push a fix directly to your branch.

nukoyluoglu:
The SQLite error seems to persist

mrigger:
I pushed a commit to ignore the SQLite error. The CI now passes, and I'll merge the PR after the other tasks finish.

PR #120, opened 2020-07-22
Support Subqueries in Postgres

nukoyluoglu:
This PR implements support for subqueries for the PostgreSQL implementation, as well as add a function to create subqueries in the PostgresTLPBase class, which can be called as desired.

### DIFF
--- a/src/sqlancer/postgres/ast/PostgresJoin.java
+++ b/src/sqlancer/postgres/ast/PostgresJoin.java
@@ -4,6 +4,7 @@ import sqlancer.Randomly;
 import sqlancer.postgres.PostgresSchema.PostgresDataType;
 import sqlancer.postgres.PostgresSchema.PostgresTable;
 import sqlancer.postgres.ast.PostgresSelect.PostgresCTE;
+import sqlancer.postgres.ast.PostgresSelect.PostgresFromTable;
 
 public class PostgresJoin implements PostgresExpression {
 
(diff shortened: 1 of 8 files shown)


src/sqlancer/postgres/ast/PostgresJoin.java
```diff
@@ -18,17 +19,33 @@ public static PostgresJoinType getRandom() {
     private final PostgresTable table;
     private final PostgresExpression onClause;
     private final PostgresJoinType type;
+    private PostgresCTE CTE = null;
```
mrigger, line 22: I think it would be cleaner here to have a single field (e.g., `PostgresExpression joinedElement`) which either holds a table or a CTE. This requires implementing an additional class for a table reference, like https://github.com/sqlancer/sqlancer/blob/master/src/sqlancer/tidb/ast/TiDBTableReference.java. Do you think this would be possible?

src/sqlancer/postgres/ast/PostgresJoin.java
```diff
@@ -15,18 +17,36 @@ public static PostgresJoinType getRandom() {
...
     }
 
-    private final PostgresTable table;
+    private final PostgresTableReference tableReference;
     private final PostgresExpression onClause;
     private final PostgresJoinType type;
 
-    public PostgresJoin(PostgresTable table, PostgresExpression onClause, PostgresJoinType type) {
-        this.table = table;
+    public static class PostgresTableReference implements PostgresExpression {
```
mrigger, line 24: I'm sorry, but I actually overlooked that a `PostgresFromTable` class already existed. This is what I had in mind when proposing the `PostgresTableReference` class, since it basically allows to treat a subquery and table uniformly as an expression. Would it be possible to remove this class? Sorry again.

src/sqlancer/postgres/ast/PostgresSelect.java
```diff
@@ -56,6 +56,29 @@ public PostgresDataType getExpressionType() {
         }
     }
 
+    public static class PostgresCTE implements PostgresExpression {
```
mrigger, line 59: Do you think that `PostgresSubquery` would be a better name? I noticed that you changed `CTE` to `subquery` also in other places. If yes, It would make sense to perform the naming change also in the other files.

src/sqlancer/postgres/ast/PostgresJoin.java
```diff
@@ -15,18 +17,36 @@ public static PostgresJoinType getRandom() {
...
+            this.tableReference = new PostgresFromTable(table, Randomly.getBoolean());
+        }
+
+        public PostgresExpression getTableReference() {
+            return tableReference;
+        }
+
+    }
+
+    public PostgresJoin(PostgresTableReference tableReference, PostgresExpression onClause, PostgresJoinType type) {
```
mrigger, line 42: Related to the comment above: I think we should use a `PostgresExpression` type here.

nukoyluoglu:
@mrigger Could you relay the information about why the TiDB test failed?

mrigger:
I've restarted the Travis CI for now. I noticed that TiDB started failing non-deterministically yesterday. I'll create a subsequent PR to disable the TiDB checks temporarily.

mrigger:
Can I merge the PR after the CI checks passed (it's still marked as a draft)?

nukoyluoglu:
Yes, definitely. Thank you!

mrigger:
Great, thanks again for your contribution!

In future contributions, you might want to consider using history rewriting (e.g., `git rebase`) to break up the PR into minimal commits. For example, rather than having the commit that renames CTE to subquery, `git rebase` would have been applicable to change the original commit.

PR #124, opened 2020-07-23
[ClickHouse] Temporary disable log engine. Return CI checks

### LINKED ISSUE #121: Disabled ClickHouse CI
In https://github.com/sqlancer/sqlancer/pull/101, we disabled the ClickHouse CI due to some new failures that blocked us from merging the PR. We should fix the issue and enable the CI again as soon as possible, to prevent that we break the ClickHouse implementation during refactoring.

qoega:
Closes #121


qoega:
I'll fix it in another PR #125.

PR #129, opened 2020-07-23
Implement a local state class and replace the queryString in the state

mrigger:
The idea behind this PR is to have a local state object that lives only for one call to the `check()` method in the `TestOracle`. The test oracle can add information that is relevant only if the oracle detects an error of throws an exception, in which case the statements are added to the `StateToReproduce` statements, which are subsequently written to the log file.

@nukoyluoglu, do you have any feedback since you recently improved the logging logic?

### DIFF
--- a/src/sqlancer/ProviderAdapter.java
+++ b/src/sqlancer/ProviderAdapter.java
@@ -3,6 +3,8 @@ package sqlancer;
 import java.io.FileWriter;
 import java.sql.SQLException;
 
+import sqlancer.StateToReproduce.OracleRunReproductionState;
+
 public abstract class ProviderAdapter<G extends GlobalState<O, ?>, O> implements DatabaseProvider<G, O> {
 
     private final Class<G> globalClass;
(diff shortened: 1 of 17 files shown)


src/sqlancer/ProviderAdapter.java
```diff
@@ -41,11 +43,14 @@ public void generateAndTestDatabase(G globalState) throws SQLException {
 
             TestOracle oracle = getTestOracle(globalState);
             for (int i = 0; i < globalState.getOptions().getNrQueries(); i++) {
-                try {
-                    oracle.check();
-                    globalState.getManager().incrementSelectQueryCount();
-                } catch (IgnoreMeException e) {
+                try (OracleRunReproductionState localState = globalState.getState().createLocalState()) {
```
nukoyluoglu, line 46: Is the` localState.close() ` function automatically called after the `try` block is executed here? Or does the `localState.close()` function need to be explicitly called after the block is executed? I am not exactly familiar with how the `Closeable `interface functions in that regard.

mrigger: By using the `try` block, the `close()` method is automatically called when execution leaves this block. I'm not so happy about the magic that is currently involved in this (it's quite in transparent how and when what happens), but I think it's an improvement in the sense that it generalizes the previous `queryString` by allowing to log multiple statements and not having to manually reset the state.

PR #132, opened 2020-07-24
Support SELECT * and fetching multiple columns

nukoyluoglu:
This PR is to allow `select.fetchColumns` to be null, in which case `SELECT *` would be queried. It also allows picking multiple columns to fetch rather than the first one.

### DIFF
--- a/src/sqlancer/postgres/oracle/tlp/PostgresTLPBase.java
+++ b/src/sqlancer/postgres/oracle/tlp/PostgresTLPBase.java
@@ -12,6 +12,7 @@ import sqlancer.TestOracle;
 import sqlancer.gen.ExpressionGenerator;
 import sqlancer.postgres.PostgresGlobalState;
 import sqlancer.postgres.PostgresSchema;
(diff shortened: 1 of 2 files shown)


src/sqlancer/postgres/oracle/tlp/PostgresTLPBase.java
```diff
@@ -65,7 +65,12 @@ public void check() throws SQLException {
     }
 
     List<PostgresExpression> generateFetchColumns() {
-        return Arrays.asList(new PostgresColumnValue(targetTables.getColumns().get(0), null));
+        List<PostgresExpression> fetchColumns = new ArrayList<>();
```
mrigger, line 68: I think you could copy the structure in https://github.com/sqlancer/sqlancer/blob/master/src/sqlancer/sqlite3/oracle/tlp/SQLite3TLPBase.java#L58 to have a column named `*` to implement a `SELECT *`. Like this, we can still the detect the case where the columns were erroneously not set.

mrigger:
LGTM. Thanks!

PR #133, opened 2020-07-24
Add option to specify connection URL

nukoyluoglu:
This PR adds the option for the user to specify the initial connection URL in the Postgres implementation. This allows more flexibility in connecting with different users, hosts, ports, and optional connection parameters. The logging also includes the full URL instead of only the database name. The provided URL is checked for correct format, and parsed. If a username and password is specified in the URL, these overwrite the username and password set by the MainOptions (with sqlancer as default). When calling DriverManager to get connection, the username and password are removed from the URL string to avoid confusion.

### DIFF
--- a/src/sqlancer/postgres/PostgresProvider.java
+++ b/src/sqlancer/postgres/PostgresProvider.java
@@ -2,6 +2,8 @@ package sqlancer.postgres;
 
 import java.io.FileWriter;
 import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
@@ -218,15 +220,46 @@ public final class PostgresProvider extends ProviderAdapter<PostgresGlobalState,
 
     @Override
     public Connection createDatabase(PostgresGlobalState globalState) throws SQLException {
-        String url = "jdbc:postgresql://localhost:5432/test";
+        String username = globalState.getOptions().getUserName();
+        String password = globalState.getOptions().getPassword();
+        String entryPath = "/test";
+        String entryURL = globalState.getDmbsSpecificOptions().connectionURL;
+        // trim URL to exclude "jdbc:"
+        if (entryURL.startsWith("jdbc:")) {
+            entryURL = entryURL.substring(5);
+        }
+        try {
+            URI uri = new URI(entryURL);
+            String userInfoURI = uri.getUserInfo();
+            String pathURI = uri.getPath();
+            if (userInfoURI != null) {
+                // username and password specified in URL take precedence
+                if (userInfoURI.contains(":")) {
+                    String[] userInfo = userInfoURI.split(":", 2);
+                    username = userInfo[0];
+                    password = userInfo[1];
+                } else {
+                    username = userInfoURI;
+                    password = null;
+                }
+                int userInfoIndex = entryURL.indexOf(userInfoURI);
+                String preUserInfo = entryURL.substring(0, userInfoIndex);
+                String postUserInfo = entryURL.substring(userInfoIndex + userInfoURI.length() + 1);
+                entryURL = preUserInfo + postUserInfo;
+            }
+            if (pathURI != null) {
+                entryPath = pathURI;
+            }
+        } catch (URISyntaxException e) {
+            throw new AssertionError(e);
+        }
+        String entryDatabaseName = entryPath.substring(1);
         String databaseName = globalState.getDatabaseName();
-        Connection con = DriverManager.getConnection(url, globalState.getOptions().getUserName(),
-                globalState.getOptions().getPassword());
-        globalState.getState().logStatement("\\c test;");
+        Connection con = DriverManager.getConnection("jdbc:" + entryURL, username, password);
+        globalState.getState().logStatement(String.format("\\c %s;", entryDatabaseName));
         globalState.getState().logStatement("DROP DATABASE IF EXISTS " + databaseName);
         String createDatabaseCommand = getCreateDatabaseCommand(databaseName, con, globalState);
         globalState.getState().logStatement(createDatabaseCommand);
-        globalState.getState().logStatement("\\c " + databaseName);
         try (Statement s = con.createStatement()) {
             s.execute("DROP DATABASE IF EXISTS " + databaseName);
         }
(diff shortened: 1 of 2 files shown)


src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -218,24 +220,51 @@ protected TestOracle getTestOracle(PostgresGlobalState globalState) throws SQLEx
...
+            String userInfoURI = uri.getUserInfo();
+            String pathURI = uri.getPath();
+            if (userInfoURI != null) {
+                // username and password specified in URL take precedence
+                String[] userInfo = userInfoURI.split(":", 2);
+                username = userInfo[0];
+                password = userInfo[1];
+                int userInfoIndex = entryURL.indexOf(userInfoURI);
+                entryURL = entryURL.substring(0, userInfoIndex)
+                        + entryURL.substring(userInfoIndex + userInfoURI.length() + 1);
```
mrigger, line 242: I think it would be more readable to concatenate `username` and `password`, rather than doing `entryURL.substring(userInfoIndex + userInfoURI.length() + 1)` (if that's correct). Efficiency is not really a concern here, and I think doing so is less error prone.

src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -218,24 +220,51 @@ protected TestOracle getTestOracle(PostgresGlobalState globalState) throws SQLEx
...
+            URI uri = new URI(entryURL);
+            String userInfoURI = uri.getUserInfo();
+            String pathURI = uri.getPath();
+            if (userInfoURI != null) {
+                // username and password specified in URL take precedence
+                String[] userInfo = userInfoURI.split(":", 2);
+                username = userInfo[0];
+                password = userInfo[1];
+                int userInfoIndex = entryURL.indexOf(userInfoURI);
+                entryURL = entryURL.substring(0, userInfoIndex)
```
mrigger, line 241: I think extracting `entryURL.substring(0, userInfoIndex)` to a variable would improve readability. Is it the host name?

src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -218,24 +220,51 @@ protected TestOracle getTestOracle(PostgresGlobalState globalState) throws SQLEx
...
             s.execute("DROP DATABASE IF EXISTS " + databaseName);
         }
         try (Statement s = con.createStatement()) {
             s.execute(createDatabaseCommand);
         }
         con.close();
-        con = DriverManager.getConnection("jdbc:postgresql://localhost:5432/" + databaseName,
-                globalState.getOptions().getUserName(), globalState.getOptions().getPassword());
+        int pathIndex = entryURL.indexOf(entryPath);
+        String testURL = entryURL.substring(0, pathIndex + 1) + databaseName
```
mrigger, line 264: Same here: I think it would be cleaner to define the variables once, and then use these, rather than using the `substring` method.

src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -218,24 +220,51 @@ protected TestOracle getTestOracle(PostgresGlobalState globalState) throws SQLEx
...
+            if (pathURI != null) {
+                entryPath = pathURI;
+            }
+        } catch (URISyntaxException e) {
+            throw new AssertionError(e);
+        }
         String databaseName = globalState.getDatabaseName();
-        Connection con = DriverManager.getConnection(url, globalState.getOptions().getUserName(),
-                globalState.getOptions().getPassword());
-        globalState.getState().logStatement("\\c test;");
```
mrigger, line 225: Do we need to remove this? Logging a `\c test` is useful when reducing a test case manually, since changing to the default database `test` makes it possible to drop the current database of the test case.

src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -218,24 +220,51 @@ protected TestOracle getTestOracle(PostgresGlobalState globalState) throws SQLEx
...
+        Connection con = DriverManager.getConnection("jdbc:" + entryURL, username, password);
+        globalState.getState().logStatement(entryURL);
```
mrigger, line 252: What is the entry URL here? If we log it without appending a `-- ` first, is the log still executable as a SQL script?

src/sqlancer/postgres/PostgresProvider.java
nukoyluoglu: Actually, I am removing username and password here instead. Otherwise, if username and password are specified both in the DriverManager call and in the url passed to the Driver Manager, the conflict causes errors. So the concatenation I am doing is [urlBeforeUserInfo]+[urlAfterUserInfo].

src/sqlancer/postgres/PostgresProvider.java
nukoyluoglu: The only issue is that these are a combination of many possible things that the user can decide to include in the URL. This makes it hard to define variables, but allows more flexibility in setting optional parameters.

src/sqlancer/postgres/PostgresProvider.java
nukoyluoglu: This is the connection string. Does logging involve executing SQL statements? I thought they were for printing, more clarification on that would be great in case I missed something!

src/sqlancer/postgres/PostgresProvider.java
nukoyluoglu: It is everything that comes before the userInfo, so "postgresql:".

src/sqlancer/postgres/PostgresProvider.java
mrigger: Things written to the log should typically either be SQL statements or comments (prefixed by `--`), so that the log file can basically be directly used as an input to a DBMS. I should update the documentation about this.

src/sqlancer/postgres/PostgresProvider.java
mrigger: Okay, I see. Thanks for the explanation! Maybe as a future PR, we could consider adding test cases to check if connection strings are correctly parsed (although it doesn't have a high priority at the moment).

src/sqlancer/postgres/PostgresProvider.java
mrigger: Okay, thanks for the explanation! If you think that additional variable names would not convey additional information, we can also leave it like it is.

mrigger:
Thanks! Let me know when the PR is ready to be merged. Please squash the two commits.

src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -218,24 +220,54 @@ protected TestOracle getTestOracle(PostgresGlobalState globalState) throws SQLEx
...
+        }
+        try {
+            URI uri = new URI(entryURL);
+            String userInfoURI = uri.getUserInfo();
+            String pathURI = uri.getPath();
+            if (userInfoURI != null) {
+                // username and password specified in URL take precedence
+                String[] userInfo = userInfoURI.split(":", 2);
+                username = userInfo[0];
+                password = userInfo[1];
```
thanodnl, line 239: What happens when only the username is provided and no password?
I think we will get an out of bound exception on the last line here.

connection url to test: `postgres://foo@localhost:5432/test`

nukoyluoglu: I added a check to see if a : separator exists and parse it accordingly. DriverManager.getConnection() seems to safely accept null as a valid password, so I set the password to null if there is no password provided

nukoyluoglu:
> Thanks! Let me know when the PR is ready to be merged. Please squash the two commits.

The PR is ready to be merged, thank you!

mrigger:
Thanks! Sorry for the delay with merging, I was on holidays until including yesterday.

PR #134, opened 2020-07-28
Fix in comparing table columns

nukoyluoglu:
Currently, the method used to compare AbstractTableColumns a) assumes that the column was initiated with a specified (which is not always the case, see PostgresColumn constructer), and b) incorrectly compares between table name and column name (on the line `table.getName().contentEquals(getName())`). This is a fix for both of these problems.


mrigger:
Assuming that you probably don't plan to add a test case for now, should I go ahead and merge the PR?

mrigger:
I noticed that you closed this PR before merging. Was this deliberate?

nukoyluoglu:
Nope, I thought it was merged and was cleaning up my branches. Sorry!

PR #136, opened 2020-07-30
Check for null second result in PostgresTLPAggregateOracle

nukoyluoglu:
Currently, the code only checks whether the first result fetched by the TLP Aggregate Oracle is null. This misses the case where the second case could be null and cause nullpointer errors.


mrigger:
LGTM!

nukoyluoglu:
The build does not fail when I run mvn tests locally. Could you check why these failed?

mrigger:
Did you try to use `mvn verify` (see https://github.com/sqlancer/sqlancer/blob/master/docs/DEVELOPMENT.md#continuous-integration-and-test-suite)? It seems like two PMD checks failed.

PR #141, opened 2020-08-04
Re-enable the ClickHouse tests

mrigger:


qoega:
I guess I need to look at this failures. I guess it is easier to add errors to skips as it is nontrivial to limit columns for `order by` in aggregations only to `group by` ones.

mrigger:
Okay, sounds good. Thanks for looking into this!

PR #147, opened 2020-08-10
Prepared statement execution & Deterministic states to reproduce

nukoyluoglu:
This PR has 2 purposes implemented via the 2 commits:
1. Currently, for prepared statements, since the template without fills is being passed in as the query string, this causes the log outputs to include '?' in them rather than the fill-ins. I resorted back to passing in the filled-in statement as the query string, and passing both the template and the fill-ins as variable-length arguments.
2. Currently, the order of reading from the Postgres schema is non-deterministic and depends on the order Postgres outputs the SELECTed rows. This has caused issues in using a seed to reproduce a run. This commit makes the ordering deterministic by adding ORDER BY clauses.

### DIFF
--- a/src/sqlancer/QueryAdapter.java
+++ b/src/sqlancer/QueryAdapter.java
@@ -61,9 +61,9 @@ public class QueryAdapter extends Query {
     public boolean execute(GlobalState<?, ?> globalState, String... fills) throws SQLException {
         Statement s;
         if (fills.length > 0) {
-            s = globalState.getConnection().prepareStatement(getQueryString());
-            for (int i = 0; i < fills.length; i++) {
-                ((PreparedStatement) s).setString(i + 1, fills[i]);
+            s = globalState.getConnection().prepareStatement(fills[0]);
+            for (int i = 1; i < fills.length; i++) {
+                ((PreparedStatement) s).setString(i, fills[i]);
             }
         } else {
             s = globalState.getConnection().createStatement();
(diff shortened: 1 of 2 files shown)


src/sqlancer/QueryAdapter.java
```diff
@@ -61,9 +61,9 @@ public String getQueryString() {
     public boolean execute(GlobalState<?, ?> globalState, String... fills) throws SQLException {
         Statement s;
         if (fills.length > 0) {
-            s = globalState.getConnection().prepareStatement(getQueryString());
-            for (int i = 0; i < fills.length; i++) {
-                ((PreparedStatement) s).setString(i + 1, fills[i]);
+            s = globalState.getConnection().prepareStatement(fills[0]);
```
mrigger, line 64: I might have forgotten, but was there a specific reason why the template string is passed together with the fills, and not as part of a separate argument? I think packing both together is confusing (also considering that the name is misleading, if not only fills are passed).

nukoyluoglu: I had them separately beforehand, but I recall you wanted me to include the template in the query string, which resulted in the logs to not include the fills. Since I cannot have two variable length arguments, adding the template argument would require this argument to be used in all cases, including the cases where we don’t use prepared statements (but just the query). This would alter the behavior of the tool in large scale in my opinion.

mrigger: Sorry for getting confused about this, and thanks for the explanation. Do you think it would make sense to derive a `PreparedQuery` interface that extends `Query`, and a `PreparedQueryAdapter` that extends `QueryAdapter`? My concern is that the current approach is somewhat surprising, since, if I understand correctly, the query string is left empty and the two types of arguments are passed in the varargs argument.

nukoyluoglu: In the new approach in this PR, the query string isn‘t left empty. Instead it is set to be the filled-in version of the statement (but that is done explicitly, there are examples of that in the CitusProvider class in my other PR). This is so that when logging, the filled in version that is stored in the query is logged. I could implement a separate interface to do this implicitly as you suggested though.

mrigger: I think having this separated out as a separate (sub)class and interface would be better. Do you want to do that in a follow-up PR, so that we can get the Citus PR merged faster?

nukoyluoglu: Sounds good, but I wanted to clarify what you mean. Would you like to merge this PR and make the change we talked about later on, or would you like to cancel this PR? I will change the Citus implementation accordingly if it is the latter.

mrigger: I thought about first merging this PR, so you can do the changes after merging the Citus implementation.

nukoyluoglu: Sounds good to me, feel free to merge this PR!

Issue #32, opened 2020-06-10
Abstraction for expected error messages


mrigger:
To specify the expected error messages for a SQL statement, we currently use `Collection<String>`, see
https://github.com/sqlancer/sqlancer/blob/af4374fb62419aed966c074b7888ddc78a98313e/src/sqlancer/Query.java#L27

It would be desirable to introduce an abstraction to also allow, for example, specifying regular expressions to cover multiple expected error messages at once, and thus make expected error messages more descriptive (since without using regular expressions, we need to check for the shortest common string).

mrigger:
In https://github.com/sqlancer/sqlancer/blob/a195cdd0b4a97cf1113c4bded2e1a41281d4a26a/src/sqlancer/common/query/ExpectedErrors.java, we could do the following:
* Add a field with a list of regular expressions.
* Add methods `addRegex` and `addAllRegexes`.
* Check whether any regexes match in `errorIsExpected`.
* Add test cases to https://github.com/sqlancer/sqlancer/blob/a195cdd0b4a97cf1113c4bded2e1a41281d4a26a/test/sqlancer/TestExpectedErrors.java.
* Start refactoring by using the newly-added functionality at some places (as a proof of concept). For example, the CockroachDB implementation might be a good place: https://github.com/sqlancer/sqlancer/blob/a195cdd0b4a97cf1113c4bded2e1a41281d4a26a/src/sqlancer/cockroachdb/CockroachDBErrors.java#L214.

mrigger:
@albertZhangTJ, do you want to have a look at this issue? I think working on this would be relevant, as we also need to think about how we could generate or specify the list of `ExpectedErrors` in your DSL.

PR #149, opened 2020-08-12
Add an expected error class

mrigger:
This class should introduce additional flexibility when dealing with expected errors. For example, future improvements could add support for regular expressions.

Partly addresses https://github.com/sqlancer/sqlancer/issues/32.

### DIFF
--- a/src/sqlancer/ComparatorHelper.java
+++ b/src/sqlancer/ComparatorHelper.java
@@ -31,7 +31,7 @@ public final class ComparatorHelper {
         return Math.abs(a - b) < 0.0001 * Math.max(Math.abs(a), Math.abs(b));
     }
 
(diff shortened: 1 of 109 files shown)


mrigger:
@nukoyluoglu or @qoega, do you want to have a quick look at this PR to see if the changes make sense to you, since they also affect your implementations?

qoega:
My part looks OK.
Actually I haven't figured out yet if I can actually use regex for errors like `Something happened with TOKEN when TOKEN2`.

mrigger:
Thanks for looking over the changes! Currently, this is not supported, but I plan to add such functionality in of the next PRs. It should then be possible to write an expected error specification like `"Something happened with .* when .*`"

nukoyluoglu:
Looks good to me, thanks for applying the changes to the Citus classes as well!

PR #152, opened 2020-08-13
Add of Citus in Readme & Citus CI jobs

nukoyluoglu:
Draft for testing the Travis job for Citus


src/sqlancer/citus/README.md
```diff
@@ -0,0 +1,141 @@
+# SQLancer for Citus (PostgreSQL extension)
```
mrigger, line 1: I think a better location for this would be the https://github.com/sqlancer/sqlancer/tree/master/docs directory (unless you have a strong opinion against this). We can also link to it from the main `README.md`.

There is much information that is redundant to the main `README.md`, which is undesirable, since it makes it difficult to keep the documentation is up-to-date. Please document only the Citus-specific parts.

src/sqlancer/citus/README.md
```diff
@@ -0,0 +1,141 @@
...
+The following commands run the Citus implementation of SQLancer using Ternary Logic Query Partitioning (TLP):
+
+```
+cd target
+java -jar SQLancer-0.0.1-SNAPSHOT.jar --num-threads 4 citus --oracle QUERY_PARTITIONING
+```
+
+How to configure the run and how to interpret the output is explained [here](https://github.com/sqlancer/sqlancer#using-sqlancer).
+
+Here is a list of main and Citus implementation-specific options that can be passed in as flags using the following format:
```
mrigger, line 38: The option list will likely be outdated soon. Please remove this, or restrict it to the Citus-specific options, if you expect them to not change, or if you want to keep this part up-to-date.

docs/Citus README.md
```diff
@@ -0,0 +1,42 @@
...
+
+How to configure the run and how to interpret the output is explained [here](https://github.com/sqlancer/sqlancer#using-sqlancer).
+
+# Contributing
+
+The instructions for setting up a development environment for contributing to SQLancer can be found [here](https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md).
+
+## Updating expected Citus errors list
+
+The `CitusCommon.java` file in the `src/sqlancer/citus/gen/` directory should be continuously updated to reflect the currently unsupported functionalities and the willingness to ignore active bugs. The `addCitusErrors()` method adds errors that should be ignored if raised, either because a SQL command generated by SQLancer is not supported by Citus, or because a bug that has already been identified is redundantly re-appearing.
```
thanodnl, line 34: This could use a small description on explaining the concept of expected errors, why it is important and then how to update. Now this might be hard to understand its significance if someone on the citus team without background knowledge needs to update the expected errors.

nukoyluoglu:
@mrigger Currently the CI test for Citus is a bit flaky, as there are certain commands that might fail due to errors that are not fixed in the current Citus version used in the CI test. An example: https://travis-ci.com/github/sqlancer/sqlancer/builds/180496594 - It would be very difficult to always keep the Citus version used in the CI test up to date, and there is no guarantee that all syntax errors found in Citus could be fixed promptly. Do you have any suggestions in making sure that tests that fail due to bugs in the DBMS implementation don’t affect the CI test?

mrigger:
I think you could add a class like https://github.com/sqlancer/sqlancer/blob/master/src/sqlancer/tidb/TiDBBugs.java with static fields to ignore known bugs until they are fixed in the version that the CI uses. The static fields are referenced in the code fragment that work around the bug, or add expected errors. Once the bug is fixed, such a class makes it easy to identify which code fragments need changes.

nukoyluoglu:
> I think you could add a class like https://github.com/sqlancer/sqlancer/blob/master/src/sqlancer/tidb/TiDBBugs.java with static fields to ignore known bugs until they are fixed in the version that the CI uses. The static fields are referenced in the code fragment that work around the bug, or add expected errors. Once the bug is fixed, such a class makes it easy to identify which code fragments need changes.

I think the CitusCommon class currently does this by ignoring errors for bugs that are not yet fixed, and we can include these errors there as well. However, this would mean that we would have to ignore errors in the code until the fix made on the master branch of the Citus repository is reflected on the release version used in the CI job. This delay shouldn't be a big problem, though, and we might consider compiling Citus from source code in the CI jobs as well to avoid this delay.

mrigger:
I think there is no clear path on what to do. If we test against the latest Citus version, a bug in Citus could block the development of SQLancer (see https://github.com/sqlancer/sqlancer/issues/3), so also that solution is not optimal. At this point in time, it seems that either option (i.e., using the latest development or release version) would be okay.

PR #154, opened 2020-08-14
Fix the random number generation

mrigger:
Previously, the random number data generation consistently generated the same data sequence for every thread's database. This seems to have been introduced in commit 183b93e43b829f12bb47b95853379286b1491029.

### DIFF
--- a/src/sqlancer/Main.java
+++ b/src/sqlancer/Main.java
@@ -296,17 +296,17 @@ public final class Main {
         private final MainOptions options;
         private final O command;
         private final String databaseName;
-        private final long seed;
         private StateLogger logger;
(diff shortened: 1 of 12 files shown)


mrigger:
@qoega @nukoyluoglu @hannesmuehleisen This is a serious bug, since this basically means that SQLancer could find bugs only for the first database created by each thread, wasting the subsequent machine resources. I'm very sorry that I didn't discover this earlier. If you suddenly find a number of additional bugs, this might likely be due to this fix.

hannesmuehleisen:
Ah it happens to everyone

nukoyluoglu:
Thank you for letting us know and fixing it!

PR #160, opened 2020-08-20
Re-enable ClickHouse

qoega:


mrigger:
It seems that this broke the build: https://travis-ci.com/github/sqlancer/sqlancer/jobs/375747149

Could you please have a look at this?

PR #180, opened 2020-08-24
Miscellaneous fixes to Citus & Postgres

nukoyluoglu:
This PR:
1- Comments out the creation of the databases in Citus error logs as these are not reproduceable SQL commands (to reproduce, creating an empty Citus cluster with the appropriate workers and then running the file is possible with the new edit).
2- Fixes a bug that caused no functions to ever be used in the PostgresExpressionGenerator
3- Does a general clean-up of the Citus package source code
4. Adds expected error raised when multiple constraints with same name are trying to be added

### DIFF
--- a/docs/Citus README.md
+++ b/docs/Citus README.md
@@ -23,7 +23,27 @@ cd target
 java -jar SQLancer-0.0.1-SNAPSHOT.jar --num-threads 4 citus --oracle QUERY_PARTITIONING
 ```
 
-How to configure the run and how to interpret the output is explained in [SQLancer - Using SQLancer](https://github.com/sqlancer/sqlancer#using-sqlancer).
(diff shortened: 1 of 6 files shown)


mrigger:
It seems that our builds are currently failing due to a Travis issue (see https://travis-ci.community/t/openjdk14-started-to-fail/9700).

nukoyluoglu:
@mrigger would you like me to run mvn verify locally to confirm those are functioning until this problem is fixed? Other than that, I can do sample runs both for Citus and Postgres in this new branch without any problem

mrigger:
https://github.com/sqlancer/sqlancer/pull/181 now also changes the Travis configuration to use OpenJDK rather than OracleJDK. It seems that this is working. After merging that PR, we can merge this PR.

mrigger:
Could you please rebase to merge this PR?

Issue #161, opened 2020-08-20
Modernize the PQS implementation


mrigger:
Since NoREC and TLP have been implemented, I haven't been paying attention to the PQS implementation, which is now probably outdated. It would be desirable to make it work again, and modernize the code.

PR #181, opened 2020-08-24
Provide an option to the maximum length of string constants and set i…

mrigger:
…t to a lower value

See https://github.com/sqlancer/sqlancer/issues/161.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ branches:
 matrix:
   include:
     - name: MariaDB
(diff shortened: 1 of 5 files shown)


mrigger:
Postgres is currently failing with an error `relation "x" already exists` when using `ALTER TABLE`. I'm merging this nevertheless, since https://github.com/sqlancer/sqlancer/pull/180 contributes a fix to this. /cc @nukoyluoglu

PR #182, opened 2020-08-24
fix quick start command

zhangysh1995:
git a hint for users where is the jar file


README.md
```diff
@@ -27,7 +27,7 @@ git clone https://github.com/sqlancer/sqlancer
 cd sqlancer
 mvn package -DskipTests
 cd target
```
mrigger, line 29: Since we use `cd target` here, specifying `./target/` below would break the quick-start instructions. Thank you for your PR nevertheless!

PR #241, opened 2020-11-05
Add known Citus error 'cannot modify views over distributed tables'

halilozanakgul:
Adds known and expected Citus error message 'cannot modify views over distributed tables'.
Citus does not support the modification of views that are created over distributed tables.

We have used SQLancer again during our release process and found some interesting bugs thanks to it.


mrigger:
Thanks a lot for the PR, and I'm thrilled to hear that you are still using SQLancer to find bugs!

PR #242, opened 2020-11-11
Refactor Logging to remove SQL Dependency, keeping same Functionality

pfu3tz:
As of this commit the classes StateLogger and StateToReproduce are no longer dependent on SQL strings. The functionality is now supported by a LoggableFactory, specifically for all current DBMSs the SQLLoggableFactory. Parts of the functionality may be removed later, for example commentStatements is now marked depricated.

### DIFF
--- a/src/sqlancer/clickhouse/ClickHouseProvider.java
+++ b/src/sqlancer/clickhouse/ClickHouseProvider.java
@@ -9,8 +9,8 @@ import java.util.stream.Collectors;
 import sqlancer.AbstractAction;
 import sqlancer.GlobalState;
 import sqlancer.IgnoreMeException;
-import sqlancer.ProviderAdapter;
 import sqlancer.Randomly;
+import sqlancer.SQLProviderAdapter;
 import sqlancer.StatementExecutor;
 import sqlancer.clickhouse.ClickHouseProvider.ClickHouseGlobalState;
 import sqlancer.clickhouse.gen.ClickHouseCommon;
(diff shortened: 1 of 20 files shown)


src/sqlancer/clickhouse/ClickHouseProvider.java
```diff
@@ -133,4 +135,9 @@ public Connection createDatabase(ClickHouseGlobalState globalState) throws SQLEx
     public String getDBMSName() {
         return "clickhouse";
     }
+
+    @Override
+    public LoggableFactory getLoggableFactory() {
+        return new SQLLoggableFactory();
```
mrigger, line 141: For code reuse, it would be nice to implement a `SQLProviderAdapter` class that extends `ProviderAdapter` and implements the `getLoggableFactory` method, so that this method only needs to be implemented once for all SQL-based DBMSs.

src/sqlancer/common/log/DirectLog.java
```diff
@@ -0,0 +1,15 @@
+package sqlancer.common.log;
+
+public class DirectLog implements Loggable{
```
mrigger, line 3: Maybe `LoggedString` would be a better name?

mrigger (commented):
The changes look good, thanks! I have two minor improvement suggestions.

A small nit: For the future, it would be preferable to write commit messages in sentence case (e.g.., "Refactor logging to remove SQL dependency, keeping the same functionality").

mrigger:
LGTM! Thanks

PR #243, opened 2020-11-16
Refactor Exceptions to remove SQLDependency while keeping the same functionality

pfu3tz:
Multiple high-level classes such as DatabaseProvider carry already the semantics for non-SQL DBMS except for the fact that a lot of methods still throw SQLExceptions from the java.sql package. This commit exchanges SQLException with Exception in signatures to solve this.

Also changes the configuration of pmd to exclude the rule "SignatureDeclareThrowsException"


mrigger:
LGTM, thanks!

PR #247, opened 2020-11-26
 Rework Connection to DB to support non-SQL connections

pfu3tz:
In order to achieve this goal, multiple core classes had to change the generics to add a new one C. The connection object of the sql package is now packed into the class SQLConnection. There is now a SQLGlobalState that infers the SQLConnection for the global state and the SQLProviderAdapter infers this for the DBMS specific providers.


src/sqlancer/StatementExecutor.java
```diff
@@ -59,6 +60,7 @@ public void executeStatements() throws Exception {
             assert nextAction != null;
             assert nrRemaining[i] > 0;
             nrRemaining[i]--;
+            @SuppressWarnings("rawtypes")
```
mrigger, line 63: Would `Query<?>` work here, rather than suppressing the warning?

mrigger (commented):
LGTM, thanks! I have a minor improvement suggestion. Also, for the next PRs, please pay attention to the commit guidelines (specifically the format of commit messages and using `git rebase` rather than merges), see https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#commit-history.

src/sqlancer/StatementExecutor.java
pfu3tz: I acually tried that before and it gives a type exception with line 69 executeStatement().

pfu3tz: I think the problem comes from the fact that StatementExecuter does not hold the generics for connection. But as soon as we give StatementExecuter this type extension we probably get into trouble again at other locations...

PR #248, opened 2020-11-30
Clean logic to skip internal indexes.

GreggSchofield:
Looking forward to hearing your feedback!

### DIFF
--- a/src/sqlancer/common/DBMSCommon.java
+++ b/src/sqlancer/common/DBMSCommon.java
@@ -5,7 +5,7 @@ import java.util.regex.Pattern;
 
 public final class DBMSCommon {
 
-    private static final Pattern sqlancerIndexPattern = Pattern.compile("^i\\d+");
+    private static final Pattern SQLANCER_INDEX_PATTERN = Pattern.compile("^i\\d+");
 
     private DBMSCommon() {
     }
@@ -23,7 +23,7 @@ public final class DBMSCommon {
     }
 
     public static boolean matchesIndexName(String indexName) {
-        Matcher matcher = sqlancerIndexPattern.matcher(indexName);
+        Matcher matcher = SQLANCER_INDEX_PATTERN.matcher(indexName);
         return matcher.matches();
     }
 
--- a/src/sqlancer/common/DBMSCommon.java
+++ b/src/sqlancer/common/DBMSCommon.java
@@ -1,0 +1,30 @@
+package sqlancer.common;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class DBMSCommon {
+
+    private static final Pattern sqlancerIndexPattern = Pattern.compile("i\\d+");
+
+    private DBMSCommon() {
+    }
+
+    public static String createTableName(int nr) {
+        return String.format("t%d", nr);
+    }
+
+    public static String createColumnName(int nr) {
+        return String.format("c%d", nr);
+    }
+
+    public static String createIndexName(int nr) {
+        return String.format("i%d", nr);
+    }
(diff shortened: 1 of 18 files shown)


mrigger:
Thank you for the PR and sorry for the late response (I caught a cold). I believe that I left the original FIXME note since checking for the length of indexes being 2 does not provide any guarantees that we select only indexes that we created - the check might still erroneously include internal indexes with a name length of 2, and erroneously exclude indexes that we would want to select even when they exceed the length of 2 (e.g., `i11`). The PR does thus moves the workaround to the SQL side and does not really address the underlying issue.

I think a clean fix would look as follows: I found that Postgres uses a method in https://github.com/sqlancer/sqlancer/blob/61f61fe6c6e3bc64921b7dc296ef824e3ac2daca/src/sqlancer/sqlite3/gen/SQLite3Common.java#L33 to create an index of the specified format. It would be great to create a new class (e.g., named `sqlancer.common.DBMSCommon`), move the methods `createTableName`, `createColumnName` and `createIndexName` to there, and create a method `static boolean matchesIndexName(String indexName)` that returns true if the index name matches the format of the index name created by `createIndexName`. You could then use this newly-created method in the `getIndexes` method. Alternatively, it might be useful to check whether Postgres exposes this information somewhere and check for this.

Would you like to try that? Thanks again for the PR!

GreggSchofield:
Thanks for your feedback Manuel! No worries I have been fighting a cold myself.

Ok I understand what you where trying to do now. Regarding your final comment about checking whether Postgres exposes information about its indexes I believe Postgres indexes are named using the 'pg_' prefix however I could not find any official documentation confirming this. I have implemented the solution you suggested by introducing the DBMSCommon utility class as thought this could possibly be re-used by other databases sqlancer supports (both currently and in the future).

All the best,

Gregg

src/sqlancer/common/DBMSCommon.java
```diff
@@ -0,0 +1,30 @@
+package sqlancer.common;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class DBMSCommon {
+
+    private static final Pattern sqlancerIndexPattern = Pattern.compile("^i\\d+");
```
mrigger, line 8: I think this should be `SQLANCER_INDEX_PATTERN` according to the Java naming convention.

mrigger (commented):
Hi Gregg, thanks a lot for the changes, and I hope that you'll get better soon. The changes look good overall.

I just have a minor comment regarding the naming of a constant. If you use `mvn verify -DskipTests=true`, you see that Checkstyle complains about it. Typically, this would be caught by the Travis gate, but I just realized that Travis has a new pricing model and that the SQLancer repository is already out of credits - I think we'll migrate to GitHub actions soon.

It would be awesome if you could change the variable name, and we could merge the PR right after that. If it's not too much hassle, please use `git rebase -i` to fix this in the original commit that introduced the constant, so that we have a clean commit history (see https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#commit-history).

src/sqlancer/common/DBMSCommon.java
GreggSchofield: Ahh yes thank you!

mrigger:
Thanks for the PR again, I will merge it now. If you'd like to contribute more PRs, please follow the guidelines mentioned above:

> If it's not too much hassle, please use `git rebase -i` to fix this in the original commit that introduced the constant, so that we have a clean commit history (see https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#commit-history).

PR #275, opened 2021-01-12
Fix ClickHouse Insert Generator error when empty column list was used

qoega:
Fixes incorrect inserts like
```
INSERT INTO t0() VALUES ()
```


mrigger:
LGTM! It seems that currently, the formatter check is failing. Please use `mvn formatter:format` to address the failing check, and `git commit --amend`.

mrigger:
Thanks for the PR!

PR #277, opened 2021-01-19
[ClickHouse] Fix aggregation results in TLP

qoega:
Behavior is fixed here https://github.com/ClickHouse/ClickHouse/pull/16123


mrigger:
LGTM, thanks!

Issue #280, opened 2021-01-19
Introduce releases


mrigger:
It would be desirable to publish regular releases of SQLancer, both as GitHub releases and as artifacts hosted on Maven Central.

mrigger:
I have requested OSSRH hosting: https://issues.sonatype.org/browse/OSSRH-63537

PR #282, opened 2021-01-19
Change the JAR name from SQLancer-* to sqlancer-*

mrigger:
This commit changes the JAR naming format from `SQLancer-*.jar` to `sqlancer-*.jar` to follow the typical naming conventions (see https://maven.apache.org/guides/mini/guide-naming-conventions.html). This is in preparation for creating proper releases (see https://github.com/sqlancer/sqlancer/issues/280).

@qoega @hannesmuehleisen @thanodnl @jordanlewis I will wait for a day or two before merging - please let me know if I should wait longer, in case this change might affect your CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -295,7 +295,7 @@ jobs:
       - name: Build
         run: mvn -B package -DskipTests=true
       - name: Shortly run DuckDB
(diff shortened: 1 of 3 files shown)


jordanlewis:
Thanks for the notification! No objections here.

PR #294, opened 2021-01-22
[ClickHouse] Apply setting to original query also

qoega:
There are also Aggregations over 0 rows in original queries for TLP Aggregate.


mrigger:
LGTM!

Issue #95, opened 2020-07-10
Make the host and port configurable


mrigger:
Currently, the JDBC string containing the host name and port is hardcoded. For example, consider the connection string for MySQL:

https://github.com/sqlancer/sqlancer/blob/e62224edcb43fe421dc87601f75090739e34e595/src/sqlancer/mysql/MySQLProvider.java#L248

It would be desirable to make this configurable. We could provide an option to set the JDBC string for each DBMS. For example, for MySQL, the string could look like this:

```
"jdbc:mysql://$(HOST_NAME):$(PORT)/?serverTimezone=UTC&useSSL=false&allowPublicKeyRetrieval=true"
```
We could then also provide two options to set the host name and port, which are then used to replace the placeholder in the JDBC string.

This would make everything configurable, but also be convenient for users. Most users could go with the default options. Some users might need to configure the port and host. Only very few users probably would need to change the JDBC query string.

mrigger:
https://github.com/sqlancer/sqlancer/pull/133 implemented similar functionality for Postgres.

Ethan-DB:
May I try it? If possible, please assign it to me.

mrigger:
Yes, you are very welcome to work on this! As a heads up: I will be quite busy until March 17, so I might not have time to review pull requests or give feedback until after this date.

Ethan-DB:
- [x] postgres
- [ ] MySQL

EthanDBer:
Has this issue been resolved? May it be closed? @mrigger

mrigger:
Yes, thanks for addressing this issue and for the reminder!

PR #312, opened 2021-03-09
Fix#95: support host and port arguments configuration for different DBMSs in SQLancer

Ethan-DB:
# What problem does this PR solve?
Issue Number: #95

# Summary
Supporting host/port arguments for the follow databases:

- [x] PostgreSQL
- [x] MySQL
- [x] MariaDB
- [x] Clickhouse
- [x] CockRoachDB
- [x] TiDB 

# Release note
No release note.

### DIFF
--- a/src/sqlancer/MainOptions.java
+++ b/src/sqlancer/MainOptions.java
@@ -55,7 +55,6 @@ public class MainOptions {
     @Parameter(names = "--port", description = "The port used to log into the DBMS")
     private String port = "sqlancer"; // NOPMD
 
-
     @Parameter(names = "--print-progress-information", description = "Whether to print progress information such as the number of databases generated or queries issued", arity = 1)
     private boolean printProgressInformation = true; // NOPMD
 
@@ -158,11 +157,11 @@ public class MainOptions {
         return password;
     }
 
-    public String getHost(){
+    public String getHost() {
         return host;
     }
 
-    public String getPort(){
+    public String getPort() {
         return port;
     }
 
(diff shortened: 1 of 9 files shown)


src/sqlancer/MainOptions.java
```diff
@@ -49,6 +49,13 @@
     @Parameter(names = "--password", description = "The password used to log into the DBMS")
     private String password = "sqlancer"; // NOPMD
 
+    @Parameter(names = "--host", description = "The host used to log into the DBMS")
+    private String host = "sqlancer"; // NOPMD
+
+    @Parameter(names = "--port", description = "The port used to log into the DBMS")
+    private String port = "sqlancer"; // NOPMD
```
mrigger, line 56: Are these meaningful default values? I assume "sqlancer" is not a valid port.

mrigger: I am wondering whether these should go into `MainOptions` or in the DBMS-specific options. Putting them into `MainOptions` makes sense on a conceptual level. However, putting them into DBMS-specific options (e.g., specifying these in `SQLGlobalState`) might have the advantage of being able to provide default values that are specific to the DBMS. For example, `MySQLGlobalState` could, in its constructor, set the default values for `host` and `port`. These default values would then also be visible in the help options. What do you think?

Ethan-DB: > Are these meaningful default values? I assume "sqlancer" is not a valid port.

It is modified in `xxProvider.java`.

mrigger: I see. Would just leaving them `null` work, given that the `sqlancer` values seem to be placeholders?

Ethan-DB: > I see. Would just leaving them `null` work, given that the `sqlancer` values seem to be placeholders?

I think `sqlancer` is perfect. Because the `host` may be `null`. 
 
```java
if("sqlancer".equals(host)){
     host = default_value;
}
```

Ethan-DB:
@mrigger PTAL

PR #315, opened 2021-03-12
Fix#95: support host and port arguments configuration for different DBMSs in SQLancer

EthanDBer:
# What problem does this PR solve?
Issue number: #95 
# Summary
Supporting host/port arguments for the follow databases:
- [x] PostgreSQL
- [x] MySQL
- [x] MariaDB
- [x] TiDB
- [x] Clickhouse
- [x] CockRoachDB
- [x] Citus extends PostgreSQL 
# Release note
No release note


src/sqlancer/MainOptions.java
```diff
@@ -49,6 +49,12 @@
     @Parameter(names = "--password", description = "The password used to log into the DBMS")
     private String password = "sqlancer"; // NOPMD
 
+    @Parameter(names = "--host", description = "The host used to log into the DBMS")
+    private String host = "sqlancer"; // NOPMD
```
mrigger, line 53: Since these options are not supported by all DBMSs (yet): Would it be possible to insert a (static) check method that is called by every provider that does not support these? This is optional for this PR, but it might be a useful future improvement.

mrigger (commented):
Thanks a lot for the PR. The changes look great overall.

I think that we need to expose the default port and host to the user in some way. See my thoughts in this respect as an in-line comment. Let me know if you have any follow-up questions or thoughts.

Thanks again for the great PR!

src/sqlancer/clickhouse/ClickHouseProvider.java
```diff
@@ -102,9 +104,19 @@ public void generateDatabase(ClickHouseGlobalState globalState) throws Exception
...
+        }
+        if ("sqlancer".equals(port)) {
+            port = "8123";
+        }
+
         ClickHouseOptions clickHouseOptions = globalState.getDmbsSpecificOptions();
         globalState.setClickHouseOptions(clickHouseOptions);
-        String url = "jdbc:clickhouse://localhost:8123/default";
+        // String url = "jdbc:clickhouse://localhost:8123/default";
+        String url = "jdbc:clickhouse://" + host + ":" + port + "/defult";
```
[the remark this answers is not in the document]
Ethan-DB: I will modify it next time pull request.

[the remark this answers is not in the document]
Ethan-DB: > Typo: this should be default.

I will modify it next time pull request.

src/sqlancer/clickhouse/ClickHouseProvider.java
```diff
@@ -102,9 +104,19 @@ public void generateDatabase(ClickHouseGlobalState globalState) throws Exception
 
     @Override
     public SQLConnection createDatabase(ClickHouseGlobalState globalState) throws SQLException {
+        host = globalState.getOptions().getHost();
+        port = globalState.getOptions().getPort();
+        if ("sqlancer".equals(host)) {
```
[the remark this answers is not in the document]
Ethan-DB: This process actually involves two points:
(1) The user does not specify the host and port, then we need to replace it with the default value, which is different in different databases. This may be better set in xxOptions.java of each database.
(2) If the user sets one of host and port, we should use the value provided by the user to replace the default value of these systems. This should probably be set in xxProvider.java.

src/sqlancer/mysql/MySQLProvider.java
```diff
@@ -31,6 +31,10 @@
 import sqlancer.mysql.gen.tblmaintenance.MySQLRepair;
 
 public class MySQLProvider extends SQLProviderAdapter<MySQLGlobalState, MySQLOptions> {
+    protected String username;
```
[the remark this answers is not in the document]
Ethan-DB: Indeed. These are all local variable.  I will modify it the next pull request.

src/sqlancer/MainOptions.java
Ethan-DB: For the `host`, I think that it may use 'localhost' is more appropriate. But for `port` I think it is more reasonable to use placeholders.  Or you can use placeholders for both parameters. After all, this is not the core part of the software. It just wants users to connect to the database with custom `host` and `port` parameters.

Ethan-DB: > Since these options are not supported by all DBMSs (yet): Would it be possible to insert a (static) check method that is called by every provider that does not support these? This is optional for this PR, but it might be a useful future improvement.

We modify the description information to remind users that this value will be reset by the system. Is it feasible?
```
@Parameter(names = "--host", description = "The host used to log into the DBMS. If the user does not specify this value, it will be reset to localhost by the system")
    private String host = "sqlancer"; // NOPMD

@Parameter(names = "--host", description = "The host used to log into the DBMS. If the user does not specify this value, it will be reset to database system default port by the system")
    private String host = "sqlancer"; // NOPMD
```

In fact, another important function of placeholders is to facilitate us to determine whether the user has specified this parameter, so invalid values ​​should be used as placeholders. The value `sqlancer` is a good choice.  These two values ​​are actually essentially different from username and password. The default of host is `localhost`. The default of `port` is database parameter.

EthanDBer:
TBR @mrigger

src/sqlancer/clickhouse/ClickHouseProvider.java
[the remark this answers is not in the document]
mrigger: Yes, there are multiple ways to implement this. It would be great to have this, so that the user knows about the default values.

src/sqlancer/MainOptions.java
mrigger: I do not yet understand yet why we need any placeholders in the options. None of the other options have deliberately invalid default values, I believe. If you want to indicate that no port or host have been set, should they be `null`?

EthanDBer: > I do not yet understand yet why we need any placeholders in the options. None of the other options have deliberately invalid default values, I believe. If you want to indicate that no port or host have been set, should they be `null`?

Yes, null is also possible. But according to the previous design, does the system still have to reset the host and port to the corresponding to `localhost` and the default port of database respectively? If you think `null` is more appropriate, I am willing to modify it. LMK, please.

src/sqlancer/mysql/MySQLProvider.java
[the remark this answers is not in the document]
mrigger: I do not really understand. Could you expand on what issues there might be when using a local variable rather than a field?

src/sqlancer/MainOptions.java
mrigger: Yes, I believe that either `null`, while also printing the default values when the help text is printed, or implementing them as DBMS-specific options (rather than in `MainOptions`) would be ways to achieve this.

mrigger:
Sorry for the delayed response! As mentioned, it would be great to address the two concerns I mentioned: (1) we want to display the default values that each DBMS uses to the user as part of the help text and (2) not use invalid default values (i.e., either use `null`  or implement them as DBMS-specific options). Please let me know if there are any unclarities, and I will provide more details.

EthanDBer:
TBR @mrigger 
If there is anything inappropriate, please let me know!

src/sqlancer/cockroachdb/CockroachDBOptions.java
```diff
@@ -23,7 +23,7 @@
 import sqlancer.common.oracle.CompositeTestOracle;
 import sqlancer.common.oracle.TestOracle;
 
-@Parameters(separators = "=", commandDescription = "Test CockroachDB")
+@Parameters(separators = "=", commandDescription = "CockroachDB (default port: 26257, default host: localhost)")
```
mrigger, line 26: Rather than hard-coding the port and localhost, could you please introduce constants for them that you use in both the description and when setting the values?

For example, in `CockroachDBOptions` you can have a field `final static int DEFAULT_PORT = 26257;` and reference it both here as `"CockroachDB (default port: " + CockroachDBOptions.DEFAULT_PORT + ...` and below in `createDatabase`. By doing so, the help text and actual default values will always be consistent.

src/sqlancer/MainOptions.java
```diff
@@ -49,6 +49,12 @@
     @Parameter(names = "--password", description = "The password used to log into the DBMS")
     private String password = "sqlancer"; // NOPMD
 
+    @Parameter(names = "--host", description = "The host used to log into the DBMS")
+    private String host = null; // NOPMD
+
+    @Parameter(names = "--port", description = "The port used to log into the DBMS")
+    private int port = -1; // NOPMD
```
mrigger, line 56: Rather than checking for `-1`, could you please introduce a constant like `NO_PORT_SET` and use it also in the `createDatabase` methods?

src/sqlancer/tidb/TiDBProvider.java
```diff
@@ -132,8 +134,19 @@ public void generateDatabase(TiDBGlobalState globalState) throws Exception {
...
+        if (host == null) {
+            host = "localhost";
+        }
+        if (port == -1) {
+            port = 4000;
+        }
+
         String databaseName = globalState.getDatabaseName();
-        String url = "jdbc:mysql://127.0.0.1:4000/";
+        // String url = "jdbc:mysql://127.0.0.1:4000/";
```
mrigger, line 147: Ideally, we could remove the commented-out code (but you don't need to do this for this PR).

src/sqlancer/mysql/MySQLProvider.java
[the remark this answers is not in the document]
mrigger: Ideally, we would still use local variables here, but if you are reluctant to do this for this PR, I can also do this in a future follow-up PR.

mrigger:
> TBR @mrigger
> If there is anything inappropriate, please let me know!

Thanks a lot again for your changes and for your patience! After introducing constants and using them to replace the hard-coded values, we can likely merge the PR.

src/sqlancer/cockroachdb/CockroachDBOptions.java
EthanDBer: "CockroachDB (default port: " + CockroachDBOptions.DEFAULT_PORT + ... "   I think it might not work.
It works in `createDatabase`.

EthanDBer: Indeed, the current help text and default value can be consistent here. 
Thank you for your suggestions.  
TBR @mrigger

EthanDBer:
TBR @mrigger Thanks for your suggestions.  Please review the latest revisions, and if necessary, please suggest revisions.

src/sqlancer/cockroachdb/CockroachDBOptions.java
mrigger: Sorry again for the late response - I had some full-day obligations in the last couple of days, and will be able to respond much faster in the future.

Why does using the constant in the annotation text not work?

mrigger:
Thanks a lot for your PR and for incorporating my improvement suggestions. I highly appreciate it! Merging the PR now.

For future PRs, please pay attention to a "clean" commit history by avoid merges as well as rewriting the history for incorporating feedback (and removing accidentally committed files), see https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#commit-history.

EthanDBer:
> Thanks a lot for your PR and for incorporating my improvement suggestions. I highly appreciate it! Merging the PR now.
> 
> For future PRs, please pay attention to a "clean" commit history by avoid merges as well as rewriting the history for incorporating feedback (and removing accidentally committed files), see https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#commit-history.

Thanks for your suggestions. It’s a great honor to contribute to this project.

PR #325, opened 2021-03-27
Add a new H2 expected error for foreign constraints

mrigger:
Consider the following test case:

```sql
CREATE TABLE t0(c0 INT);
CREATE TABLE t1(c0 INT, FOREIGN KEY(c0) REFERENCES t0(C0)); -- Constraint "PRIMARY KEY | UNIQUE (C0)" not found;
```

Due to a recent H2 change, the second `CREATE TABLE` statement now results in an error, while previously, it was executed successfully.

### DIFF
--- a/src/sqlancer/h2/H2TableGenerator.java
+++ b/src/sqlancer/h2/H2TableGenerator.java
@@ -102,6 +102,7 @@ public class H2TableGenerator {
             }
             errors.add("are not comparable");
             errors.add(" cannot be updatable by a referential constraint with"); // generated columns
+            errors.add("not found"); // Constraint "PRIMARY KEY | UNIQUE (C0)" not found;
         }
         sb.append(")");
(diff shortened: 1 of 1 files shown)


mrigger:
@katzyn is this semantic change in H2 deliberate?

katzyn:
The SQL Standard clearly requires presence of unique (or primary key) constraint on the referenced columns and DBMS 
usually require that too with some exclusions.

I have no idea why H2 allowed creation of referential constraint without a unique constraint, but such behavior was error-prone, some people created a constraint on one set of columns and tried to reference an another set. A decision was made to fix this issue (with exclusion for MySQL compatibility mode). MariaDB also require an existing constraint, unlike MySQL.

BTW, it would be better to check `SQLException.getSQLState()` instead of error messages.

mrigger:
Thanks a lot for the valuable input!

PR #344, opened 2021-04-26
Update CockroachDB to v20.2.8 and enable tests

mgartner:
CockroachDB tests were disabled after they revealed a bug in v20.2.6,
https://github.com/cockroachdb/cockroach/issues/62281. This bug was
fixed and v20.2.8. This commit updates the CockroachDB version and
enables the tests.


mrigger:
LGTM, thanks!

PR #350, opened 2021-05-08
Add config tests for ArangoDB

pfu3tz:
With the configuration "withOptimizerRuleTests" it is now possible to test ArangoDB queries further. A randomly generated query is executed with all optimizer flags enabled and with a random number of optimizer flags disabled.


mrigger:
LGTM!

PR #373, opened 2021-07-19
Fix PostgresProvider that use wrong database name

gfphoenix78:
The database name of entryURL is entryDatabaseName, not entryPath.


mrigger (approved):
Thanks a lot for the PRs! Sorry for not responding earlier. I did not receive a notification about them from GitHub (I should probably look into my settings).

PR #375, opened 2021-07-19
Fix compilation failure for some generic parameter types in IDEA

gfphoenix78:
This PR fixes issue #375

### DIFF
--- a/src/sqlancer/arangodb/ArangoDBOptions.java
+++ b/src/sqlancer/arangodb/ArangoDBOptions.java
@@ -11,6 +11,7 @@ import com.beust.jcommander.Parameter;
 import sqlancer.DBMSSpecificOptions;
 import sqlancer.OracleFactory;
 import sqlancer.arangodb.test.ArangoDBQueryPartitioningWhereTester;
+import sqlancer.arangodb.ArangoDBProvider.ArangoDBGlobalState;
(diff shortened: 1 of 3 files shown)


mrigger:
This looks good to me, but causes a Checkstyle error due to the changed lexicographical order:
```
src/sqlancer/arangodb/ArangoDBOptions.java:[14] (imports) CustomImportOrder: Wrong lexicographical order for 'sqlancer.arangodb.ArangoDBProvider.ArangoDBGlobalState' import. Should be before 'sqlancer.arangodb.test.ArangoDBQueryPartitioningWhereTester'.
```
Could you please change the import order so this check passes? You can validate locally by executing `mvn verify -DskipTests=true`.

gfphoenix78:
> This looks good to me, but causes a Checkstyle error due to the changed lexicographical order:
> 
> ```
> src/sqlancer/arangodb/ArangoDBOptions.java:[14] (imports) CustomImportOrder: Wrong lexicographical order for 'sqlancer.arangodb.ArangoDBProvider.ArangoDBGlobalState' import. Should be before 'sqlancer.arangodb.test.ArangoDBQueryPartitioningWhereTester'.
> ```
> 
> Could you please change the import order so this check passes? You can validate locally by executing `mvn verify -DskipTests=true`.

Thanks for your suggestion.

PR #396, opened 2021-09-17
Add support for access methods

halilozanakgul:
This PR adds support for table access methods in Postgres.
A `SELECT` query finds the existing table access methods in the Postgres database and the tables are randomly created as one of those access methods with the `USING` syntax.

### DIFF
--- a/src/sqlancer/postgres/gen/PostgresTableGenerator.java
+++ b/src/sqlancer/postgres/gen/PostgresTableGenerator.java
@@ -23,6 +23,7 @@ public class PostgresTableGenerator {
     private boolean columnHasPrimaryKey;
     private final StringBuilder sb = new StringBuilder();
     private boolean isTemporaryTable;
+    private boolean isPartitionedTable;
     private final PostgresSchema newSchema;
     private final List<PostgresColumn> columnsToBeAdded = new ArrayList<>();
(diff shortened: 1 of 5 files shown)


src/sqlancer/postgres/gen/PostgresTableGenerator.java
```diff
@@ -54,6 +55,8 @@ public PostgresTableGenerator(String tableName, PostgresSchema newSchema, boolea
         errors.add("no collation was derived for partition key column");
         errors.add("inherits from generated column but specifies identity");
         errors.add("inherits from generated column but specifies default");
+        /* this error is specific to Citus extension's columnar implementation */
+        errors.add("unlogged columnar tables are not supported");
```
mrigger, line 59: Perhaps it would make more sense to put this into the `CitusTableGenerator`? See https://github.com/sqlancer/sqlancer/blob/62a76d999f2834925a14b2921c17d34ccdc2ff47/src/sqlancer/citus/gen/CitusTableGenerator.java#L12.

halilozanakgul: I needed to copy `createTables` and `generate` functions of Postgres to achieve this.

mrigger:
Thanks for this PR and for incorporating the feedback!

PR #397, opened 2021-09-17
Add Citus 10.1 support

halilozanakgul:
This PR updates the Citus version from 9.5 to 10.1
The PR also renames the function names that start with `master_` to `citus_` and handles  the new `citus_tables` view to be compatible with Citus 10.1


mrigger:
LGTM. Thanks!

PR #399, opened 2021-09-20
Remove JOIN Oracle in CockroachDB

nehageorge:
This oracle does not necessarily produce the same results
for both queries, thus rendering it an invalid TLP oracle.


mrigger:
Thanks for the PR and for identifying this issue! Also providing the counterexample that you shared with me earlier:

```sql
CREATE TABLE t1(k INT);
INSERT INTO t1(k) VALUES (1), (2);
CREATE TABLE t2(k INT);
INSERT INTO t2(k) VALUES (2), (3);
SELECT * FROM t1 LEFT JOIN t2 ON FALSE UNION ALL
SELECT * FROM t1 LEFT JOIN t2 ON FALSE UNION ALL
SELECT * FROM t1 LEFT JOIN t2 ON TRUE;
SELECT * FROM t1 LEFT JOIN t2 ON t1.k = t2.k UNION ALL
SELECT * FROM t1 LEFT JOIN t2 ON NOT(t1.k = t2.k) UNION ALL
SELECT * FROM t1 LEFT JOIN t2 ON (t1.k = t2.k) IS NULL;
```
Although I do not remember clearly, it might be that I accidentally committed this test oracle even though it looks like a failed experiment.

Issue #333, opened 2021-04-07
Automatic test case reduction


mrigger:
It would be great to implement test case reduction strategies in SQLancer itself. The advantage of doing this in SQLancer, rather than an external tool like C-Reduce, is that we have all the information on whether a specific bug is still triggered (e.g., that two queries should yield the same result) and that the reduction can likely go further than when having an external interestingness test (e.g., since a predicate in the `WHERE` condition of two queries can be reduced simultaneously).

See:
* https://en.wikipedia.org/wiki/Delta_debugging
* https://embed.cs.utah.edu/creduce/

mrigger:
Thanks for your interest and your kind words, @nicdard!

A way would be to start with reducing statements only. I previously prototyped a very hacky implementation, which I just uploaded at https://github.com/sqlancer/sqlancer/commit/66f57c69b40656ba36e96285ac57217b2606cfa3, for SQLite and the NoREC oracle. The core idea is to have a class `FoundBugException` that is thrown once a bug is found and that contains a `Reproducer` which is the interestingness test (i.e., for NoREC comparing whether the optimized and unoptimized queries compute the same result) that can be replayed on the current, possibly reduced, database. The current implementation of the reducer tries to drop statements and some hard-coded keywords. The implementation is hacky because I did not write the main logic of the reducer and logger in a clean way. I hope it is still useful to get started (and feel to copy&paste any code that seems useful). In the `pom.xml` file, I specified an old version of SQLite, so you can conveniently test the reducer.

A good way to go forward would be to first implement such a reducer that removes statements only for a single DBMS, then try to generalize it to multiple DBMSs (e.g., by implementing a common interface), and then progress to reduce also expressions, keywords, etc. A good candidate might be DuckDB, since I believe that the developers might appreciate this (CC @hannesmuehleisen @Mytherin).

For a quick feedback cycle, it might be helpful to create small pull requests (e.g., one for adding the reproducer class and implementing it for an oracle or oracles of a database).

nicdard:
Sorry for the delay, I was under pressure at work, since I am finishing my internship.

Thank you very much for the guidance and the code. I have two questions more:
* how do you apply C-Reduce to reduce the tests? Can you provide an example?
[...]

PR #404, opened 2021-10-01
Bootstrap automatic test case reduction.

nicdard:
Partially fix: https://github.com/sqlancer/sqlancer/issues/333

Using the reference implementation given [here](https://github.com/sqlancer/sqlancer/commit/66f57c69b40656ba36e96285ac57217b2606cfa3) I have created a generified first version for integrated test case reduction, as discussed offline with @mrigger. It is using the same fixpoint iteration as showed in the code linked.

Also, I have refactored (only partially) the code for `DuckDbNoRECOracle` to apply the reduction. Actually NoREC oracles can be refactored and I believe much more code reusability can be achieved.

This is only the starting point: I have tried to divide the actual reduction operation from logging, however I am not super happy as it is now, maybe we can improve also this part more in the future. The problem that I see is that the state is stored and passed from one method to the other. Modifying this whould require a great effort and it should be properly discussed I believe, so for now I just kept it this way, trying to separate filling the state object at runtime from actually writing the statements to file.

### DIFF
--- a/src/sqlancer/DatabaseProvider.java
+++ b/src/sqlancer/DatabaseProvider.java
@@ -30,6 +30,21 @@ public interface DatabaseProvider<G extends GlobalState<O, ?, C>, O extends DBMS
      */
     void generateAndTestDatabase(G globalState) throws Exception;
 
+    /**
+     * Reduce a state which triggers an error in a test oracle.
+     *
+     * @param e
+     *          the exception carrying on the error.
+     * @param stateToReduce
+     *          the state to be reduced.
+     * @param newGlobalState
+     *          a copy state where to apply reduction.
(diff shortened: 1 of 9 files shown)


mrigger:
Thanks a lot for the PR and your work on this, and sorry for the long-overdue review! I think it would actually be useful to split this PR into multiple parts, starting with the `Reproducer` logic (i.e., adding the interface and applying it in DuckDB). We could use the reproducers already now to check whether a bug initially triggers (not only in the reducer to check whether a bug is still triggered).

mrigger:
Closing this PR due to inactivity and https://github.com/sqlancer/sqlancer/pull/597.

PR #428, opened 2021-12-09
Fix union operator

entropaos:


mrigger:
Thanks for spotting this and creating the PR! It seems that many queries now fail with a `result of range union would not be contiguous` error. To count this as expected behavior, we need to add the string as an expected error message (e.g., at https://github.com/sqlancer/sqlancer/blob/ce0e951b46e0f2c4974204c447662ac40d610521/src/sqlancer/postgres/gen/PostgresCommon.java#L25). Could you please do that?

entropaos:
> 

@mrigger  As the method name suggested, I add the known error to `addCommonRangeExpressionErrors()`. It should make sense.

mrigger:
Thanks for the PR! The H2 test is failing due to an unrelated issue, so I'm merging this now.

Issue #8, opened 2020-06-02
Plugin system for the DBMS components


mrigger:
We should make the SQLancer core independent from the DBMS implementations using a plugin system (e.g., by using the [ServiceLoader infrastructure](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html)).

PR #458, opened 2022-01-19
Use a ServiceLoader to pick up database testing implementations

mrigger:
This change allows SQLancer to pick up database providers in other JARs than the SQLancer JAR, if they are on the classpath.


mrigger:
See https://github.com/sqlancer/sqlancer/issues/8.

PR #484, opened 2022-03-25
rename variables

ghost62:
Renamed some variables to improve readability.
Files: sqlancer->cockroachdb-> CockroachDBProvider.java, sqlancer->mariadb->ast-> MariaDBBinaryOperator


mrigger:
Thanks for the contribution! I believe that this does not obviously improve readability and even if, there are also other locations in the source code where this should be done for consistency as well. I'm thus closing this PR.

PR #485, opened 2022-03-25
extracted method

ghost62:
Extracted query builder to a new method from the create method


mrigger:
Thanks for the PR! However, since I don't see how this improves the code base, I'm closing it.

PR #486, opened 2022-03-25
refactor2 move method extract class push down method

ghost62:
Moved method to a different class that demolished feature envy smell, extracted class to reduce 1 multifaceted abstraction and pushed down method to remove Rebellious Hierarchy.

PR is requested as part of a University assignment on refactoring.


ghost62:
Thanks for the feedback. I will try again in a better way.

PR #492, opened 2022-04-01
Fix incorrect exit code and possible data race of shutdownThreads

gavinchou:
* if --max-generated-database is set to a positive number and run without error,
  the exit code should be 0 instead of the one set in MainOptions
* shutdownThreads should be atomic due to multi-thread access to it


mrigger:
Thanks a lot for the PR and great catch! It seems the CI currently fails with a formatting error. Could you please use `mvn verify -DskipTests=true` to reproduce this issue locally and fix it (using `git commit --amend`)?

mrigger:
LGTM, thanks!

PR #497, opened 2022-04-18
Update `DuckDB` SQLancer implementation

kokrui:
Simple PR that bumps DuckDB to its latest version 0.3.3, and updates the implementation based on the SQLancer fork used by the DuckDB team (specifically, the `persistent` branch which is used in their CI).

Also enables in SQLancer CI tests for DuckDB again (specifically, with the query partitioning oracle which is DuckDB's default oracle)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -302,7 +302,7 @@
     <dependency>
       <groupId>org.duckdb</groupId>
       <artifactId>duckdb_jdbc</artifactId>
-      <version>0.2.5</version>
+      <version>0.3.3</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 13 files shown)


mrigger (approved):
Great job, the changes look good to me! Do you have an idea what causes the CI failures? Not sure if the mismatch between 0.0 and -0.0 indicates a bug or requires a canonicalization rule (e.g., to convert -0.0 to 0.0 when comparing floating-point zeros).

kokrui:
Yeah, will investigate a bit today!

mrigger (approved):
LGTM!

PR #500, opened 2022-04-22
Allow result set canonicalization before equality check

kokrui:
Implements an overloaded version of `ComparatorHelper.assumeResultSetsAreEqual` that takes in an additional parameter, `canonicalizationRule`, which is a function that all elements of both result sets are mapped through before they are checked for equality.

Additionally, implements tests for the `assumeResultSetsAreEqual` function!

Inspired by the need for DuckDB-specific canonicalization noticed while implementing #497

### DIFF
--- a/test/sqlancer/TestComparatorHelper.java
+++ b/test/sqlancer/TestComparatorHelper.java
@@ -1,0 +1,54 @@
+package sqlancer;
+
+import static org.junit.jupiter.api.Assertions.assertThrowsExactly;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+public class TestComparatorHelper {
+    // TODO: Implement tests for the other ComparatorHelper methods
+
+    @Test
+    public void testAssumeResultSetsAreEqualWithEqualSets() {
+        List<String> r1 = Arrays.asList("a", "b", "c");
+        List<String> r2 = Arrays.asList("a", "b", "c");
+        ComparatorHelper.assumeResultSetsAreEqual(r1, r2, "", Arrays.asList(""), null);
+
+    }
+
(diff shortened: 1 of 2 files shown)


test/sqlancer/TestComparatorHelper.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Test
+    public void testAssumeResultSetsAreEqualWithUnequalLengthSets() {
+        List<String> r1 = Arrays.asList("a", "b", "c");
+        List<String> r2 = Arrays.asList("a", "b", "c", "d", "g");
+        // NullPointerException is raised instead of AssertionError because state is null and the state.getState()...
+        // line occurs before AssertionError is thrown, but it's good enough as an indicator that one of the Exceptions
+        // is raised
+        assertThrowsExactly(NullPointerException.class, () -> {
```
mrigger, line 28: I feel like checking for any kind of `Throwable` would be more general here, but specifically checking for a `NullPointerException` is also fine.

mrigger (approved):
The changes look good to me! Ideally, you could rewrite the commit history by integrating the changes from the two typo fixes within the other two commits (e.g., using `git rebase -i`), but we could also go ahead and merge the current version.

kokrui:
will rebase and merge!

test/sqlancer/TestComparatorHelper.java
mrigger: I see! It might still be debatable whether this is a part of the method's contract (i.e., its API) or an implementation artifact. Either case is fine - at most, we have to update the unit test if the implementation ever changes.

kokrui: ah, makes sense!

PR #514, opened 2022-05-28
Add an expected error on usage of the chr() routine.

danolivo:
According to PG15 commit e9e63b7, chr() in next release will return an error
on negative input. Be prepared to that change.


mrigger:
LGTM. Thanks!

PR #524, opened 2022-06-16
Swap parameters of Untyped generateExpressions to match Typed

kokrui:
Quick refactor of UntypedExpressionGenerator's `generateExpressions` method so that it's paramaters go `(int nr, int depth)` to keep it consistent with TypedExpressionGenerator which does `(T type, int nr, int depth)`


mrigger (approved):
LGTM!

PR #540, opened 2022-08-01
disable EXTRA_FLOAT_DIGITS

bajinsheng:
We disable the EXTRA_FLOAT_DIGITS to avoid the false positive as follows:
 
```sql
CREATE TABLE t1 (c0 BOOL);
SET SESSION EXTRA_FLOAT_DIGITS=-15;
UPSERT INTO t1 (c0) VALUES(false);

SELECT SUM(0.57 :::FLOAT); -- 0.6
SELECT SUM(0.57 :::FLOAT)::DECIMAL; -- 0.57000000000000000000
```


mrigger:
For documentation: The test oracle for testing the `AVG` uses a cast to `DECIMAL`.

PR #573, opened 2022-10-04
Added wrapped exceptions support

qvad:
Some database drivers throws exceptions that may be wrapped by a high-level one. In this case exception text that is actually allowed will be treated as an error.

New code walks through all existing causes and search for the text there.

### DIFF
--- a/src/sqlancer/common/query/SQLQueryAdapter.java
+++ b/src/sqlancer/common/query/SQLQueryAdapter.java
@@ -98,17 +98,15 @@ public class SQLQueryAdapter extends Query<SQLConnection> {
     public void checkException(Exception e) throws AssertionError {
         Throwable ex = e;
 
-        while (true) {
+        while (ex != null) {
             if (expectedErrors.errorIsExpected(ex.getMessage())) {
                 return;
-            } else if (ex.getCause() != null) {
-                ex = ex.getCause();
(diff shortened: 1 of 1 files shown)


src/sqlancer/common/query/SQLQueryAdapter.java
```diff
@@ -96,9 +96,19 @@ public String getUnterminatedQueryString() {
     }
 
     public void checkException(Exception e) throws AssertionError {
-        if (!expectedErrors.errorIsExpected(e.getMessage())) {
-            throw new AssertionError(query, e);
+        Throwable ex = e;
+
+        while (true) {
```
mrigger, line 101: Thanks for the contribution. I think the code should be easier to understand if you use `ex.getCause() != null` rather than `true` in the `while` condition, which would also simplify the `else if` to an `else`. Or would this yield an unexpected result?

qvad: Agreed, it's overcomplicated a bit :) Thanks

This is simplified version
```
Throwable ex = e;

while (ex != null) {
    if (expectedErrors.errorIsExpected(ex.getMessage())) {
        return;
    } else {
        ex = ex.getCause();
    }
}

throw new AssertionError(query, e);
```

PR #586, opened 2022-10-10
test: reproduce ci test errors

hanyisong:
Signed-off-by: Yisong Han <yisong8686@gmail.com>


mrigger:
Okay, I see, it seems that the Eclipse configuration file is needed, so we cannot delete/modify it for now.

mrigger:
Not sure if there is any good solution or common practice that we could implement?

PR #587, opened 2022-10-10
Update CockroachDB to v22.2.0-beta.4

mgartner:
This is currently failing due to https://github.com/cockroachdb/cockroach/issues/89024, but it should be fixed in the next beta or RC. I'll update this PR when that is release.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,11 +109,11 @@ jobs:
         run: mvn -B package -DskipTests=true
       - name: Set up CockroachDB
         run: |
-          wget -qO- https://binaries.cockroachdb.com/cockroach-v21.1.7.linux-amd64.tgz | tar  xvz
-          cd cockroach-v21.1.7.linux-amd64/ && ./cockroach start-single-node --insecure &
+          wget -qO- https://binaries.cockroachdb.com/cockroach-v22.2.0-beta.4.linux-amd64.tgz | tar  xvz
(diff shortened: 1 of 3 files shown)


mrigger:
Great, thanks! We can merge it as soon as the CI tests pass.

qoega:
Probably you want to just ignore this errors? 

```
EXPLAIN SELECT VARIANCE(((359867990)/(1513271258))), MAX(((' ?') COLLATE de)), MAX(TIMESTAMP '1969-12-08T08:41:57') FROM t72, t1 WHERE (TIMESTAMP '1970-01-22T10:04:39') BETWEEN SYMMETRIC (t1.c1) AND (t1.c1) GROUP BY t72.c0 HAVING MAX(true) ORDER BY ((-832926214)*(1490194401)), t72.c0 LIMIT NULL OFFSET 457769557;
```

```
org.postgresql.util.PSQLException: ERROR: internal error: expected required columns to be a subset of output columns
[526](https://github.com/sqlancer/sqlancer/actions/runs/3222424613/jobs/5271481266#step:7:527)
```

mgartner:
Thanks for pointing that error out. It's an internal error and an indication that there is something wrong with a query plan, so we don't want to ignore it. It could be the same problem as https://github.com/cockroachdb/cockroach/issues/88037. We will investigate.

mrigger:
Is this PR ready to be merged? If yes, please rebase or merge so that the PR is up-to-date with the base branch.

mgartner:
Yes, I think this is ready to merge.

mrigger:
Great, thanks!

PR #588, opened 2022-10-12
Add Provider and Schema for QuestDB

SuriZhang:
Add Provider and Schema classes for QuestDB.
Generators are only skeletons for now. They will be implemented in the upcoming PRs.

QuestDB is connected via PostgreSQL JDBC with default setting: https://questdb.io/docs/reference/api/postgres/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -347,6 +347,11 @@
       <artifactId>arangodb-java-driver</artifactId>
       <version>6.9.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.questdb</groupId>
+      <artifactId>questdb</artifactId>
+      <version>6.5.3</version>
+    </dependency>
   </dependencies>
   <reporting>
     <plugins>
--- a/src/sqlancer/questdb/QuestDBOptions.java
+++ b/src/sqlancer/questdb/QuestDBOptions.java
@@ -1,0 +1,55 @@
+package sqlancer.questdb;
+
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.List;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+
+import sqlancer.DBMSSpecificOptions;
+import sqlancer.OracleFactory;
+import sqlancer.common.oracle.TestOracle;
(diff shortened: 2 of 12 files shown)


pom.xml
```diff
@@ -347,6 +347,11 @@
       <artifactId>arangodb-java-driver</artifactId>
       <version>6.9.0</version>
     </dependency>
+    <dependency>
```
mrigger, line 350: Perhaps in this PR, you could only add the questdb dependency. Updating the other dependencies might have some unintended side effects (e.g., CI test failures).

src/sqlancer/questdb/QuestDBOptions.java
```diff
@@ -0,0 +1,112 @@
...
+        + " default host: " + QuestDBOptions.DEFAULT_HOST + ")")
+public class QuestDBOptions implements DBMSSpecificOptions<QuestDBOracleFactory> {
+    public static final String DEFAULT_HOST = "localhost";
+    public static final int DEFAULT_PORT = 8812;
+
+    @Parameter(names = "--oracle")
+    public QuestDBOracleFactory oracle = QuestDBOracleFactory.NOREC;
+
+    public enum QuestDBOracleFactory implements OracleFactory<QuestDBGlobalState> {
+        NOREC {
```
mrigger, line 27: Perhaps you can keep the enum empty for now? Later on, you can decide if you want to add one or multiple oracles.

mrigger (commented):
Thanks a lot for the PR. Could you take a look at the minor comments, and amend the current commit with the changes? The CI tests fail due to some errors that you can reproduce with `mvn verify -DskipTests=True`

pom.xml
mrigger: It seems all the checks pass, so it might be fine to keep the updates.

SuriZhang: CI test was the only one failed. Anyways I reverted back the version bump.

src/sqlancer/questdb/QuestDBOptions.java
SuriZhang: Yes, fixed this : )

PR #591, opened 2022-10-17
Add DuckDB canonicalization rule for `-0` -> `0`

kokrui:
While testing SQLancer on DuckDB, noticed from some errors that the previous implementation only canonicalized double/float negative zeros -- this rule adds the same rule for integers.


mrigger (approved):
LGTM, thanks! Could you please rebase after the Databend CI flakines fix?

PR #592, opened 2022-10-17
Add expected DuckDB timestamp errors

kokrui:
DuckDB added new error messages related to the subtraction of timestamps. Our existing DuckDB expression generator can generate expressions involving these errors, so I'm adding them in as expected errors!

Also, removing some expected errors that should no longer be expected and thus are obsolete


mrigger (approved):
LGTM, thanks! Could you please rebase?

PR #595, opened 2022-10-19
fix: don't pass username or password to duckdb

Mause:
It's an inmemory database and does not support them

### DIFF
--- a/src/sqlancer/MainOptions.java
+++ b/src/sqlancer/MainOptions.java
@@ -5,9 +5,12 @@ import com.beust.jcommander.Parameters;
 
 import sqlancer.Randomly.StringGenerationStrategy;
 
+import java.util.Objects;
+
 @Parameters(separators = "=", commandDescription = "Options applicable to all DBMS")
 public class MainOptions {
     public static final int NO_SET_PORT = -1;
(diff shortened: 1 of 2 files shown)


mrigger:
Thanks a lot for the PR! We can either merge as is, or you could consider the following improvement suggestion: we could throw an exception if a username or password is passed before we establish the connection. In order to check this reliably, you could introduce two new methods `isDefaultPassword` and `isDefaultUsername` in the `MainOption` class that check the respective fields (e.g., https://github.com/sqlancer/sqlancer/blob/0d61b0d23b5d5711fbc0b07013e54b93586cfc86/src/sqlancer/MainOptions.java#L51) for their default values. Should we just merge this PR, or would you like to implement this suggestion?

Mause:
@mrigger I had a go at implementing that, let me know if I've understood you correctly

mrigger:
Thanks, that's an even more elegant solution than I had in mind. :) We can merge once the format checks pass (`mvn verify -DskipTests=True`).

Mause:
Thanks for the fast turnaround!

mrigger:
Likewise, thanks a lot!

PR #597, opened 2022-10-21
Starting architecture for using a reducer.

kshehata:
Hey Manuel! I've started going through the architecture for adding reducer support on this branch. Would you please take a look and let me know if this is heading in the right direction?

Here's what I've done so far:

1. Used the FoundBugException from the previous PRs for capturing state and checking if the bug reproduces.
2. Added an interface for Reducer support. Idea is that eventually that should be made more generic.
3. Moved all of the reducing code that was in Main into an example reducer file. The idea is to eventually replace that with something more robust.

I'm thinking that before merging this should probably be hidden behind a command line flag, but just wanted to get your feedback at this stage.

Thanks!

### DIFF
--- a/src/sqlancer/FoundBugException.java
+++ b/src/sqlancer/FoundBugException.java
@@ -3,17 +3,16 @@ package sqlancer;
 import sqlancer.sqlite3.SQLite3GlobalState;
 
 public class FoundBugException extends RuntimeException {
+    private static final long serialVersionUID = 1L;
+    private final Reproducer reproducer;
 
     public interface Reproducer {
-        public abstract boolean bugStillTriggers(SQLite3GlobalState globalState);
+        boolean bugStillTriggers(SQLite3GlobalState globalState);
 
-        public default void outputHook(SQLite3GlobalState globalState) {
+        default void outputHook(SQLite3GlobalState globalState) {
 
         }
     }
-
-    private static final long serialVersionUID = 1L;
-    private Reproducer reproducer;
 
     public FoundBugException(String string, Reproducer reproducer) {
         super(string);
--- a/src/sqlancer/FoundBugException.java
+++ b/src/sqlancer/FoundBugException.java
@@ -1,0 +1,27 @@
+package sqlancer;
+
+import sqlancer.sqlite3.SQLite3GlobalState;
+
+public class FoundBugException extends RuntimeException {
+
+    public interface Reproducer {
+        public abstract boolean bugStillTriggers(SQLite3GlobalState globalState);
+
+        public default void outputHook(SQLite3GlobalState globalState) {
+
+        }
+    }
+
+    private static final long serialVersionUID = 1L;
+    private Reproducer reproducer;
+
+    public FoundBugException(String string, Reproducer reproducer) {
+        super(string);
+        this.reproducer = reproducer;
+    }
+
+    public Reproducer getReproducer() {
+        return reproducer;
+    }
+
+}
--- a/src/sqlancer/Main.java
+++ b/src/sqlancer/Main.java
@@ -50,8 +50,6 @@ public final class Main {
     public static final class StateLogger {
 
         private final File loggerFile;
-        private final File reducedFile;
-
         private File curFile;
         private FileWriter logFileWriter;
         public FileWriter currentFileWriter;
@@ -85,7 +83,6 @@ public final class Main {
             }
             ensureExistsAndIsEmpty(dir, provider);
             loggerFile = new File(dir, databaseName + ".log");
-            reducedFile = new File(dir, databaseName + "-reduced.log");
             logEachSelect = options.logEachSelect();
             if (logEachSelect) {
                 curFile = new File(dir, databaseName + "-cur.log");
@@ -123,15 +120,6 @@ public final class Main {
                 } catch (IOException e) {
                     throw new AssertionError(e);
                 }
-            }
-            return logFileWriter;
-        }
-
-        private FileWriter getReducedWriter() {
-            try {
-                logFileWriter = new FileWriter(reducedFile);
-            } catch (IOException e) {
-                throw new AssertionError(e);
             }
             return logFileWriter;
         }
@@ -181,16 +169,6 @@ public final class Main {
                 currentFileWriter.flush();
             } catch (IOException e) {
                 throw new AssertionError();
-            }
-        }
-
-        public void logReduced(StateToReproduce state) {
-            FileWriter logFileWriter = getReducedWriter();
-            printState(logFileWriter, state);
-            try {
-                logFileWriter.close();
-            } catch (IOException e) {
-                throw new AssertionError(e);
             }
         }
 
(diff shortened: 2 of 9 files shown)


src/sqlancer/Main.java
```diff
@@ -124,6 +127,17 @@ private FileWriter getLogFileWriter() {
             return logFileWriter;
         }
 
+        private FileWriter getReducedWriter() {
+            // if (logFileWriter == null) {
```
mrigger, line 131: perhaps you can remove the comment?

src/sqlancer/Main.java
```diff
@@ -172,6 +186,17 @@ private void write(Loggable loggable) {
             }
         }
 
+        public void logReduced(StateToReproduce state) {
+            FileWriter logFileWriter = getReducedWriter();
+            printState(logFileWriter, state);
+            try {
+                logFileWriter.close();
+            } catch (IOException e) {
+                // TODO Auto-generated catch block
```
mrigger, line 195: here, we could throw an `AssertionError` instead.

src/sqlancer/Main.java
```diff
@@ -312,30 +345,30 @@ public void run() throws Exception {
...
                 try {
-                    logger.getCurrentFileWriter().close();
-                    logger.currentFileWriter = null;
-                } catch (IOException e) {
-                    throw new AssertionError(e);
+                    provider.generateAndTestDatabase(state);
+
+                } catch (FoundBugException e) {
+                    try {
+                        logger.getCurrentFileWriter().close();
```
mrigger, line 353: This seems quite hacky (perhaps part of my initial prototype implementation), not sure if there is a better way to implement this functionality.

src/sqlancer/Main.java
```diff
@@ -490,6 +523,7 @@ private boolean run(MainOptions options, ExecutorService execService,
                     DBMSExecutor<?, ?, ?> executor = executorFactory.getDBMSExecutor(databaseName, r);
                     try {
                         executor.run();
+                        System.out.println("Executor completed!");
```
mrigger, line 526: I think we should not check-in this line.

src/sqlancer/ReducerExample.java
```diff
@@ -0,0 +1,120 @@
...
+        do {
+            observedChange = false;
+            knownToReproduceBugStatements = tryReduction(state, reproducer, newGlobalState,
+            knownToReproduceBugStatements, (candidateStatements, i) -> {
+                candidateStatements.remove((int) i);
+                return true;
+            });
+        } while (observedChange);
+
+        for (String s : TOKENS) {
```
mrigger, line 48: For an initial implementation, I'd suggest keeping it general and simple, by only removing complete statements. This might simplify the implementation.

mrigger (commented):
Thanks, Kareem! This PR definitely goes in the right direction.

I agree that it would be a good idea to hide things behind an option flag.

What do you think of a PR that adds the essential classes and interfaces (like `FoundBugException` or `Reducer`) and adopts the `FoundBugException` without actually changing any of the visible behavior, and then a follow-up PR that adopts the new classes and the reducer? I believe we could merge the first PR rather quickly.

To avoid the formatting-check error, you can execute `mvn formatter:format` (and perhaps check whether `mvn verify -DskipTests=True` results in any errors.

src/sqlancer/Main.java
kshehata: Removed.

src/sqlancer/Main.java
kshehata: Removed.

src/sqlancer/Main.java
kshehata: TBH, I don't understand enough of how Main works. This is taken from your original prototype, and I'm open to suggestions for how to do it better!

src/sqlancer/Main.java
kshehata: Sorry about that, removed!

src/sqlancer/ReducerExample.java
kshehata: Done.

kshehata:
Sure thing! I've removed any externally visible changes that I can see and hidden it behind a command line flag. I've also fixed the formatting and PMD problems. Is this simple enough to move forward with, or is there more that needs to be deferred?

src/sqlancer/FoundBugException.java
```diff
@@ -0,0 +1,26 @@
+package sqlancer;
+
+import sqlancer.sqlite3.SQLite3GlobalState;
+
+public class FoundBugException extends RuntimeException {
+    private static final long serialVersionUID = 1L;
+    private final Reproducer reproducer;
+
+    public interface Reproducer {
+        boolean bugStillTriggers(SQLite3GlobalState globalState);
```
mrigger, line 10: Just noticed that this is SQLite-specific. Would it be possible to use generics here?

src/sqlancer/ReducerExample.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
+import sqlancer.FoundBugException.Reproducer;
+import sqlancer.common.query.Query;
+import sqlancer.sqlite3.SQLite3GlobalState;
+
+public class ReducerExample<G extends GlobalState<O, ?, C>, O extends DBMSSpecificOptions<?>, C extends SQLancerDBConnection>
```
mrigger, line 12: Perhaps `SimpleReducer` or `StatementReducer` is a better name? No need to address this in this PR.

src/sqlancer/sqlite3/gen/ddl/SQLite3IndexGenerator.java
```diff
@@ -44,6 +44,7 @@ private SQLQueryAdapter create() throws SQLException {
         errors.add("non-deterministic use of date() in an index");
         errors.add("non-deterministic use of datetime() in an index");
         errors.add("The database file is locked");
+        errors.add("Abort due to constraint violation");
```
mrigger, line 47: Is this added error intended? I suspect that this might be for an older or more recent version of SQLite that we are currently testing?

mrigger (commented):
Great, thanks! Could have a look at the comments (the one about the change in the index generator and SQLite-specific class in the `Reproducer` interface), and see if it is possible to address them easily or not? After that, we can merge the PR - please rebase on the latest master branch.

For code style violations in future PRs, it would be great if you could amend the changes to the commit that introduced them (see https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#commit-history).

src/sqlancer/ReducerExample.java
kshehata: Done.

src/sqlancer/sqlite3/gen/ddl/SQLite3IndexGenerator.java
kshehata: No, I think this was an artifact of a merge or something previous. I've removed it.

src/sqlancer/FoundBugException.java
kshehata: I'm trying to replace with the generics now but I'm getting all kinds of weird errors. I'll keep working on it, but for now let's assume this is fixed and review everything else.

kshehata:
> Great, thanks! Could have a look at the comments (the one about the change in the index generator and SQLite-specific class in the `Reproducer` interface), and see if it is possible to address them easily or not? After that, we can merge the PR - please rebase on the latest master branch.

I've rebased to master (BTW: consider [renaming to main](https://www.zdnet.com/article/github-to-replace-master-with-main-starting-next-month/) when you get a chance). I'm working on the generics now, but let's move ahead with the review to make sure there are no other issues that need to be addressed.

> For code style violations in future PRs, it would be great if you could amend the changes to the commit that introduced them (see https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#commit-history).

I'm not sure I understand. You'd like me to modify a previous commit to address a comment? But then how would you see what changes I've made since you last reviewed, wouldn't that make more work for you? Why not just squash the whole branch before merging?

src/sqlancer/FoundBugException.java
kshehata: Just realized a fatal problem in our design: [Throwables cannot be Generic](https://stackoverflow.com/questions/501277/why-doesnt-java-allow-generic-subclasses-of-throwable). As such, I don't think we can keep a reference to a generic Reproducer without throwing away the generic type info, which is problematic. I think we'll need to rethink how this all works if we're to make this properly generic, and probably move away from using Exceptions for finding bugs.

mrigger:
> > Great, thanks! Could have a look at the comments (the one about the change in the index generator and SQLite-specific class in the `Reproducer` interface), and see if it is possible to address them easily or not? After that, we can merge the PR - please rebase on the latest master branch.
> 
> I've rebased to master (BTW: consider [renaming to main](https://www.zdnet.com/article/github-to-replace-master-with-main-starting-next-month/) when you get a chance). I'm working on the generics now, but let's move ahead with the review to make sure there are no other issues that need to be addressed.

Yes, performing the rename has been on my todo list for a while now. Should we merge the PR for now, or would you like to first address this properly?

> 
> > For code style violations in future PRs, it would be great if you could amend the changes to the commit that introduced them (see https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#commit-history).
> 
> I'm not sure I understand. You'd like me to modify a previous commit to address a comment? But then how would you see what changes I've made since you last reviewed, wouldn't that make more work for you? Why not just squash the whole branch before merging?

I was referring to the code style violations. Squashing as an alternative also sounds good.

mrigger (approved):
LGTM! Would you like that we merge the PR now, or after we found a way to address the generics issue? Either is fine with me.

PR #602, opened 2022-10-26
Implement generators and ast for questdb

SuriZhang:
A couple generators are implemented in this PR along with ASTs.
In this version, 
- only INT, FLOAT, BOOLEAN and NULL types are supported;
- CREATE TABLE, TRUNCATE TABLE and ALTER TABLE (only add or drop index) are tested;
- 1 bug report has been filed to QuestDB https://github.com/questdb/questdb/issues/2689


src/sqlancer/questdb/gen/QuestDBTableGenerator.java
```diff
@@ -27,13 +35,24 @@ public SQLQueryAdapter getQuery(QuestDBGlobalState globalState) {
             sb.append(" ");
             sb.append(columns.get(i).getType());
         }
-        sb.append(");");
+        sb.append(")");
+        // test index at Create Table - bug report already submitted
```
mrigger, line 39: Could you please adopt the pattern described at https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#unfixed-bugs to ignore unfixed bugs?

mrigger (commented):
Almost ready to merge. Could you please have a look at my inline comment?

src/sqlancer/questdb/gen/QuestDBTableGenerator.java
SuriZhang: fixed

PR #605, opened 2022-11-03
Implement more expressions and add SELECT, WHERE to test

SuriZhang:
- Implement Unary/Binary operator generators for testing
- Implement `SELECT` visitor
- Add expected errors found when running sqlancer against QuestDB 
- Fix `Inf` String representation in Double

### DIFF
--- a/src/sqlancer/questdb/QuestDBBugs.java
+++ b/src/sqlancer/questdb/QuestDBBugs.java
@@ -2,9 +2,6 @@ package sqlancer.questdb;
 
 public final class QuestDBBugs {
 
-    // https://github.com/questdb/questdb/issues/2689
-    public static boolean bug2689 = true;
-
     private QuestDBBugs() {
     }
 }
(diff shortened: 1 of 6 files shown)


src/sqlancer/questdb/QuestDBBugs.java
```diff
@@ -3,7 +3,8 @@
 public final class QuestDBBugs {
 
     // https://github.com/questdb/questdb/issues/2689
-    public static boolean bug2689 = true;
+    // set this to false since this error breaks db metadata and interrupts all succeeding query exeuctions
+    public static boolean bug2689;
```
mrigger, line 7: https://github.com/questdb/questdb/issues/2689 It seems that the bug was fixed, so we can remove this flag?

mrigger (approved):
LGTM! Do we still need the bug flag for `bug2689`? It seems this issue was already fixed (see https://github.com/questdb/questdb/issues/2689). Could you please rebase or merge based on the latest master so we can merge the PR?

PR #606, opened 2022-11-04
Add QuestDB to supported DBMS

SuriZhang:
Add QuestDB to supported DBMS in README


mrigger (approved):
LGTM, thanks! If you have time, it would be great if you could also add a CI test to ensure that your implementation does not break during future refactorings (e.g., see https://github.com/sqlancer/sqlancer/blob/master/.github/workflows/main.yml#L97).

PR #607, opened 2022-11-04
[ClickHouse] Updated clickhouse-jdbc and reworked expression generation

qoega:
- Added functions support. 
- Added tables aliases that change how column is addressed - required for JOINs etc.
- Added JOIN clause generation.
- Added tests for Operators visitor
- Updated clickhouse-jdbc driver
- Added support for Bool type

PS. Almost rewrote everything as there was really low probability to generate correct JOIN ON clause and table aliases were not used for column names.
Simplest example is something like
``` 
SELECT left.c0, right.c0
FROM t0 AS left
INNER JOIN t0 as right
ON left.c1 = right.c1
```
Here you in `fetchColumns` I needed table alias. And ON is not a simple expression, even not a binary comparison with `EQUALS` as there single `=` is used and left and right sides should use columns from different columns. 
It become even worse when you have several JOINs - you can use any column from previous joined tables, but it is better to not lose table alias anyway.


mrigger (approved):
LGTM, thanks a lot!

PR #624, opened 2022-11-28
Add Reducer capabilities by introducing a Reproducer that can be retrieved from TestOracle

kshehata:
Quick preview of what I'm working on. This adds the type information to return a Reproducer, but doesn't modify the `check()` method. Is this heading in the right direction? If so, I can add the Reducer changes tomorrow.

### DIFF
--- a/src/sqlancer/StateToReproduce.java
+++ b/src/sqlancer/StateToReproduce.java
@@ -9,7 +9,7 @@ import sqlancer.common.query.Query;
 
 public class StateToReproduce {
 
-    private final List<Query<?>> statements = new ArrayList<>();
+    private List<Query<?>> statements = new ArrayList<>();
 
     private final String databaseName;
 
@@ -126,6 +126,10 @@ public class StateToReproduce {
 
     public OracleRunReproductionState createLocalState() {
         return new OracleRunReproductionState();
+    }
+
+    public void setStatements(List<Query<?>> statements) {
+        this.statements = statements;
     }
 
 }
--- a/src/sqlancer/DatabaseProvider.java
+++ b/src/sqlancer/DatabaseProvider.java
@@ -24,6 +24,8 @@ public interface DatabaseProvider<G extends GlobalState<O, ?, C>, O extends DBMS
      * @param globalState
      *            the state created and is valid for this method call.
      *
+     * @return Reproducer if a bug is found and a reproducer is available.
+     *
      * @throws Exception
      *             if creating the database fails.
      *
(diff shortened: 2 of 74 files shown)


mrigger:
Looks good, thanks!

mrigger:
Currently, the tests are failing to the newly-added HSQLDB implementation.

kshehata:
Extended this with Reproducer support in SQLite3's NoREC Oracle and the StatementReducer to use it. PTAL and let me know if this looks good so far?

I'm trying to figure out now how to connect the two. The problem I'm facing is that Main has what is needed to make a new GlobalState for the Reducer to use, but ProviderAdapter has the Reproducer. Here are my ideas so far:

1. Have ProviderAdapter catch the AssertionError and return the reproducer if it exists, or rethrow if it doesn't.
2. Make another call to ProviderAdapter for reducing the query.
3. Make ProviderAdapter cache the Reproducer and return it in a separate call. This seems not great because that means the Reproducer is being cached twice.
4. Have ProviderAdapter clone the GlobalState. Is that possible?

Let me know what you think. I'm less familiar with this part of the design, so I'm not sure what the best choice is.

src/sqlancer/StateToReproduce.java
```diff
@@ -128,4 +128,8 @@ public OracleRunReproductionState createLocalState() {
         return new OracleRunReproductionState();
     }
 
+    public void setStatements(List<Query<?>> statements) {
+        this.statements = statements;
```
mrigger, line 132: Perhaps it might be safer to defensively copy the list?

mrigger (commented):
I'm not very sure. I think it would be reasonable to implement a `copy` method in `GlobalState`. Perhaps some care would be required to ensure that we deep-copy `StateToReproduce`. The other options also sound reasonable. Perhaps you can go ahead with the method you feel is most appropriate?

Btw, I'm boarding a long-distance flight and there might be some delays with my responses in the next few days. Sorry for that!

kshehata:
I've updated ProviderAdapter to catch AssertionError so it can pull down the Reproducer and return it to Main. Main then uses similar logic to the previous PR to do the reducing. I've also rebased to master. It's hidden behind a flag and doesn't change any of the actual output logs, just outputs to stdout. Is this now at a good state to merge? I'll squash before merging.

mrigger (approved):
Thanks, LGTM! I will use the GitHub user interface to update the branch and squash the commits.

mrigger:
Sorry, it seems that due to the new HSQLDB component, the build currently fails.

kshehata:
I've fixed HSQLDB and it should pass all tests now. Please go ahead with the squash and merge when possible. Thanks!

kshehata:
Test seem to be failing as apt-get cannot find Citus ?

`Unable to locate package postgresql-13-citus-10.1`

mrigger:
Yes, merging the PR despite the failure.

@halilozanakgul: could you please have a look at the Citus issue?

kshehata:
Fantastic, thanks!

PR #625, opened 2022-11-29
Implements expressions for select clause in HSQLDB

nish-d:
Fixed some Select and where clauses. Implemented Update and order by statements.


mrigger (approved):
LGTM, thanks!

PR #629, opened 2022-12-06
Support QPG

bajinsheng:
The basic framework of QPG.
Todo: Update documents after 9 Dec.


src/sqlancer/Main.java
```diff
@@ -211,6 +243,13 @@ private void printState(FileWriter writer, StateToReproduce state) {
             }
         }
 
+        private String processQueryPlan(String queryPlan) {
```
mrigger, line 246: Could we use a more insightful name here, like `removeNamesFromQueryPlans` or so?

src/sqlancer/DatabaseProvider.java
```diff
@@ -32,6 +32,18 @@
...
+     * The experimental feature: QPG.
+     *
+     * @param globalState
+     *            the state created and is valid for this method call.
+     *
+     * @throws Exception
+     *             if testing fails.
+     *
+     */
+    void qpg(G globalState) throws Exception;
```
mrigger, line 45: I think `qpg` is quite cryptic for people not already familiar with the technique. Could we use another method name, like `generateAndTestDatabaseWithQueryGuidance`?

src/sqlancer/ProviderAdapter.java
```diff
@@ -67,7 +91,47 @@ public Reproducer<G> generateAndTestDatabase(G globalState) throws Exception {
         return null;
     }
 
-    protected abstract void checkViewsAreValid(G globalState);
+    @Override
+    public void qpg(G globalState) throws Exception {
+        try {
+            generateDatabase(globalState);
+            checkViewsAreValid(globalState);
+            globalState.getManager().incrementCreateDatabase();
```
mrigger, line 99: The logic here seems to be the same as for `generateAndTestDatabase`. Perhaps we can have one common method `generateDatabase` and then the two methods, for testing with QPG and one with randomly generating tests?

src/sqlancer/SQLProviderAdapter.java
```diff
@@ -20,25 +21,22 @@ public LoggableFactory getLoggableFactory() {
...
     @Override
-    protected void checkViewsAreValid(G globalState) {
+    protected void checkViewsAreValid(G globalState) throws SQLException {
         List<? extends AbstractTable<?, ?, ?>> views = globalState.getSchema().getViews();
         for (AbstractTable<?, ?, ?> view : views) {
             SQLQueryAdapter q = new SQLQueryAdapter("SELECT 1 FROM " + view.getName() + " LIMIT 1");
             try {
                 if (!q.execute(globalState)) {
-                    dropView(globalState, view.getName());
+                    throw new AssertionError();
```
mrigger, line 30: Why do we throw an `AssertionError` here rather than trying to drop the view? The `SELECT` might fail when the view is invalid.

src/sqlancer/MainOptions.java
```diff
@@ -50,6 +50,27 @@ public class MainOptions {
     @Parameter(names = "--print-failed", description = "Logs failed insert, create and other statements without results", arity = 1)
     private boolean loggerPrintFailed = true; // NOPMD
 
+    @Parameter(names = "--log-execution-result", description = "Logs the execution result of each statement (requires --log-each-select to be enabled)", arity = 1)
+    private boolean logExecutionResult = true; // NOPMD
+
+    @Parameter(names = "--log-query-plan", description = "Logs the query plans of each statemen", arity = 1)
```
mrigger, line 56: Of each statement or query?

src/sqlancer/MainOptions.java
```diff
@@ -50,6 +50,27 @@ public class MainOptions {
...
+    @Parameter(names = "--enable-qpg", description = "Enable the experimental feature Query Plan Guidance (QPS)", arity = 1)
+    private boolean enableQPG;
+
+    @Parameter(names = "--enable-random-qpg", description = "Enable the random mutation of Query Plan Guidance (QPS) (requires --enable-qpg to be enabled)", arity = 1)
+    private boolean enableRandomQPG;
+
+    @Parameter(names = "--qpg-min-interval", description = "Specifies the minimal number of iterations between table mutations")
+    private static int qpgMinInterval = 1000;
+
+    @Parameter(names = "--qpg-k", description = "Specifies the k parameter of the QPG (The value should be 1/k)")
```
mrigger, line 68: What does `k` describe?

src/sqlancer/MainOptions.java
```diff
@@ -50,6 +50,27 @@ public class MainOptions {
     @Parameter(names = "--print-failed", description = "Logs failed insert, create and other statements without results", arity = 1)
     private boolean loggerPrintFailed = true; // NOPMD
 
+    @Parameter(names = "--log-execution-result", description = "Logs the execution result of each statement (requires --log-each-select to be enabled)", arity = 1)
+    private boolean logExecutionResult = true; // NOPMD
+
+    @Parameter(names = "--log-query-plan", description = "Logs the query plans of each statemen", arity = 1)
+    private boolean logQueryPlan; // NOPMD
+
+    @Parameter(names = "--enable-qpg", description = "Enable the experimental feature Query Plan Guidance (QPS)", arity = 1)
```
mrigger, line 59: Perhaps we can use a common prefix for all QPG options? Here, we could just use `--qpg` as an option (or `--qpg-enable`).

src/sqlancer/sqlite3/SQLite3Provider.java
```diff
@@ -49,56 +55,85 @@ public class SQLite3Provider extends SQLProviderAdapter<SQLite3GlobalState, SQLi
...
+    private static double[] weightedAverageReward = new double[Action.values().length];
+    private int[] cumulativeMutationTimes;
+
+    // For post reward calculation
+    private int currentSelectRewards;
+    private int currentSelectCounts;
+    private int currentMutationOperator;
+
     // PRAGMAS to achieve good performance
     private static final List<String> DEFAULT_PRAGMAS = Arrays.asList("PRAGMA cache_size = 50000;",
```
bajinsheng, line 70: reduce this size aiming to avoid OOM.

PR #635, opened 2022-12-11
fix ci when to install citus

hawkingrei:
we can read some error message when to install citus.

```
The repository is set up! You can now install packages.
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package postgresql-13-citus-10.1
E: Couldn't find any package by glob 'postgresql-13-citus-10.1'
E: Couldn't find any package by regex 'postgresql-13-citus-10.1'
Error: Process completed with exit code 100.
``` 
Signed-off-by: Weizhen Wang <wangweizhen@pingcap.com>


mrigger (approved):
Great, thanks for updating!

@halilozanakgul there are a couple of new Citus issues that would need to be addressed (or indicate bugs in Citus?). Not sure if you or anyone else has time to look at them?

PR #638, opened 2022-12-14
tidb: turn off bug switch

hawkingrei:
https://github.com/pingcap/tidb/issues/38295 is resolved.

Signed-off-by: Weizhen Wang <wangweizhen@pingcap.com>

### DIFF
--- a/src/sqlancer/tidb/TiDBBugs.java
+++ b/src/sqlancer/tidb/TiDBBugs.java
@@ -2,21 +2,8 @@ package sqlancer.tidb;
 
 // do not make the fields final to avoid warnings
 public final class TiDBBugs {
-
-    // https://github.com/pingcap/tidb/issues/15987
-    public static boolean bug15987;
-
-    // // https://github.com/pingcap/tidb/issues/15988
-    public static boolean bug15988;
-
     // https://github.com/pingcap/tidb/issues/16028
     public static boolean bug16028;
-
-    // https://github.com/pingcap/tidb/issues/16020
-    public static boolean bug16020;
-
-    // https://github.com/pingcap/tidb/issues/15990
-    public static boolean bug15990;
 
     // https://github.com/pingcap/tidb/issues/15844
     public static boolean bug15844;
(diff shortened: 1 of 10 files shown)


src/sqlancer/tidb/TiDBBugs.java
```diff
@@ -55,7 +52,7 @@ public final class TiDBBugs {
     public static boolean bug35652 = true;
 
     // https://github.com/pingcap/tidb/issues/38295
-    public static boolean bug38295 = true;
+    public static boolean bug38295;
```
mrigger, line 55: Do you think we should remove the code that is executed when this bug is set to `true` as well, or should we keep it in to serve as documentation?

hawkingrei: In fact, I find the bug switch set as false if the bug is fixed. I think it may be a coding habit of sqlancer. 

https://github.com/sqlancer/sqlancer/blob/102ec94c87f281b66a1ae34d0c75a3b98f5ebe03/src/sqlancer/tidb/TiDBBugs.java#L6-L13

hawkingrei: https://github.com/pingcap/tidb/issues/38295 has been fixed. I think we can remove the code and limit. Although we can find the ``` assertion failed```, it is not the same problem.

hawkingrei: Can I remove all false bug switches if we think it is fixed?

hawkingrei: I find the problem in the CI. Let me study in depth. >。<

mrigger: > Can I remove all false bug switches if we think it is fixed?

Yes, I think we could remove them, which would reduce also maintenance costs.

mrigger: I figured that some implementations keep the flags and link to the relevant bug reports in the code (for documentation purposes?) as you mentioned - for me, either way is fine.

mrigger:
Are the errors that cause the TiDB tests to fail due to this PR or unrelated? If they are unrelated, we could go ahead and merge the PR.

hawkingrei:
> Are the errors that cause the TiDB tests to fail due to this PR or unrelated? If they are unrelated, we could go ahead and merge the PR.

it has some problem in the tidb with unistore. I remove some useless bug switch and the other is still useful.

mrigger (approved):
LGTM!

PR #640, opened 2022-12-16
Escaping newlines in logs

bajinsheng:
For some databases, especially SQLite, it is required to escape the newlines in the log for a reproducible log.


mrigger (approved):
LGTM, thanks!

PR #642, opened 2022-12-16
fix bug in CREATE TABLE

bajinsheng:
The assertion error should not happen when executing `EXPLAIN CREATE TABLE` statement.


mrigger (approved):
Interesting, I didn't know that it was possible to execute a statement like `EXPLAIN CREATE TABLE t0(c0 INT);`. Does only SQLite support this from the database systems that you tested?

bajinsheng:
SQLite supports this feature: https://github.com/sqlancer/sqlancer/blob/master/src/sqlancer/sqlite3/gen/SQLite3ExplainGenerator.java#L22

After `EXPLAIN`, any statement is possible

PR #643, opened 2022-12-16
Fix: no table selected

bajinsheng:
When no table is selected, the execution should be aborted to avoid execution errors in QPG.

### DIFF
--- a/src/sqlancer/common/schema/AbstractSchema.java
+++ b/src/sqlancer/common/schema/AbstractSchema.java
@@ -41,7 +41,12 @@ public class AbstractSchema<G extends GlobalState<?, ?, ?>, A extends AbstractTa
     }
 
     public A getRandomTable(Predicate<A> predicate) {
-        return Randomly.fromList(getDatabaseTables().stream().filter(predicate).collect(Collectors.toList()));
+        List<A> filtered = getDatabaseTables().stream().filter(predicate).collect(Collectors.toList());
+        if (filtered.isEmpty()) {
+            throw new IgnoreMeException();
+        } else {
+            return Randomly.fromList(filtered);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


mrigger (commented):
Rather than modifying the existing method, perhaps we could use a `getRandomTableOrBailout` method such as the one below that throws an exception when no table exists? In many contexts, it might be clear that a table exists, and not throwing an exception in such cases might help detect otherwise overlooked bugs (in SQLancer).

bajinsheng:
I think it should be a bug as all dbs assume the returned list has at least one table and directly operate the object.
https://github.com/sqlancer/sqlancer/search?q=getRandomTable

bajinsheng:
I cannot remember it exactly, but it should be some corner cases in which no table is returned, but there is no bug.

bajinsheng:
Like this, it gets a random table and assumes the returned list is not empty: https://github.com/sqlancer/sqlancer/blob/695dda5741d5c9587535f4fe8d469bf4acd33ff3/src/sqlancer/cockroachdb/gen/CockroachDBShowGenerator.java#L25

But its assumption is not always true. It is a potential bug that was not exposed before.

bajinsheng:
We can close it as it is, and apply this fix when observing this issue again.

mrigger:
> I think it should be a bug as all dbs assume the returned list has at least one table and directly operate the object.

Not sure if I would call this a bug - all implementations relying on this method indeed assume that at least one table exists, and I can't think of a circumstance where this doesn't hold. Did QPG result in a corner case where this assumption no longer holds? If so, perhaps we can rename `getRandomTable` to `getRandomTableOrBailout` to follow the naming convention in this class?

mrigger:
> We can close it as it is, and apply this fix when observing this issue again.

Okay, sounds good.

PR #644, opened 2022-12-16
add interfaces to get random virtual/rtree table names

bajinsheng:
This is a lack of random name generation for virtual and rtree tables.


mrigger (commented):
Virtual tables and rtree are specific to SQLite. Perhaps putting these in the `SQLite3Schema` or so would be a better fit?

bajinsheng:
`databaseTables` is private and cannot be accessed in `SQLite3Schema`

mrigger:
We can use the getter `getDatabaseTables()`, I think?

PR #646, opened 2022-12-16
add databend bugs

hanyisong:


src/sqlancer/databend/DatabendErrors.java
```diff
@@ -30,7 +30,7 @@ public static void addExpressionErrors(ExpectedErrors errors) {
         if (DatabendBugs.bug9264) {
             errors.add("assertion failed: offset + length <= self.length");
         }
-        errors.add("validity's length must be equal"); // bug
+        // errors.add("validity's length must be equal"); // bug
```
mrigger, line 33: Perhaps we can remove this comment instead?

hanyisong: This error was triggered with a small probability before, I forgot to leave a log.

mrigger: Okay, I see. Then let's leave the comment in for now.

mrigger:
By the way, before reporting bugs, it is useful to reduce the bug-inducing test case to a minimal version that still triggers the bug (e.g., by repeatedly removing statements as long as the error/incorrect result is triggered). This also helps in identifying what potential features trigger the bug, whose generation can be disabled until the bug is fixed.

mrigger:
@kshehata recently added experimental support for test case reduction (see https://github.com/sqlancer/sqlancer/pull/624). Perhaps you could check whether this is useful for your efforts in testing Databend?

hanyisong:
Thanks for the advice, I was too anxious.

PR #648, opened 2022-12-16
Add interface to return queries by oracles

bajinsheng:
For QPG, we need to obtain the queries that are returned from oracles to measure query plans.

### DIFF
--- a/src/sqlancer/sqlite3/oracle/SQLite3NoRECOracle.java
+++ b/src/sqlancer/sqlite3/oracle/SQLite3NoRECOracle.java
@@ -102,6 +102,11 @@ public class SQLite3NoRECOracle extends NoRECBase<SQLite3GlobalState> implements
         return reproducer;
     }
 
+    @Override
+    public String getLastQueryString() {
+        return optimizedQueryString;
+    }
+
     private Function<SQLite3GlobalState, Integer> getUnoptimizedQuery(SQLite3Select select,
             SQLite3Expression randomWhereCondition) throws SQLException {
         SQLite3PostfixUnaryOperation isTrue = new SQLite3PostfixUnaryOperation(PostfixUnaryOperator.IS_TRUE,
--- a/src/sqlancer/cockroachdb/oracle/CockroachDBNoRECOracle.java
+++ b/src/sqlancer/cockroachdb/oracle/CockroachDBNoRECOracle.java
@@ -150,4 +150,9 @@ public class CockroachDBNoRECOracle extends NoRECBase<CockroachDBGlobalState>
         return count;
     }
 
+    @Override
+    public String getLastQueryString() {
+        return optimizedQueryString;
+    }
+
 }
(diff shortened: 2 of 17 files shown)


bajinsheng:
If you like this way, I require more time to do it.

1. The Reproducer only supports NoREC oracle of SQLite. It requires much more time to implement more oracles and databases like this: 
2. The `Reproducer` needs to be redesigned.
3. I am not familiar with what the reproducer does, so there are more potential conflicts and issues.
4. `Reproducer` name is not suitable.
5. Do You want to return the `Reproducer` or continue the current way that obtains it by a function?

src/sqlancer/ProviderAdapter.java
```diff
@@ -52,7 +52,7 @@ public Reproducer<G> generateAndTestDatabase(G globalState) throws Exception {
 
                     } catch (AssertionError e) {
                         Reproducer<G> reproducer = oracle.getLastReproducer();
-                        if (reproducer != null) {
+                        if (reproducer.getReproducerState()) {
```
mrigger, line 55: Is there any reason why need this method? I think it would be better to check for `null`: either we have the reproducer, or it is `null`. Having a reproducer object that indicates that we don't have the actual state is a bit confusing, I think.

src/sqlancer/Reproducer.java
```diff
@@ -1,5 +1,15 @@
...
 public interface Reproducer<G extends GlobalState<?, ?, ?>> {
     boolean bugStillTriggers(G globalState);
+
+    boolean getReproducerState();
+
+    void setReproducerState(boolean ifReproducerEnabled);
+
+    List<String> getQueriesAsString();
+
+    void addQuery(String query);
```
mrigger, line 14: Perhaps `addQueryString` to match the above method?

src/sqlancer/Reproducer.java
```diff
@@ -1,5 +1,15 @@
 package sqlancer;
 
+import java.util.List;
+
 public interface Reproducer<G extends GlobalState<?, ?, ?>> {
     boolean bugStillTriggers(G globalState);
```
mrigger, line 6: Perhaps now, we can rename this method to `bugTriggers`, given that we create the reproducer also if no bug was triggered?

src/sqlancer/sqlite3/oracle/SQLite3NoRECOracle.java
```diff
@@ -90,11 +116,12 @@ public void check() throws SQLException {
             throw new IgnoreMeException();
         }
         if (optimizedCount != unoptimizedCount) {
```
mrigger, line 118: I think we could simplify the code by adding `reproducer = new SQLite3NoRECReproducer(optimizedQuery, unoptimizedQuery, Arrayas.asList(optimizedQueryString, unoptimizedQueryString));` above the if check and removing the other parts.

src/sqlancer/ProviderAdapter.java
bajinsheng: `reproducer` is returned by each call of `check`, instead of bug only, so we need to distinguish the returned `reproduer` denotes a bug or not.

mrigger (approved):
LGMT, thanks!

PR #650, opened 2022-12-17
fix: broken grammar of LIKELIHOOD

bajinsheng:
To avoid the grammar error: `The second argument of LIKELIHOOD must be a float number within 0.0 -1.0`


bajinsheng:
This fix works for SQLite >=3.6.0

mrigger (commented):
I assume with this change, we can also remove the corresponding expected error below? https://github.com/sqlancer/sqlancer/blob/b5eb2153dfd06bf50aca0f98eb32383b389a933f/src/sqlancer/sqlite3/SQLite3Errors.java#L24

bajinsheng:
Oh yes. Didn't notice that.

mrigger (approved):
LGTM!

PR #655, opened 2022-12-17
Add support for FLOAT for PostgreSQL Binary Arithmetic Operators

albertZhangTJ:


src/sqlancer/postgres/ast/PostgresBinaryArithmeticOperation.java
```diff
@@ -58,15 +58,21 @@ public PostgresConstant apply(PostgresConstant left, PostgresConstant right) {
         private String textRepresentation;
 
         private static PostgresConstant applyBitOperation(PostgresConstant left, PostgresConstant right,
```
albertZhangTJ, line 60: This function name seems a bit confusing. Should it be `applyBinaryOperation`?

mrigger: You are right, or perhaps `applyBinaryArithmeticOperation` - either would be better.

src/sqlancer/postgres/ast/PostgresBinaryArithmeticOperation.java
```diff
@@ -58,15 +58,21 @@ public PostgresConstant apply(PostgresConstant left, PostgresConstant right) {
...
-                BinaryOperator<Long> op) {
+                BinaryOperator<Float> op) {
```
mrigger, line 61: Using `BinaryOperator<Float>` in both an integer and floating-point context seems error-prone. Not sure if we should add another class instead?

mrigger (commented):
Thanks for the PR! I'm wondering if we really need PQS support for floats for Postgres, as it might be difficult to get correct due to potential minor rounding differences. If we want to add support, we should make sure that it doesn't cause any false alarms (e.g., by testing corner cases).

src/sqlancer/postgres/ast/PostgresBinaryArithmeticOperation.java
albertZhangTJ: Sure. I will handle this by adding common classes for these data types in a separate PR.

albertZhangTJ:
My apologies for my carelessness. I will update to try to handle the rounding error.

I will also add some test cases to validate the code.

PR #659, opened 2022-12-18
fix: switch off vectorize cockroachdb

bajinsheng:
The vectorize option is `on` by default: https://www.cockroachlabs.com/docs/v22.2/vectorized-execution#configure-vectorized-execution so we are required to randomly switch it off here.


mrigger (commented):
Since vectorization is on by default, perhaps we can just remove the option to make vectorization more likely and the code here?

mrigger:
https://github.com/sqlancer/sqlancer/blob/master/src/sqlancer/cockroachdb/gen/CockroachDBSetSessionGenerator.java also randomly turns vectorization off.

bajinsheng:
> https://github.com/sqlancer/sqlancer/blob/master/src/sqlancer/cockroachdb/gen/CockroachDBSetSessionGenerator.java also randomly turns vectorization off.

So we fully remove this switch here?

mrigger (approved):

PR #662, opened 2022-12-18
Refactor UpdateGenerator classes using AbstractUpdateGenerator class

eltfshr:
The following classes have been refactored to extend AbstractUpdateGenerator class:
- CockroachDBUpdateGenerator
- DuckDBUpdateGenerator
- H2UpdateGenerator
- HSQLDBUpdateGenerator
- OceanBaseUpdateGenerator
- PostgresUpdateGenerator
- SQLite3UpdateGenerator
- TiDBUpdateGenerator
- YCQLUpdateGenerator
- YSQLUpdateGenerator

\* MariaDBUpdateGenerator is not refactored because it has a different way of getting random columns


src/sqlancer/hsqldb/gen/HSQLDBUpdateGenerator.java
```diff
@@ -4,54 +4,53 @@
...
+        updateColumns(columns);
+        return new SQLQueryAdapter(sb.toString(), errors);
+    }
 
+    @Override
+    protected void updateValue(HSQLDBColumn column) {
+        Node<HSQLDBExpression> expr;
+        expr = gen.generateConstant(column.getType());
+        sb.append(HSQLDBToStringVisitor.asString(expr));
+        if (Randomly.getBooleanWithSmallProbability()) {
```
mrigger, line 47: I think this actually goes after the `updateColumn` call (as an `UPDATE` statement has only a single `WHERE` clause at the end of the statement). This seems to be a bug in the code before your refactoring (cc @nish-d). Perhaps you can fix this as part of this refactoring?

mrigger (commented):
Thanks a lot for the PR. It looks good to me overall. Could you please have a look at the comments to check whether you can address them?

mrigger (approved):
LGTM, thanks!

PR #672, opened 2022-12-22
Add regex support for ExpectedError

albertZhangTJ:


src/sqlancer/common/query/ExpectedErrors.java
```diff
@@ -12,6 +15,10 @@
 public class ExpectedErrors {
 
     private final Set<String> errors = new HashSet<>();
+    private final List<Pattern> regexes = new ArrayList<>();
+
+    private boolean usingString = true;
```
mrigger, line 20: I'd suggest removing the `usingString` and `usingRegex` fields - when a string or regex is added, they should be used. This also simplifies the implementation.

albertZhangTJ: Sure! Already removed in the following commit.

mrigger (approved):
LGTM, thanks!

PR #680, opened 2023-01-03
Fix issue with unwanted Error when --log-each-select is set to false

### LINKED ISSUE #679: Program will always throw AssertionError if --log-each-select is set to false
To reproduce, run
`java -jar sqlancer-*.jar --num-threads 4 --log-each-select false mysql --oracle TLP_WHERE`

albertZhangTJ:
Attempt to fix #679 
Manually tested on MySQL

### DIFF
--- a/src/sqlancer/SQLGlobalState.java
+++ b/src/sqlancer/SQLGlobalState.java
@@ -20,8 +20,10 @@ public abstract class SQLGlobalState<O extends DBMSSpecificOptions<?>, S extends
         if (success && getOptions().printSucceedingStatements()) {
             System.out.println(q.getQueryString());
         }
-        if (logExecutionTime) {
-            getLogger().writeCurrent(" -- " + timer.end().asString());
+        if (getOptions().logEachSelect()){
+            if (logExecutionTime) {
+                getLogger().writeCurrent(" -- " + timer.end().asString());
+            }
         }
         if (q.couldAffectSchema()) {
             updateSchema();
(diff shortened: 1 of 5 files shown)


mrigger:
@bajinsheng could you please review this PR, since you have recently worked on the logic for writing out query plans?

src/sqlancer/mongodb/MongoDBProvider.java
```diff
@@ -72,8 +72,10 @@ protected void executeEpilogue(Query<?> q, boolean success, ExecutionTimer timer
             if (success && getOptions().printSucceedingStatements()) {
                 System.out.println(q.getLogString());
             }
-            if (logExecutionTime) {
-                getLogger().writeCurrent("// " + timer.end().asString());
+            if (getOptions().logEachSelect()) {
```
mrigger, line 75: I think the logic would be slightly easier to understand if you have the logic in one if statement (here and above): `if (getOptions().logEachSelect() && logExecutionTime)`

albertZhangTJ: Sure! Fixed in a newer commit.

src/sqlancer/SQLGlobalState.java
```diff
@@ -20,8 +20,10 @@ protected void executeEpilogue(Query<?> q, boolean success, ExecutionTimer timer
         if (success && getOptions().printSucceedingStatements()) {
             System.out.println(q.getQueryString());
         }
-        if (logExecutionTime) {
```
SIPRAA09, line 23: if (getOptions().logEachSelect() && logExecutionTime) {
            getLogger().writeCurrent(" -- " + timer.end().asString());
        }
        if (q.couldAffectSchema()) {
            updateSchema();
        }

PR #681, opened 2023-01-03
Rewrite ClickHouse query generation

qoega:


qoega:
At least locally I run 5 minutes per test and it worked.
Screenshot 2023-01-03 at 21 32 52

qoega:
Muted this issue for now https://github.com/ClickHouse/ClickHouse/issues/44866
We already have a fix in new query analyzer and it will be enabled soon by default. In old code we probably will not fix it.

src/sqlancer/clickhouse/ClickHouseErrors.java
```diff
@@ -62,7 +62,17 @@ public static void addExpectedExpressionErrors(ExpectedErrors errors) {
         errors.add("ILLEGAL_DIVISION");
         errors.add("DECIMAL_OVERFLOW");
         errors.add("Cannot convert out of range floating point value to integer type");
+        errors.add("Unexpected inf or nan to integer conversion");
+        errors.add("No such name in Block::erase"); // https://github.com/ClickHouse/ClickHouse/issues/42769
```
mrigger, line 66: For such unfixed issues, would you like to adopt the pattern described in https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#unfixed-bugs? I think this would make the code more maintainable (let me know if you don't want to).

src/sqlancer/clickhouse/gen/ClickHouseExpressionGenerator.java
```diff
@@ -53,48 +53,128 @@ private enum Expression {
         UNARY_POSTFIX
     }
 
-    public ClickHouseExpression generateExpressionWithColumns(List<ClickHouseColumnReference> columns, int depth) {
-        if (columns.isEmpty() || depth > 0 && Randomly.getBoolean()) {
+    public ClickHouseExpression generateExpressionWithColumns(List<ClickHouseColumnReference> columns, int complexity) {
```
mrigger, line 56: I think `complexity` is quite ambiguous. Perhaps `remainingDepth` would be a more descriptive term?

src/sqlancer/clickhouse/gen/ClickHouseExpressionGenerator.java
```diff
@@ -53,48 +53,128 @@ private enum Expression {
...
         }
     }
 
+    public ClickHouseExpression generateAggregateExpressionWithColumns(List<ClickHouseColumnReference> columns,
+            int complexity) {
+        if (Randomly.getBooleanWithRatherLowProbability()) {
+            return new ClickHouseAggregate(generateExpressionWithColumns(columns, complexity - 1),
+                    ClickHouseAggregate.ClickHouseAggregateFunction.getRandom());
+        }
+        if (columns.isEmpty() || complexity <= 2 && Randomly.getBooleanWithRatherLowProbability()) {
```
mrigger, line 92: Instead of the raw integers `2`, perhaps it might improve maintainability to introduce a constant?

mrigger (commented):
Thanks! Let me know which (if any) of the suggestions you want to incorporate.

src/sqlancer/clickhouse/ClickHouseErrors.java
qoega: I remember you showed this one already. I usually do not know what should be disabled to disable behavior with errors. But I know that that issue will soon be fixed and I will remove this row

src/sqlancer/clickhouse/gen/ClickHouseExpressionGenerator.java
qoega: Ok

src/sqlancer/clickhouse/gen/ClickHouseExpressionGenerator.java
qoega: I really don't have any clue which numbers should I use there.
There is this minimum depth where I may generate some trivial leaves with constants or columns. There is depth itself and I saw strange things that will smaller depth expressions were incorrect with higher probability. Maybe I should treat other places where I put constants where I generate different types of expressions(where/order by/join on). Actually I did something like that while testing the result.

qoega (commented):
I will rename `complexity`.
What shall I do with constants used for expressions generations I will think asynchronously. At least I want to see how it behaves now in CI - if I need to tune them or it just works not.

Those changes were uncommited for months I guess and it will fix failures that I see in our CI. Because of them I have those checks in always green mode and it takes too much effort to filter problems with generator itself vs actual issues. So my plan is to see what will happen with checks after this changes 

[Link to checks failures history stats](https://play.clickhouse.com/play?user=play#U0VMRUNUIGNoZWNrX25hbWUsIHRlc3RfbmFtZSBBUyBuYW1lLCB0b1N0YXJ0T2ZXZWVrKGNoZWNrX3N0YXJ0X3RpbWUpIGFzIHdlZWssIGNvdW50KCkgQVMgYywgYXJnTWF4KHJlcG9ydF91cmwsIGNoZWNrX3N0YXJ0X3RpbWUpCkZST00gY2hlY2tzCldIRVJFIAogICBjaGVja19uYW1lIExJS0UgJ1NRTGFuY2VyJScKICAgQU5EICh0ZXN0X3N0YXR1cyBMSUtFICdGJScgT1IgdGVzdF9zdGF0dXMgTElLRSAnRSUnKQogICBBTkQgcHVsbF9yZXF1ZXN0X251bWJlciA9IDAKR1JPVVAgQlkgd2VlaywgY2hlY2tfbmFtZSwgbmFtZQpPUkRFUiBCWSBjIERFU0MK)

mrigger:
Sounds good! I will merge after fixing the CI formatting error (which should be fixable by executing `mvn formatter:format`.

qoega:
> fixing the CI formatting error (which should be fixable by executing mvn formatter:format.

I usually break it even running all this many times. Recently I learned those but still break style check 
```
mvn -B verify -DskipTests=true
```

PR #687, opened 2023-01-11
Update ClickHouse expected errors

qoega:
Added couple new.
Removed fixed issues.


qoega:
I will check why `Different order of columns in UNION subquery` is still triggered separately

qoega:
You can merge this one too.

mrigger:
LGTM! Sorry, I was on the road and didn't see this one.

qoega:
No problem. Just wrote in case you thought I want to check errors that I faced in first version.

By the way after improvements more interesting cases started to appear like error in JIT or by some reason no result after adding HAVING expression which is always true. I'm collecting interesting cases

PR #691, opened 2023-01-12
Fix comparator helper for GroupBy Oracle in ClickHouse

qoega:


qoega:
I think TiDB and CockroachDB "Required" flags may be removed for now. They fail a lot.

mrigger:
Yes, good point, we'll set them to be optional.

PR #697, opened 2023-01-18
Support CnosDB

ZuoTiJia:
Hello, I am from the CnosDB development team. We implemented Sqlancer of CnosDB referring to the code of Postgres in Sqlancer. This tool helped us find some bugs.


mrigger:
Awesome, thanks for the PR! Starting to review the PR asap. In the meanwhile, could you please execute `mvn -B verify -DskipTests=true` and fix the warnings?

ZuoTiJia:
> Awesome, thanks for the PR! Starting to review the PR asap. In the meanwhile, could you please execute `mvn -B verify -DskipTests=true` and fix the warnings?

Thank you for reminding

src/sqlancer/cnosdb/CnosDBComparatorHelper.java
```diff
@@ -0,0 +1,163 @@
...
+import sqlancer.cnosdb.client.CnosDBResultSet;
+import sqlancer.cnosdb.query.CnosDBSelectQuery;
+import sqlancer.common.query.ExpectedErrors;
+
+public final class CnosDBComparatorHelper {
+
+    private CnosDBComparatorHelper() {
+    }
+
+    public static boolean isEqualDouble(String first, String second) {
```
mrigger, line 21: For this and other methods, is there actually a need to re-implement them? It seems that, at least for this method and the one below, the code is the same as for https://github.com/sqlancer/sqlancer/blob/67751dc3d83b880cd8c7ed53416b8b157a2ba2ea/src/sqlancer/ComparatorHelper.java#L21-L37

src/sqlancer/cnosdb/CnosDBGlobalState.java
```diff
@@ -0,0 +1,53 @@
...
+    public static final char IMMUTABLE = 'i';
+    public static final char STABLE = 's';
+    public static final char VOLATILE = 'v';
+
+    private final Map<String, Character> functionsAndTypes = new HashMap<>();
+    private final List<Character> allowedFunctionTypes = Arrays.asList(IMMUTABLE, STABLE, VOLATILE);
+
+    @Override
+    protected void executeEpilogue(Query<?> q, boolean success, ExecutionTimer timer) throws Exception {
+        boolean logExecutionTime = getOptions().logExecutionTime();
```
mrigger, line 24: This seems to be a copy of https://github.com/sqlancer/sqlancer/blob/67751dc3d83b880cd8c7ed53416b8b157a2ba2ea/src/sqlancer/SQLGlobalState.java#L19-L27

I think it should be possible to inherit from `SQLGlobalState` instead?

src/sqlancer/cnosdb/CnosDBLoggableFactory.java
```diff
@@ -0,0 +1,59 @@
...
+        errors.addAll(CnosDBExpectedError.expectedErrors());
+        return new CnosDBOtherQuery(queryString, errors);
+    }
+
+    @Override
+    public CnosDBQueryAdapter commentOutQuery(Query<?> query) {
+        String queryString = query.getLogString();
+        String newQueryString = "-- " + queryString;
+        ExpectedErrors errors = new ExpectedErrors();
+        errors.addAll(CnosDBExpectedError.expectedErrors());
```
mrigger, line 40: I believe that commenting out a SQL statement should not result in any errors, so the expected errors could be empty here?

src/sqlancer/cnosdb/CnosDBLoggableFactory.java
```diff
@@ -0,0 +1,59 @@
...
+
+import sqlancer.cnosdb.query.CnosDBOtherQuery;
+import sqlancer.cnosdb.query.CnosDBQueryAdapter;
+import sqlancer.common.log.Loggable;
+import sqlancer.common.log.LoggableFactory;
+import sqlancer.common.log.LoggedString;
+import sqlancer.common.query.ExpectedErrors;
+import sqlancer.common.query.Query;
+
+public class CnosDBLoggableFactory extends LoggableFactory {
```
mrigger, line 14: I think inheriting from https://github.com/sqlancer/sqlancer/blob/67751dc3d83b880cd8c7ed53416b8b157a2ba2ea/src/sqlancer/common/log/SQLLoggableFactory.java#L9 and overriding methods whose implementation differs would prevent some code duplication here.

src/sqlancer/cnosdb/ast/CnosDBAggregate.java
```diff
@@ -0,0 +1,112 @@
...
+                CnosDBDataType.UINT),
+        COUNT(CnosDBDataType.INT) {
+            @Override
+            public CnosDBDataType[] getInputTypes(CnosDBDataType returnType) {
+                return new CnosDBDataType[] { CnosDBDataType.getRandomType() };
+            }
+        },
+        SUM(CnosDBDataType.INT, CnosDBDataType.DOUBLE, CnosDBDataType.UINT), APPROX_MEDIAN(CnosDBDataType.DOUBLE);
+
+        // Currently these aggregate functions have bugs https://github.com/cnosdb/cnosdb/issues/786
```
mrigger, line 33: As an optional suggestion, I think it would be more elegant to include the currently commented-out code. If, for example, you require a random aggregate function, you could still remove the buggy aggregate functions from the selection. The general pattern is described at https://github.com/sqlancer/sqlancer/blob/67751dc3d83b880cd8c7ed53416b8b157a2ba2ea/CONTRIBUTING.md#implementing-support-for-a-new-dbms, and I see you added a `CnosDBBugs` class already.

src/sqlancer/cnosdb/ast/CnosDBBinaryArithmeticOperation.java
```diff
@@ -0,0 +1,123 @@
...
+
+public class CnosDBBinaryArithmeticOperation extends BinaryOperatorNode<CnosDBExpression, CnosDBBinaryOperator>
+        implements CnosDBExpression {
+
+    public CnosDBBinaryArithmeticOperation(CnosDBExpression left, CnosDBExpression right, CnosDBBinaryOperator op) {
+        super(left, right, op);
+    }
+
+    @Override
+    public CnosDBConstant getExpectedValue() {
```
mrigger, line 21: Currently, the PR includes many `getExpectedValue` implementations, but does not support PQS, for which they are needed. Perhaps we can remove these implementations?

src/sqlancer/cnosdb/ast/CnosDBBinaryArithmeticOperation.java
```diff
@@ -0,0 +1,123 @@
...
+
+        public static CnosDBBinaryOperator getRandom(CnosDBDataType dataType) {
+            List<CnosDBBinaryOperator> ops = new ArrayList<>(Arrays.asList(values()));
+            switch (dataType) {
+            case DOUBLE:
+            case UINT:
+            case STRING:
+                ops.remove(MODULO);
+                ops.remove(EXPONENTIATION);
+                ops.remove(EXPONENTIATION);
```
mrigger, line 104: Why are these removed twice? A comment to explain this would also be helpful.

src/sqlancer/cnosdb/ast/CnosDBFunction.java
```diff
@@ -0,0 +1,276 @@
...
+                return type == CnosDBDataType.STRING;
+            }
+
+            @Override
+            public CnosDBDataType[] getInputTypesForReturnType(CnosDBDataType returnType, int nrArguments) {
+                return new CnosDBDataType[] { CnosDBDataType.STRING };
+            }
+
+        };
+        // NULL_IF(2, "nullif") {
```
mrigger, line 140: Perhaps we can remove this commented-out code?

src/sqlancer/cnosdb/ast/CnosDBFunctionWithUnknownResult.java
```diff
@@ -0,0 +1,110 @@
...
+    ACOS("acos", CnosDBDataType.DOUBLE, CnosDBDataType.DOUBLE),
+    ASIN("asin", CnosDBDataType.DOUBLE, CnosDBDataType.DOUBLE),
+    ATAN2("atan2", CnosDBDataType.DOUBLE, CnosDBDataType.DOUBLE, CnosDBDataType.DOUBLE),
+    COS("cos", CnosDBDataType.DOUBLE, CnosDBDataType.DOUBLE), SIN("sin", CnosDBDataType.DOUBLE, CnosDBDataType.DOUBLE),
+    SQRT("sqrt", CnosDBDataType.DOUBLE, CnosDBDataType.DOUBLE),
+    TAN("tan", CnosDBDataType.DOUBLE, CnosDBDataType.DOUBLE),
+    DATA_PART("date_part", CnosDBDataType.INT, CnosDBDataType.STRING, CnosDBDataType.TIMESTAMP);
+
+    // because bug of arrow-csv https://github.com/apache/arrow-rs/issues/3547
+    // NOW("now", CnosDBDataType.TIMESTAMP),
```
mrigger, line 66: As an optional suggestion, as in the other instance, we could apply the `CnosDBBugs` pattern here.

mrigger (commented):
Thanks a lot for the PR! Could you please take a look at my comments, which are minor issues and suggestions?

Currently, the implementation contains much effectively dead code, namely the `getExpectedValue` methods needed for Pivoted Query Synthesis (PQS), which is not supported as a test oracle. Would it be possible to remove these methods?

src/sqlancer/cnosdb/CnosDBGlobalState.java
ZuoTiJia: SQLGlobalState need SQLConnection. Currently our implementation does not use jdbc.

ZuoTiJia:
Thanks for your review, I removed some dead code.

src/sqlancer/cnosdb/CnosDBLoggableFactory.java
mrigger: I noticed that this is still unaddressed. Is it not possible to inherit because of some reason related to not using JDBC?

src/sqlancer/cnosdb/CnosDBGlobalState.java
mrigger: Okay, I see. It would be ideal to still re-use some code (e.g., by factoring out methods), but this could also be as part of a follow-up PR.

src/sqlancer/cnosdb/CnosDBExpectedError.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public final class CnosDBExpectedError {
+    private CnosDBExpectedError() {
+    }
+
+    public static List<String> expectedErrors() {
+        List<String> errors = new ArrayList<>();
```
mrigger, line 11: Here, it would be ideal to directly create an `ExpectedError` object rather than a list.

mrigger (commented):
Almost ready to merge! Could you please still have a look at the other comments?

src/sqlancer/cnosdb/CnosDBLoggableFactory.java
ZuoTiJia: At present, our implementation does not use JDBC. At the beginning of development, JDBC support was not perfect, so we used the http interface, and we will optimize it later.

src/sqlancer/cnosdb/ast/CnosDBAggregate.java
ZuoTiJia: Thanks. I removed the comment and applied that pattern.

src/sqlancer/cnosdb/ast/CnosDBBinaryArithmeticOperation.java
ZuoTiJia: Thanks, I remove those dead code.

src/sqlancer/cnosdb/ast/CnosDBFunction.java
ZuoTiJia: Thanks, I remove those comment.

src/sqlancer/cnosdb/ast/CnosDBFunctionWithUnknownResult.java
ZuoTiJia: Thanks, I remove those comment.

src/sqlancer/cnosdb/CnosDBComparatorHelper.java
```diff
@@ -0,0 +1,149 @@
...
+            String originalQueryString, List<String> combinedString, CnosDBGlobalState state) {
+        if (resultSet.size() != secondResultSet.size()) {
+            String queryFormatString = "-- %s;\n-- cardinality: %d";
+            String firstQueryString = String.format(queryFormatString, originalQueryString, resultSet.size());
+            String secondQueryString = String.format(queryFormatString, String.join(";", combinedString),
+                    secondResultSet.size());
+            state.getState().getLocalState().log(String.format("%s\n%s", firstQueryString, secondQueryString));
+            String assertionMessage = String.format("the size of the result sets mismatch (%d and %d)!\n%s\n%s",
+                    resultSet.size(), secondResultSet.size(), firstQueryString, secondQueryString);
+            Main.nrUnsuccessfulActions.addAndGet(1);
```
mrigger, line 75: The most recent commit added a couple of `Main.nrUnsuccessfulActions.addAndGet(1);` and similar calls like this. I believe we should remove them to not let the two comparator helper classes diverge until we fix the code duplication problem.

src/sqlancer/cnosdb/CnosDBExpectedError.java
```diff
@@ -0,0 +1,80 @@
...
+        ERRORS.add("The percentile argument for ApproxPercentileCont must be Float64, not Null.");
+        ERRORS.add("This feature is not implemented");
+        ERRORS.add("The function Avg does not support inputs of type Null.");
+        ERRORS.add("Coercion from [Utf8, Timestamp(Nanosecond, Some(\\\"+00:00\\\"))]");
+        ERRORS.add(
+                "Coercion from [Utf8, Float64, Utf8] to the signature OneOf([Exact([Utf8, Int64]), Exact([LargeUtf8, Int64]), Exact([Utf8, Int64, Utf8]), Exact([LargeUtf8, Int64, Utf8]), Exact([Utf8, Int64, LargeUtf8]), Exact([LargeUtf8, Int64, LargeUtf8])]) failed.");
+        ERRORS.add("Coercion from");
+    }
+
+    public static ExpectedErrors expectedErrors() {
```
mrigger, line 74: I think that this method should be called `getExpectedExpressionErrors` since all the errors seem specific to expressions? This is something that could be done in a future PR.

mrigger (approved):
LGTM, thanks a lot again for your contribution! It would be great to add CI tests (e.g., see https://github.com/sqlancer/sqlancer/blob/39c48cde9ff0ab29b950a8d3e35930b0949f36a0/.github/workflows/main.yml#L105) as one of the next steps, so we make sure that the implementation does not break in the future.

mrigger:
The branch is currently out-of-date. Could please merge latest the latest master? It seems I don't have permissions to do so.

ZuoTiJia:
> LGTM, thanks a lot again for your contribution! It would be great to add CI tests (e.g., see
> 
> https://github.com/sqlancer/sqlancer/blob/39c48cde9ff0ab29b950a8d3e35930b0949f36a0/.github/workflows/main.yml#L105
> 
> ) as one of the next steps, so we make sure that the implementation does not break in the future.

we'll add it soon

mrigger:
Thanks a lot!

Issue #707, opened 2023-01-29
Databend CI failing


mrigger:
The Databend CI tests currently fail, since SQLancer seems unable to connect to Databend:

``` 
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running sqlancer.dbms.TestDatabend
SQLancer failed creating a test database, indicating that SQLancer might have failed connecting to the DBMS. In order to change the username, password, host and port, you can use the --username, --password, --host and --port options.


java.sql.SQLException: Authenticate failed, user: "sqlancer", auth_plugin: "mysql_native_password"
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:129)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
	at com.mysql.cj.jdbc.ConnectionImpl.createNewIO(ConnectionImpl.java:828)
	at com.mysql.cj.jdbc.ConnectionImpl.<init>(ConnectionImpl.java:448)
	at com.mysql.cj.jdbc.ConnectionImpl.getInstance(ConnectionImpl.java:241)
	at com.mysql.cj.jdbc.NonRegisteringDriver.connect(NonRegisteringDriver.java:198)
	at java.sql/java.sql.DriverManager.getConnection(DriverManager.java:677)
	at java.sql/java.sql.DriverManager.getConnection(DriverManager.java:228)
	at sqlancer.databend.DatabendProvider.createDatabase(DatabendProvider.java:127)
	at sqlancer.databend.DatabendProvider.createDatabase(DatabendProvider.java:1)
	at sqlancer.Main$DBMSExecutor.testConnection(Main.java:332)
	at sqlancer.Main.executeMain(Main.java:509)
	at sqlancer.dbms.TestDatabend.testDatabendNoREC(TestDatabend.java:18)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
[...]

PR #713, opened 2023-01-31
fix: credential env for databend docker image

everpcpc:
for: #707

According to document on https://hub.docker.com/r/datafuselabs/databend


mrigger (approved):
Thanks! While connecting works again, it seems that the SQLancer implementation no longer works for the latest version. For example, the schema parsing fails due to errors such as the following (cc @hanyisong):

```
java.lang.AssertionError: BOOLEAN NULL
	at sqlancer.databend.DatabendSchema.getColumnType(DatabendSchema.java:285)
	at sqlancer.databend.DatabendSchema.getTableColumns(DatabendSchema.java:345)
```

sundy-li:
Yes, `DatabendSchema` display has changed in the latest version.

PR #714, opened 2023-01-31
Fix a Citus 'database is being accessed by other users' issue

mrigger:


mrigger:
See https://github.com/citusdata/citus/issues/6670.

PR #715, opened 2023-01-31
fix(databend):get data type error

hanyisong:


src/sqlancer/databend/DatabendSchema.java
```diff
@@ -339,6 +339,13 @@ private static List<DatabendColumn> getTableColumns(SQLConnection con, String ta
                     while (rs.next()) {
                         String columnName = rs.getString("column_name");
                         String dataType = rs.getString("data_type");
+                        if (dataType.contains("NULL")) {
+                            int idx = 0;
+                            while (idx < dataType.length() && dataType.charAt(idx) != ' ') {
```
mrigger, line 344: Perhaps `dataType.substring(0, dataType.indexOf(" "))` would be easier to understand?

mrigger (approved):
LGTM! Updating the branch before merging.

It seems that SQLancer found some (new?) logic bugs and an assertion failure in Databend according to the CI runs.

hanyisong:
> LGTM! Updating the branch before merging.
> 
> It seems that SQLancer found some (new?) logic bugs and an assertion failure in Databend according to the CI runs.

Yes, I have submitted these bugs

mrigger:
Would it be possible to ignore these bugs in https://github.com/sqlancer/sqlancer/blob/master/src/sqlancer/databend/DatabendBugs.java, so that the CI passes without errors until they are fixed?

PR #721, opened 2023-02-05
Fix QuestDB execution bug

MingCheng991129:
Change generated table name to avoid conflict with QuestDB reserved name.


mrigger (approved):
LGTM, thanks!

PR #722, opened 2023-02-05
Forbid multiple table selection from QuestDB

MingCheng991129:
Currently, selecting from multiple tables are regarded as illegal by QuestDB. E.g. `SELECT * FROM t0, t1;`

This will cause the low "successful statements" when running our code (~20%)
This PR forbids such operation, and increase the successful statements to 70%.

Running command: `java -jar sqlancer-*.jar --num-threads 1 questdb`

### DIFF
--- a/src/sqlancer/questdb/test/QuestDBQueryPartitioningBase.java
+++ b/src/sqlancer/questdb/test/QuestDBQueryPartitioningBase.java
@@ -67,15 +67,15 @@ public class QuestDBQueryPartitioningBase
     @Override
     public void check() throws SQLException {
         s = state.getSchema();
-        // Only return one table instead of multiple tables, which is regarded as illegal by QuestDB 
+        // Only return one table instead of multiple tables, which is regarded as illegal by QuestDB
         // e.g. "SELECT * FROM t0, t1;"
-        targetTable = s.getRandomTable(t -> !t.isView()); 
-        gen = new QuestDBExpressionGenerator(state).setColumns(targetTable.getColumns()); 
+        targetTable = s.getRandomTable(t -> !t.isView());
+        gen = new QuestDBExpressionGenerator(state).setColumns(targetTable.getColumns());
         initializeTernaryPredicateVariants();
         select = new QuestDBSelect();
         select.setFetchColumns(generateFetchColumns());
-        List<QuestDBTable> tables = new ArrayList<>(); 
-        tables.add(targetTable); 
+        List<QuestDBTable> tables = new ArrayList<>();
+        tables.add(targetTable);
         List<TableReferenceNode<QuestDBExpression, QuestDBTable>> tableList = tables.stream()
                 .map(t -> new TableReferenceNode<QuestDBExpression, QuestDBTable>(t)).collect(Collectors.toList());
         // Ignore JOINs for now
--- a/src/sqlancer/questdb/test/QuestDBQueryPartitioningBase.java
+++ b/src/sqlancer/questdb/test/QuestDBQueryPartitioningBase.java
@@ -29,6 +29,7 @@ public class QuestDBQueryPartitioningBase
 
     QuestDBSchema s;
     QuestDBTables targetTables;
+    QuestDBTable targetTable; // single table
     QuestDBExpressionGenerator gen;
     QuestDBSelect select;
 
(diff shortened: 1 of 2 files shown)


src/sqlancer/questdb/test/QuestDBQueryPartitioningBase.java
```diff
@@ -67,6 +67,11 @@ protected ExpressionGenerator<Node<QuestDBExpression>> getGen() {
     public void check() throws SQLException {
         s = state.getSchema();
         targetTables = s.getRandomTableNonEmptyTables();
+        if (targetTables.getTables().size() > 1) {
```
mrigger, line 70: I think it would be clearer to add a method `s.getRandomTable()` that directly returns a random table.

src/sqlancer/questdb/test/QuestDBQueryPartitioningWhereTester.java
```diff
@@ -18,6 +18,11 @@ public QuestDBQueryPartitioningWhereTester(QuestDBGlobalState state) {
     @Override
     public void check() throws SQLException {
         super.check();
+        if (select == null) {
```
mrigger, line 21: Here, I don't understand how it can happen that `select` is `null` and how it relates to the comment. Could you please explain?

src/sqlancer/questdb/test/QuestDBQueryPartitioningBase.java
```diff
@@ -66,12 +67,15 @@ protected ExpressionGenerator<Node<QuestDBExpression>> getGen() {
     @Override
     public void check() throws SQLException {
         s = state.getSchema();
-        targetTables = s.getRandomTableNonEmptyTables();
-        gen = new QuestDBExpressionGenerator(state).setColumns(targetTables.getColumns());
+        // Only return one table instead of multiple tables, which is regarded as illegal by QuestDB
+        // e.g. "SELECT * FROM t0, t1;"
+        targetTable = s.getRandomTable(t -> !t.isView());
```
mrigger, line 72: I think we can also consider views in the `SELECT` statement, so perhaps we can just select any random table?

src/sqlancer/questdb/test/QuestDBQueryPartitioningBase.java
```diff
@@ -66,12 +67,15 @@ protected ExpressionGenerator<Node<QuestDBExpression>> getGen() {
     @Override
     public void check() throws SQLException {
         s = state.getSchema();
-        targetTables = s.getRandomTableNonEmptyTables();
```
mrigger, line 69: Can we also remove the corresponding field, or is it still used anywhere?

src/sqlancer/questdb/test/QuestDBQueryPartitioningBase.java
MingCheng991129: OK, then just `targetTable = s.getRandomTable();` will be fine.

src/sqlancer/questdb/test/QuestDBQueryPartitioningBase.java
MingCheng991129: Well, I can delete it. But I'm considering we'll implement support for multiple tables (e.g. `Join`) in the future, so I think it's not a big deal to delete it probably?

MingCheng991129: But anyway, if it's better to delete it now, I'll do that in this PR :)

MingCheng991129: Deleted and committed.

src/sqlancer/questdb/test/QuestDBQueryPartitioningBase.java
```diff
@@ -28,7 +27,7 @@
         implements TestOracle<QuestDBGlobalState> {
 
     QuestDBSchema s;
-    QuestDBTables targetTables;
+    QuestDBTable targetTable; // single table
```
mrigger, line 30: This comment is not useful, since the type already specifies that this is a single table. The "why?" would be more interesting here. As this is a nitpick, merging the PR nevertheless.

mrigger (approved):

PR #727, opened 2023-02-11
Add SYMBOL type support for QuestDB

MingCheng991129:
Add `SYMBOL` type support for QuestDB because an index can only be created on `SYMBOL` type columns according to QuestDB.

### DIFF
--- a/src/sqlancer/questdb/QuestDBSchema.java
+++ b/src/sqlancer/questdb/QuestDBSchema.java
@@ -70,7 +70,7 @@ public class QuestDBSchema extends AbstractSchema<QuestDBGlobalState, QuestDBTab
                 isNullable = false;
                 break;
             case SYMBOL:
-                isNullable = false;
+                isNullable = true;
                 break;
             default:
                 isNullable = true;
--- a/src/sqlancer/questdb/QuestDBSchema.java
+++ b/src/sqlancer/questdb/QuestDBSchema.java
@@ -27,8 +27,7 @@ public class QuestDBSchema extends AbstractSchema<QuestDBGlobalState, QuestDBTab
 
         BOOLEAN, // CHAR,
         /* STRING, */
-        INT, FLOAT,
-        /* SYMBOL, */
+        INT, FLOAT, SYMBOL,
         // DATE, TIMESTAMP,
         /* GEOHASH, */
         NULL;
@@ -70,6 +69,9 @@ public class QuestDBSchema extends AbstractSchema<QuestDBGlobalState, QuestDBTab
             case BOOLEAN:
                 isNullable = false;
                 break;
+            case SYMBOL:
+                isNullable = false;
+                break;
             default:
                 isNullable = true;
             }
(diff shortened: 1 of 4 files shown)


src/sqlancer/questdb/QuestDBSchema.java
```diff
@@ -70,6 +69,9 @@ public QuestDBCompositeDataType(QuestDBDataType dataType, int size) {
             case BOOLEAN:
                 isNullable = false;
                 break;
+            case SYMBOL:
+                isNullable = false;
```
mrigger, line 73: According to https://questdb.io/docs/reference/sql/datatypes/, the data type should be nullable, no?

src/sqlancer/questdb/gen/QuestDBExpressionGenerator.java
```diff
@@ -54,6 +55,15 @@ public Node<QuestDBExpression> generateConstant() {
             return QuestDBConstant.createBooleanConstant(Randomly.getBoolean());
         case FLOAT:
             return QuestDBConstant.createFloatConstant(globalState.getRandomly().getDouble());
+        case SYMBOL:
+            StringGenerationStrategy strategy = Randomly.StringGenerationStrategy.ALPHANUMERIC;
+            // Add quotes to form a valid string for QuestDB
+            String str = '\"' + strategy.getString(globalState.getRandomly()) + '\"';
```
mrigger, line 61: I think it would be more elegant to format the symbol in the constant itself (e.g., see https://github.com/sqlancer/sqlancer/blob/40a1d26adaa0cc76050aa9380d271565862e4f95/src/sqlancer/duckdb/ast/DuckDBConstant.java#L79).

mrigger, line 61: In SQL, strings are typically in single quotes. Are you sure that QuestDB uses double quotes?

src/sqlancer/questdb/gen/QuestDBExpressionGenerator.java
```diff
@@ -54,6 +55,15 @@ public Node<QuestDBExpression> generateConstant() {
...
+            // When the value of symbol is not specified, set it to NULL (not "NULL")
+            if (str.length() == 2) {
```
mrigger, line 63: I think this would be easier to read if we check whether the string is empty above.

src/sqlancer/questdb/QuestDBSchema.java
```diff
@@ -142,6 +147,8 @@ public String toString() {
                 }
             case BOOLEAN:
                 return Randomly.fromOptions("BOOLEAN");
+            case SYMBOL:
+                return Randomly.fromOptions("SYMBOL");
```
mrigger, line 151: suggestion for a refactoring PR: if we only have one option, we can directly return the string (e.g., `return "SYMBOL"`).

mrigger (approved):

PR #729, opened 2023-02-15
Fix add index bug for QuestDB

MingCheng991129:
In QuestDB, the column to be added an index must be of type `SYMBOL`.

However, the current code calls `table.getFreeColumnName();` to choose a free column to add an index, instead of selecting a `SYMBOL` type column. This will cause errors when adding an index.


src/sqlancer/common/schema/AbstractTable.java
```diff
@@ -60,6 +60,15 @@ public C getRandomColumn() {
         return Randomly.fromList(columns);
     }
 
+    public C getRandomColumnWithFilter(String typeFilter) {
```
mrigger, line 63: I think this method should use a stream such as above (using the `filter` method), which means you can pass a `Predicate` that filters records. See https://www.javatpoint.com/java-8-stream-filter. Filtering based on the string representation of a type is not type-safe and thus error-prone.

src/sqlancer/questdb/QuestDBSchema.java
```diff
@@ -142,6 +147,8 @@ public String toString() {
                 }
             case BOOLEAN:
                 return Randomly.fromOptions("BOOLEAN");
+            case SYMBOL:
+                return Randomly.fromOptions("SYMBOL");
```
mrigger, line 151: Here (and above), you can directly return the string (i.e., `"SYMBOL"`). `Randomly.fromOptions` returns an element from the variadic arguments given. Since only one argument is given, the method directly returns the object.

src/sqlancer/common/schema/AbstractTable.java
```diff
@@ -60,6 +60,15 @@ public C getRandomColumn() {
...
+        for (C c : columns) {
+            if (c.getType().toString().equals(typeFilter)) {
+                return c;
+            }
+        }
+        return null;
```
mrigger, line 69: It might simplify the code that uses this method to throw an `IgnoreMeException` here. To reflect this in the method name, the method could be called `getRandomColumnOrBailout`.

src/sqlancer/questdb/gen/QuestDBAlterIndexGenerator.java
```diff
@@ -29,7 +31,17 @@ public static SQLQueryAdapter getQuery(QuestDBGlobalState globalState) {
         sb.append(" ");
 
         sb.append("ALTER COLUMN ");
-        String columnName = table.getFreeColumnName();
+
+        // We should always choose column with SYMBOL type
+        QuestDBColumn columnWithSymbolType = table.getRandomColumnWithFilter("SYMBOL");
+
+        if (columnWithSymbolType == null) {
```
mrigger, line 38: We could handle this case directly in the method (see my other comment in that method).

src/sqlancer/questdb/gen/QuestDBAlterIndexGenerator.java
```diff
@@ -29,7 +30,13 @@ public static SQLQueryAdapter getQuery(QuestDBGlobalState globalState) {
         sb.append(" ");
 
         sb.append("ALTER COLUMN ");
-        String columnName = table.getFreeColumnName();
+
+        // We should always choose column with SYMBOL type
+        QuestDBColumn columnWithSymbolType = table
+                .getRandomColumnOrBailout(c -> c.getType().toString().equals("SYMBOL"));
```
mrigger, line 36: Here, rather than a string comparison, could we directly compare the type? This has the advantage that the comparison is type-safe. Besides that, the PR looks good to me.

src/sqlancer/questdb/gen/QuestDBAlterIndexGenerator.java
```diff
@@ -33,7 +34,7 @@ public static SQLQueryAdapter getQuery(QuestDBGlobalState globalState) {
 
         // We should always choose column with SYMBOL type
         QuestDBColumn columnWithSymbolType = table
-                .getRandomColumnOrBailout(c -> c.getType().toString().equals("SYMBOL"));
+                .getRandomColumnOrBailout(c -> c.getType().getPrimitiveDataType().equals(QuestDBDataType.SYMBOL));
```
mrigger, line 37: A small nitpick: enums are typically compared using an equality operator (e.g., see https://stackoverflow.com/a/2937561 for the advantages).

mrigger (approved):

PR #735, opened 2023-02-27
Add cnosdb expect error and modify db schema

Benxiaohai001:
Add cnosdb expect error and modify db schema
image


src/sqlancer/cnosdb/CnosDBProvider.java
```diff
@@ -69,9 +69,10 @@ public CnosDBConnection createDatabase(CnosDBGlobalState globalState) throws Exc
         CnosDBClient client = new CnosDBClient(host, port, username, password, databaseName);
         CnosDBConnection connection = new CnosDBConnection(client);
         client.execute("DROP DATABASE IF EXISTS " + databaseName);
+        String createdb = "CREATE DATABASE " + databaseName + " WITH TTL '1000000' REPLICA 2";
```
mrigger, line 72: What does `WITH TTL '1000000'` mean? Do you think it might be useful to add a short comment why we override the default configuration?

mrigger (commented):
Thanks for the PR! Could you please have a look at my two minor comments?

mrigger:
Any updates on this? It would be great if we could merge the PR (which is now out-of-date with the base branch, so a merge would be necessary)!

PR #740, opened 2023-03-02
[SQLite3] Add length function

avishmehta68710:
Add support for length built in aggregate function in sqlite3

### DIFF
--- a/src/sqlancer/sqlite3/ast/SQLite3Function.java
+++ b/src/sqlancer/sqlite3/ast/SQLite3Function.java
@@ -44,6 +44,18 @@ public class SQLite3Function extends SQLite3Expression {
             }
         },
 
+        LENGTH(1, "LENGTH") {
+            @Override
+            public SQLite3Constant apply(SQLite3Constant... args) {
+                SQLite3Constant castValue = SQLite3Cast.castToText(args[0]);
+                if (castValue.isNull()) {
+                    return castValue;
+                } else {
+                    return SQLite3Constant.createIntConstant(castValue.asString().length());
+                }
+            }
+        },
+
         COALESCE(2, "COALESCE") {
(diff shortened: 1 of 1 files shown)


SuriZhang:
Please update the PR with a bit more details in your description to let reviewers know what are the changes in this PR also follow the contributing doc on requirements of commit history https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#commit-history

src/sqlancer/sqlite3/ast/SQLite3Function.java
```diff
@@ -44,6 +44,18 @@ public SQLite3Constant apply(SQLite3Constant... args) {
             }
         },
 
+        LENGTH(1, "LENGTH") {
```
mrigger, line 47: Thanks for the PR! The `apply` implementation is used only in PQS. I'm not sure if the Java string length maps directly to the SQLite one (also considering the different character encodings). Perhaps you could just return null (see https://github.com/sqlancer/sqlancer/blob/b19a41e0dcf7cbd5fd84f66694743dc12323e2ef/src/sqlancer/sqlite3/ast/SQLite3Function.java#L69), so that a `LENGTH` function is used by NoREC and TLP, but not by PQS?

mrigger:
I just noticed that we actually have a separate enum with functions for non-PQS implementations, which already includes `LENGTH` (see https://github.com/sqlancer/sqlancer/blob/5533c4e3c32a7f474bf1e047bda5f529fb9ef7ba/src/sqlancer/sqlite3/gen/SQLite3ExpressionGenerator.java#L420). Thus, I'm closing this PR. Thanks for spending time on this, and perhaps you can still identify some other functions not yet generated functions in that enum.

Issue #742, opened 2023-03-05
Add issue template 


ZhengLin-Li:
Currently our repo do not have a issue template.

GitHub supports Issue templates. Ref: [Issue template docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository).

The issue template will provide contributors with a structured entry point for reporting any issue or bug.

yatharth-singh:
i  can fix this issue. please assign this to me

mrigger:
I'm not convinced that having such a template is helpful, since we have a low number of issue reports, and little commonly required information (e.g., we currently don't even have regular releases).

PR #743, opened 2023-03-05
add issue template

ZhengLin-Li:
fix: #742 

Currently our repo do not have a issue template.

GitHub supports Issue templates. Ref: [Issue template docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository).

The issue template will provide contributors with a structured entry point for reporting any issue or bug.


mrigger:
While issue templates are a useful feature for many projects, I'm not very convinced that they are a good fit for this project (see my comment in #742), so would prefer not to add them at this point in time.

Issue #744, opened 2023-03-05
Include commit message and branch naming convensions in the `contributing.md`


ZhengLin-Li:
Currently, our [_Naming Conventions section_ of `contributing.md`](https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#naming-conventions) document lacks a branch naming convention and example commit messages.

This document plays a crucial role in maintaining a cohesive structure throughout the project and will greatly aid new contributors in familiarizing themselves with the project. Therefore, we need to incorporate standardized **commit message formats** and **branch naming conventions** in accordance with established community best practices.

PR #749, opened 2023-03-05
docs: add naming convention in CONTRIBUTING.md

ZhengLin-Li:
fix: https://github.com/sqlancer/sqlancer/issues/744

add commit message conventions and branching naming conventions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,59 +7,140 @@ Developing SQLancer using Eclipse is expected to work well. You can import SQLan
 ```
 File -> Import -> Existing Maven Projects -> Select the SQLancer directory as root directory -> Finish
 ```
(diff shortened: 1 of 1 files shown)


mrigger (commented):
Thanks for the PR! I'm doubtful about merging this, although the PR makes sense in principle. We currently already describe some conventions at https://github.com/sqlancer/sqlancer/blob/5c5440f5aa4aad1be0c6a8c31dad19e4ce8273c8/CONTRIBUTING.md#commit-history, and your examples actually violate some of the rules (e.g., using imperative and start the subject line with an uppercase letter). In addition, I realized that most people don't pay attention to the rules, which is why, for example, I've started squashing all contributions.

ZhengLin-Li:
Yes, I saw some other projects doing this, but it may not work for us. I will close it, and thanks for reviewing!

PR #752, opened 2023-03-06
Refactor several classes 

krishan30:
### **Refactor codes in several java files:**

- StateToReproduce.java
- StatementExecutor.java
- SQLProviderAdapter.java
- Randomly.java
- ProviderAdapter.java
- BinaryNode.java
- BinaryOperatorNode.java
- FunctionNode.java
- TernaryNode.java
- UnaryNode.java
- UnaryOperatorNode.java
- SQLLoggableFactory.java
- NoRECBase.java
- PivotedQuerySynthesisBase.java
- AbstractRowValue.java
- AbstractSchema.java
- AbstractTable.java
- DBMSCommon.java

### **Changes that are done :**

- Access Modifier of the abstract classes’ Constructor changed from public to protected because constructors of abstract classes can only be called in constructors of their subclasses.No need to make them public.
- Refactor a  method name due to incorrect spellings.(In DBMSCommon.java)
- Simplified some of the codes.
- Added some comments.

### DIFF
--- a/src/sqlancer/StatementExecutor.java
+++ b/src/sqlancer/StatementExecutor.java
@@ -70,7 +70,7 @@ public class StatementExecutor<G extends GlobalState<?, ?, ?>, A extends Abstrac
                     success = globalState.executeStatement(query);
                 } while (nextAction.canBeRetried() && !success
                         && nrTries++ < globalState.getOptions().getNrStatementRetryCount());
-            } catch (IgnoreMeException e) {
+            } catch (IgnoreMeException ignored) {
 
             }
             if (query != null && query.couldAffectSchema()) {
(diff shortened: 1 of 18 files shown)


mrigger (approved):
Thanks for the PR, LGTM! Could you please format the files using `mvn formatter:format` so that the CI check succeeds?

krishan30:
> Thanks for the PR, LGTM! Could you please format the files using `mvn formatter:format` so that the CI check succeeds?

mvn formatter:format Done

src/sqlancer/StatementExecutor.java
```diff
@@ -70,7 +70,7 @@ public void executeStatements() throws Exception {
                     success = globalState.executeStatement(query);
                 } while (nextAction.canBeRetried() && !success
                         && nrTries++ < globalState.getOptions().getNrStatementRetryCount());
-            } catch (IgnoreMeException e) {
+            } catch (IgnoreMeException ignored) {
```
ZhengLin-Li, line 73: Generally, it is not recommended to use an empty catch to silence an exception.

But in our case, do we need to print the error message or just ignore it?

krishan30:
Sorry I had a checkstyle error and I fixed it.Now I think it's good go.

mrigger (approved):
LGTM!

Issue #698, opened 2023-01-19
Citus CI failing


mrigger:
The Citus CI has been failing for a while now. For example, https://github.com/sqlancer/sqlancer/actions/runs/3956119066/jobs/6776946798 shows failures as outlined below.

```
CREATE DATABASE database2 ; --- ERROR: database "database2" is being accessed by other users
``` 

```sql
INSERT INTO t1(c0) VALUES(B'1111111111111111111111111111111111101000111000001001100000100110'), (B'1111111111111111111111111111111111010111011111001111111000100000'), (B'1111111111111111111111111111111110001100100110000101001001100011') ON CONFLICT  DO NOTHING; --- ERROR: columnar_tuple_insert_speculative not implemented
```

mrigger:
Previously reported an issue at https://github.com/citusdata/citus/issues/6670, which might have fixed this. cc @malwaregarry

malwaregarry:
Still showing up on v11.3  and v12.1. 

Perhaps we can add the error to ExpectedError or do  some rate limiting?

mrigger:
> Perhaps we can add the error to ExpectedError

I think this would just mask the issue. I think it would be better to try addressing it.

CC'ing @onurctirtir and @gokhangulbiz, since they worked on the issue/fix (see https://github.com/citusdata/citus/issues/6670).

malwaregarry:
resolved in #960 by disconnecting clients first

PR #755, opened 2023-03-09
feat: add missing python naming convention test

ZhengLin-Li:
feat: add missing python naming convention test

A log of an example run can be found here: https://github.com/ZhengLin-Li/sqlancer/actions/runs/4371501997/jobs/7647428741

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,6 +25,12 @@ jobs:
         run: mvn -B verify -DskipTests=true
       - name: Misc Tests
         run: mvn -B '-Dtest=!sqlancer.dbms.**' test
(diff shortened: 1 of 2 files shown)


mrigger (approved):
LGTM, thanks!

ZhengLin-Li:
@mrigger hi, I found that these two CI fails on every PR. Could you please have a look on what's the reason and do we need to fix them?

image

mrigger:
We have two issues open on this: https://github.com/sqlancer/sqlancer/issues/698 and https://github.com/sqlancer/sqlancer/issues/707. I currently don't have time to investigate these. If you want, you can have a look.

PR #757, opened 2023-03-11
feat: Implement delta-debugging as the reduction strategy

ColinYoungTaro:
Hello, I'm interested in the idea **test case reduction** in GSOC. This PR  implements a simple delta-debugging statement reduction strategy.

I'm new in this project and the open source community. I would appreicate it if you can point out my mistakes and offer me some guidance. Thanks :)

### DIFF
--- a/src/sqlancer/StatementReducer.java
+++ b/src/sqlancer/StatementReducer.java
@@ -2,7 +2,6 @@ package sqlancer;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
 import sqlancer.common.query.Query;
@@ -11,9 +10,15 @@ public class StatementReducer<G extends GlobalState<O, ?, C>, O extends DBMSSpec
         implements Reducer<G> {
     private final DatabaseProvider<G, O, C> provider;
     private boolean observedChange;
+    private int partitionNum;
 
     public StatementReducer(DatabaseProvider<G, O, C> provider) {
         this.provider = provider;
+    }
+
+    @FunctionalInterface
+    private interface TriFunction<A, B, C, R> {
+        R apply(A a, B b, C c);
     }
 
     @SuppressWarnings("unchecked")
@@ -28,14 +33,28 @@ public class StatementReducer<G extends GlobalState<O, ?, C>, O extends DBMSSpec
         printQueries(knownToReproduceBugStatements);
         System.out.println();
 
-        do {
+        if (knownToReproduceBugStatements.size() <= 1) {
+            return;
+        }
+
+        partitionNum = 2;
+        while (knownToReproduceBugStatements.size() >= 2) {
             observedChange = false;
+
             knownToReproduceBugStatements = tryReduction(state, reproducer, newGlobalState,
-                    knownToReproduceBugStatements, (candidateStatements, i) -> {
-                        candidateStatements.remove((int) i);
+                    knownToReproduceBugStatements, (candidateStatements, s, l) -> {
+                        List<Query<C>> remainedStatements = new ArrayList<>();
+                        candidateStatements.subList(s, s+l).clear();
                         return true;
                     });
-        } while (observedChange);
+            if (!observedChange) {
+                if (partitionNum == knownToReproduceBugStatements.size()) {
+                    break;
+                }
+                // increase the search granularity
+                partitionNum = Math.min(partitionNum * 2, knownToReproduceBugStatements.size());
+            }
+        }
 
         System.out.println("Reduced query:");
         printQueries(knownToReproduceBugStatements);
@@ -43,17 +62,24 @@ public class StatementReducer<G extends GlobalState<O, ?, C>, O extends DBMSSpec
 
     private List<Query<C>> tryReduction(G state, // NOPMD
             Reproducer<G> reproducer, G newGlobalState, List<Query<C>> knownToReproduceBugStatements,
-            BiFunction<List<Query<C>>, Integer, Boolean> reductionOperation) throws Exception {
+            TriFunction<List<Query<C>>, Integer, Integer, Boolean> reductionOperation) throws Exception {
 
         List<Query<C>> statements = knownToReproduceBugStatements;
-        for (int i = 0; i < statements.size(); i++) {
+
+        int start = 0;
+        int subLength = statements.size() / partitionNum;
+        while (start < statements.size()) {
+            // newStatements = candidate[:start] + candidate[start+subLength:]
+            // in other word, remove [start, start+subLength) from candidates
             try (C con2 = provider.createDatabase(newGlobalState)) {
                 newGlobalState.setConnection(con2);
                 List<Query<C>> candidateStatements = new ArrayList<>(statements);
-                if (!reductionOperation.apply(candidateStatements, i)) {
+                if (!reductionOperation.apply(candidateStatements, start, subLength)) {
+                    start = start + subLength;
                     continue;
                 }
-                newGlobalState.getState().setStatements(candidateStatements.stream().collect(Collectors.toList()));
+                newGlobalState.getState().setStatements(new ArrayList<>(candidateStatements));
+
                 for (Query<C> s : candidateStatements) {
                     try {
                         s.execute(newGlobalState);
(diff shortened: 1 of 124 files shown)


ZhengLin-Li:
Hi, @ColinYoungTaro Thanks for the PR! 🥳🙌

It seems that you are implementing https://en.wikipedia.org/wiki/Delta_debugging ? Overall, the changes look good, but I have several questions and suggestions.  I have left them in code review.

I cannot tell if the logic is fully correct, so other reviewers please DO check the logic part carefully.

oraluben:
Can you elaborate a little bit on what you've done (and maybe future plan to implement the reduction)?

mrigger:
Thanks a lot for the PR!

@ZhengLin-Li, I think you might have not yet made your review comments visible, as I cannot see them.

src/sqlancer/StatementReducer.java
```diff
@@ -28,32 +33,59 @@ public void reduce(G state, Reproducer<G> reproducer, G newGlobalState) throws E
         printQueries(knownToReproduceBugStatements);
         System.out.println();
 
-        do {
+        if (knownToReproduceBugStatements.size() <= 1) {
+            return;
+        }
+
+        partitionNum = 2;
+        while (knownToReproduceBugStatements.size() >= 2) {
```
ZhengLin-Li, line 41: Maybe we can set a threshold so that only when `knownToReproduceBugStatements.size()` is very large do we use this method, and we use the normal full loop to handle cases when `knownToReproduceBugStatements.size()` is relatively small. 

JDK's implementation of the sort() algorithm is similar. If this sounds reasonable, we can discuss the threshold with @mrigger 


image

src/sqlancer/StatementReducer.java
```diff
@@ -28,32 +33,59 @@ public void reduce(G state, Reproducer<G> reproducer, G newGlobalState) throws E
...
             observedChange = false;
+
             knownToReproduceBugStatements = tryReduction(state, reproducer, newGlobalState,
-                    knownToReproduceBugStatements, (candidateStatements, i) -> {
-                        candidateStatements.remove((int) i);
+                    knownToReproduceBugStatements, (candidateStatements, s, l) -> {
```
ZhengLin-Li, line 45: I'd suggest that we use the variables `startIndex` and `subLength` for better code readability. But that's just my opinion, we can discuss with @mrigger

src/sqlancer/StatementReducer.java
```diff
@@ -28,32 +33,59 @@ public void reduce(G state, Reproducer<G> reproducer, G newGlobalState) throws E
...
     private List<Query<C>> tryReduction(G state, // NOPMD
             Reproducer<G> reproducer, G newGlobalState, List<Query<C>> knownToReproduceBugStatements,
-            BiFunction<List<Query<C>>, Integer, Boolean> reductionOperation) throws Exception {
+            TriFunction<List<Query<C>>, Integer, Integer, Boolean> reductionOperation) throws Exception {
 
         List<Query<C>> statements = knownToReproduceBugStatements;
-        for (int i = 0; i < statements.size(); i++) {
+
+        int start = 0;
+        int subLength = statements.size() / partitionNum;
```
ZhengLin-Li, line 76: The variable `subLength` is outside of the loop. Doesn't it need to change when the loop goes on?

src/sqlancer/StatementReducer.java
```diff
@@ -28,32 +33,59 @@ public void reduce(G state, Reproducer<G> reproducer, G newGlobalState) throws E
...
+        while (start < statements.size()) {
+            // newStatements = candidate[:start] + candidate[start+subLength:]
+            // in other word, remove [start, start+subLength) from candidates
```
ZhengLin-Li, line 79: I am confused by this comment. From lines L46-L53, we actually add [start, start + subLength] to candidate statements. Do I get your idea correctly?

src/sqlancer/StatementReducer.java
```diff
@@ -28,32 +33,59 @@ public void reduce(G state, Reproducer<G> reproducer, G newGlobalState) throws E
...
             try (C con2 = provider.createDatabase(newGlobalState)) {
                 newGlobalState.setConnection(con2);
                 List<Query<C>> candidateStatements = new ArrayList<>(statements);
-                if (!reductionOperation.apply(candidateStatements, i)) {
+                if (!reductionOperation.apply(candidateStatements, start, subLength)) {
```
ZhengLin-Li, line 83: The `reductionOperation.apply(candidateStatements, start, subLength)` always returns true. Even though the old code does the same, need we improve this?

ZhengLin-Li (changes requested):

ColinYoungTaro:
> Can you elaborate a little bit on what you've done (and maybe future plan to implement the reduction)?


The current 'tryReduction' function removes one statement each time and execute the rest to see if the bug is still triggered. I tried implementing a simple delta-debugging reducer, which may runs faster

I think it could done as follows in the future
+ the current delta-debugging implementation is simple and hard-coded, it could be made cleaner and more effective.
+ Besides statment reducer, it would be better to apply transformations like CReduce does. i.e. removing some tokens.
+ generalize to more DBMSs and testing strategy

The plans above might have somewhere improper or wrong. I would be grateful if you can provide me some suggenstions.

src/sqlancer/StatementReducer.java
ColinYoungTaro: Delta-debugging split the hole array into several blocks and try remove one of them to see if the same error is still here. The block could be large at the begining(e.g. half of the array size) if each block removal let the bug disappear, shrink the block size.

```subLength``` is the block size. It determines how many statements should be dropped. So it should be fixed in the inner loop.

File: src/sqlancer/StatementReducer.java (hunk 4 above)
Comment (ColinYoungTaro) [reply src/sqlancer/StatementReducer.java:79]:
```
if (i >= s + l || i < s) {
      removed.add(candidateStatements.get(i));
}
candidateStatements.clear();
candidateStatements.addAll(removed);
```

It's my fault to name it too casually.  We keep the [: start) and [start+subLength:] actually.

src/sqlancer/StatementReducer.java
ColinYoungTaro: Thanks. I think it sould be improved.

mrigger:
@oraluben, did you already have a chance to have a closer look at the PR? Thanks also for your help, @ZhengLin-Li!

mrigger:
I'm wondering if we could make the reduction code somewhat testable? I think this would be needed at some point before enabling it by default, although it might be difficult to do so.

oraluben:
> @oraluben, did you already have a chance to have a closer look at the PR? Thanks also for your help, @ZhengLin-Li!

Yes, I've verified the PR works as describe. I'm also considering what's the best way to implement a easy-to-use test case reduction.

oraluben:
There're two way in my mind: as suggested offline by @mrigger , to invoke third-party reducer like creduce, or to implement a new sql-specific reducer and use it w./wo. sqlancer.

ColinYoungTaro:
> I'm wondering if we could make the reduction code somewhat testable? I think this would be needed at some point before enabling it by default, although it might be difficult to do so.

Thanks.
In the beginning I ran the sqlancer but failed to find bugs. So when I was testing my delta-debugging prototype, I copied a number of SQL statements from the log and randomly mark some of them, assuming that a bug would be triggered only if the marked statements were exectued together. Then I run the reduction to see if the marked statements were kept. Though somewhat naive, I think it might work for testing a statement reducer to some extent.

Also，the reduction had better remove some AST nodes in a statement. It might be tricky to generate test case. I am wondering if we could use an early version of DBMS. It might be easier to find bugs so that we can record the bug-triggering statements and reduce them?

oraluben (approved):
LGTM. Currently I test test case reduction by using an old version (3.28.0) of sqlite3 and a fixed random seed, so the first query will be same.

mrigger:
Great, perhaps you can update to the latest master version (e.g., by merging), and then we can merge the PR?

mrigger:
It seems the style-check test is currently failing. You can check the warnings using `mvn -B verify -DskipTests=true`.

oraluben:
Suppressed by #815

PR #770, opened 2023-03-26
Update Dockerfile

AtulRajput01:
1.Removed the --yes flag and combined it with the --no-install-recommends flag to reduce the image size and remove unnecessary dependencies.
2.Removed ENV statement and replaced it with the && operator to reduce the number of layers in the Docker image.
3.Added a RUN statement to remove the apt-get package lists after installation to reduce image size.
4.Added a non-root user and changed the owner of the /home/sqlancer directory to the sqlancer user for improved security.
5.Changed the working directory to /home/sqlancer and updated the file paths accordingly.
6.Added the USER command to switch to the sqlancer user for improved security.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,21 @@
 FROM ubuntu:21.04
 
-RUN apt-get update --yes && env DEBIAN_FRONTEND=noninteractive apt-get install openjdk-15-jdk maven --yes --no-install-recommends
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    openjdk-15-jdk \
+    maven && \
+    rm -rf /var/lib/apt/lists/*
 
-# assumes that the project has already been built
+# Create a non-root user
+RUN useradd -m sqlancer && \
+    chown -R sqlancer:sqlancer /home/sqlancer
+
+# Copy the project files
+WORKDIR /home/sqlancer
 COPY target/sqlancer-*.jar sqlancer.jar
 COPY target/lib/*.jar /lib/
+RUN chown -R sqlancer:sqlancer /home/sqlancer
+
+USER sqlancer
 
 ENTRYPOINT ["java", "-jar", "sqlancer.jar"]


oraluben:
In general, I'm not sure if this PR is necessary, especially creating a new user.

AtulRajput01:
> In general, I'm not sure if this PR is necessary, especially creating a new user.
@oraluben sir  I understand your concern about the necessity of creating a new user, but creating a non-root user is considered a best practice for security reasons. By running the container as a non-root user, we reduce the risk of potential security vulnerabilities that could be exploited by attackers to gain root-level access to the container and the host system.

AtulRajput01:
> In general, I'm not sure if this PR is necessary, especially creating a new user.

and removing unnecessary dependencies and package lists can significantly reduce the size of the Docker image, which can be beneficial for deployment and distribution.

Overall, while the changes may not be strictly necessary, they can improve the security and efficiency of the Docker image, and I believe they are worth considering.

Dockerfile
```diff
@@ -1,9 +1,21 @@
 FROM ubuntu:21.04
 
-RUN apt-get update --yes && env DEBIAN_FRONTEND=noninteractive apt-get install openjdk-15-jdk maven --yes --no-install-recommends
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    openjdk-15-jdk \
+    maven && \
+    rm -rf /var/lib/apt/lists/*
```
bajinsheng, line 7: What does it do?

bajinsheng: Why remove it?

AtulRajput01: > What does it do?   
@bajinsheng  sir command apt-get install -y --no-install-recommends openjdk-15-jdk maven installs the OpenJDK 15 and Maven packages on the Docker image. The -y flag automatically answers "yes" to any prompts during installation, and the --no-install-recommends flag tells apt-get not to install any recommended packages.  and Maven is a build automation tool used primarily for Java projects. This can help reduce the number of unnecessary packages and dependencies installed on the image, which can in turn help reduce the image size and decrease the risk of potential security vulnerabilities.

AtulRajput01: > Why remove it?

And  "rm -rf /var/lib/apt/lists/*" removes the package lists downloaded by apt-get during the installation of the packages. These package lists are not needed after the packages have been installed, and removing them can significantly reduce the size of the Docker image .

bajinsheng: Thanks for your PR. However, I think the user permission and image size are not significant issues. These commands somehow increase the complexity.

AtulRajput01: > Thanks for your PR. However, I think the user permission and image size are not significant issues. These commands somehow increase the complexity.

I understand your perspective and respect your decision. Thank you for considering my suggestions, and please let me know if there are any other changes or improvements you would like to see.

PR #771, opened 2023-03-26
Added a demo website for SQLancer along with a tutorial to generate test cases for a table.

rashika170:
I have created a demo website for SQLancer that provides resources, documentation, and a GitHub page. While the website is still a work in progress. I have also added a beginner-friendly tutorial on generating test cases for an employee table using SQLancer. This website will be a valuable resource for those looking to learn more about SQLancer.


mrigger:
Thanks for the effort! Nevertheless closing this, as there is a mismatch between what SQLancer does and how it is used, and what is suggested in the tutorial.

PR #773, opened 2023-03-29
Add Materialize support

def-:
This is the code state I used to run SQLancer against [Materialize](https://github.com/MaterializeInc/materialize): https://github.com/MaterializeInc/materialize/issues?q=is%3Aissue+author%3Adef-+in%3Atitle+sqlancer

I started by taking Postgres and removing all functionality we don't support in Materialize (yet). My original changes were kind of messily done in https://github.com/sqlancer/sqlancer/compare/master...def-:sqlancer:master

Now that it's running well enough for us, the next step is to clean it up and get the code into upstream SQLancer. Should I remove the commented out parts of code? Anything else that is required?

The only change I did in common/ is to change the index pattern to contain the table name, otherwise I noticed indexes were attempted to be created for different tables with same name, seems to be the same on Postgres backend.

I plan to add this to SQLancer CI too.


mrigger:
Thanks a lot for the PR! Will review asap. Could you also rebase/merge on the latest commit? Not having the current conflict would allow us to run the CI tests.

> I started by taking Postgres and removing all functionality we don't support in Materialize (yet).

Some other implementations like Citus inherit from the Postgres implementation. I assume that doesn't make sense here, since Materialize is not based on Postgres, right?

> Should I remove the commented out parts of code?

I think it would be clearer to remove commented-out code (unless you, e.g., want to re-add some parts as a next step, which might unnecessarily cause additional work for you).

> The only change I did in common/ is to change the index pattern to contain the table name, otherwise I noticed indexes were attempted to be created for different tables with same name, seems to be the same on Postgres backend.

I'm a bit hesitant to change the naming even if it improves clarity, since some existing tools and implementations might rely on the naming scheme as some sort of API. I think some of the existing implementations explicitly check which index names are not present yet, which might be an option to address this.

mrigger:
Great, thanks! The style checks are currently failing. You can use `mvn formatter:format` to format the files and `mvn verify -DskipTests=true` to check for other potential warnings.

configs/spotbugs-exclude.xml
```diff
@@ -7,6 +7,11 @@
 		<Bug
 			pattern="DM_DEFAULT_ENCODING,SQL_NONCONSTANT_STRING_PASSED_TO_EXECUTE" />
 	</Match>
+	<Match>
```
mrigger, line 10: I think it would have been cleaner to make the `generateOnlyKnown` field a member variable. This would also make the code a bit more future-proof (e.g., in case we create multiple providers with different options).

src/sqlancer/materialize/MaterializeExpectedValueVisitor.java
```diff
@@ -0,0 +1,163 @@
+package sqlancer.materialize;
+
+import sqlancer.materialize.ast.MaterializeAggregate;
+import sqlancer.materialize.ast.MaterializeBetweenOperation;
+import sqlancer.materialize.ast.MaterializeBinaryLogicalOperation;
+import sqlancer.materialize.ast.MaterializeCastOperation;
+//import sqlancer.materialize.ast.MaterializeCollate;
```
mrigger, line 7: nitpick: we could remove the comment

src/sqlancer/materialize/MaterializeGlobalState.java
```diff
@@ -0,0 +1,292 @@
...
+                    collNames.add(rs.getString(1));
+                }
+            }
+        }
+        return collNames;
+    }
+
+    private List<String> getOpclasses() throws SQLException {
+        List<String> opClasses = new ArrayList<>();
+        // select opcname FROM pg_opclass;
```
mrigger, line 59: In the future, one option would be to query this information from the database system directly to reduce maintenance effort (I think no SQLancer implementations do this yet).

src/sqlancer/materialize/MaterializeToStringVisitor.java
```diff
@@ -0,0 +1,345 @@
...
+    public void visit(MaterializePrefixOperation op) {
+        sb.append(op.getTextRepresentation());
+        sb.append(" (");
+        visit(op.getExpression());
+        sb.append(")");
+    }
+
+    @Override
+    public void visit(MaterializeFromTable from) {
+        // if (from.isOnly()) {
```
mrigger, line 73: nitpick: commented-out code that could be removed

src/sqlancer/materialize/MaterializeToStringVisitor.java
```diff
@@ -0,0 +1,345 @@
...
+                sb.append(") ");
+            }
+            break;
+        case ALL:
+            sb.append(Randomly.fromOptions("ALL ", ""));
+            break;
+        default:
+            throw new AssertionError();
+        }
+        // if (s.getFetchColumns() == null) {
```
mrigger, line 108: nitpick: commented-out code that could be removed

src/sqlancer/materialize/gen/MaterializeIndexGenerator.java
```diff
@@ -0,0 +1,80 @@
...
+        errors.add("could not create unique index");
+        errors.add("has no default operator class");
+        errors.add("does not support");
+        errors.add("does not support casting");
+        errors.add("unsupported UNIQUE constraint with partition key definition");
+        errors.add("insufficient columns in UNIQUE constraint definition");
+        errors.add("invalid input syntax for");
+        errors.add("must be type ");
+        errors.add("integer out of range");
+        errors.add("division by zero");
```
mrigger, line 71: some of these seem like errors caused by expression operators? in principle, such errors could perhaps be shared across generators

mrigger (approved):
Thanks a lot again for the PR! Overall, it looks good to me. I added some minor comments. You can check if you want to address any of them now - if not, we can also directly merge the PR.

Hope the PQS oracle implementation wasn't too painful - it can be difficult to get it correct without any false alarms caused by implementation errors. :)

@bajinsheng, if you have time, perhaps you can also have a quick look at the QPG implementation for Materialize?

src/sqlancer/materialize/gen/MaterializeIndexGenerator.java
def-: Thanks, I'll check for cleaning up these in a follow-up. I'm planning to add more functions/operators anyway, can do that together probably.

configs/spotbugs-exclude.xml
def-: It seems to be used quite a lot. I copied this from Postgres. If you have a clean fix for it there, I can do it the same way in Mz.

def-:
Can be submitted from my side, thanks for the review! I'll do follow-ups to extend functionality when time permits.

mrigger:
Sounds good! I just noticed that the Materialize CI tests are failing (e.g., `ERROR: unknown catalog item 't0'` when querying table `t0` in a query). Are they also failing locally? Could you perhaps fix or ignore these, so that the tests pass?

def-:
Ok, so this is also happening in CI. I saw this locally when running `MATERIALIZE_AVAILABLE=true mvn test -Dtest=TestMaterialize`, but then tried the command (`mvn package -DskipTests && cd ~/git/sqlancer/target && java -jar sqlancer-*.jar --random-seed 0 --timeout-seconds 300  --num-threads 4 --username materialize materialize --set-max-tables-mvs true`) itself a few times and could never reproduce it. Does running `mvn test` do anything different that could explain this? I couldn't reproduce any of those issues using the printed reproducer as well, so it could be concurrency-related.

def-:
Done

def-:
I think I figured it out. By running the Main method twice in the same process we get some previously initialized state. I have split it into a separate TestMaterializePQS and haven't seen the issue since.

mrigger:
Thanks a lot again for contributing the Materialize component! I just merged the PR. Two CI tests got stuck (MySQL and MariaDB), but I think this is not related to the change.

PR #774, opened 2023-04-05
tidb: support create tiflash replicas

hawkingrei:
As the title discipline, TiDB uses tiflash to accelerate the TP. so we need to use command to enable when to test it.


bajinsheng:
The style checks are currently failing. You can use `mvn formatter:format` to format the files and `mvn verify -DskipTests=true` to check for other potential warnings.

PR #776, opened 2023-04-07
tidb: support tidb_enable_non_prepared_plan_cache

hawkingrei:


mrigger:
It seems the style check test is currently still failing?

hawkingrei:
> It seems the style check test is currently still failing?

I find I must set nonPreparePlanCache true. But I want his default value to be false. Because it is an experiment function.

mrigger:
I think the style check complains because you don't need to initialize the field to false - it is false without an explicit initialization.

hawkingrei:
> I think the style check complains because you don't need to initialize the field to false - it is false without an explicit initialization.

I fixed it.

PR #780, opened 2023-04-24
Update runner to latest version of Ubuntu

bajinsheng:
GitHub stopped supporting Ubuntu 18.04 on April 1, 2023, see:

https://github.com/actions/runner-images/issues/6002

Updating to latest version of Ubuntu to avoid the error "Waiting for a runner to pick up this job" when running the CI.


bajinsheng:
Seems the new version of Ubuntu incurs compatible issues. Requires more work.

PR #785, opened 2023-05-19
fix: fix ci error QPG Tests (Materialize)

ZhengLin-Li:
CI error: https://github.com/sqlancer/sqlancer/actions/runs/5012604712/jobs/8984740893

CI run after fix:
https://github.com/sqlancer/sqlancer/actions/runs/5020796096/jobs/9002586676?pr=785

### DIFF
--- a/src/sqlancer/materialize/MaterializeProvider.java
+++ b/src/sqlancer/materialize/MaterializeProvider.java
@@ -240,8 +240,10 @@ public class MaterializeProvider extends SQLProviderAdapter<MaterializeGlobalSta
         while (globalState.getSchema().getDatabaseTables().size() < numTables) {
             try {
                 String tableName = DBMSCommon.createTableName(globalState.getSchema().getDatabaseTables().size());
+                SQLQueryAdapter dropTable = new SQLQueryAdapter("DROP TABLE IF EXISTS " + tableName);
                 SQLQueryAdapter createTable = MaterializeTableGenerator.generate(tableName, globalState.getSchema(),
                         generateOnlyKnown, globalState);
+                globalState.executeStatement(dropTable);
                 globalState.executeStatement(createTable);
             } catch (IgnoreMeException e) {
 


src/sqlancer/materialize/MaterializeProvider.java
```diff
@@ -240,8 +240,10 @@ protected void createTables(MaterializeGlobalState globalState, int numTables) t
         while (globalState.getSchema().getDatabaseTables().size() < numTables) {
             try {
                 String tableName = DBMSCommon.createTableName(globalState.getSchema().getDatabaseTables().size());
```
mrigger, line 242: Is the problem that tables might get dropped, so the consecutive numbering doesn't work? If so, perhaps we should actually create a method that picks a free table name?

ZhengLin-Li: I do not understand why the tables might get dropped. Could you point out which function will delete tables during generating database?

mrigger: I actually didn't find any. Some other implementations implement the `DROP TABLE` statement (e.g., see https://github.com/sqlancer/sqlancer/blob/master/src/sqlancer/tidb/gen/TiDBDropTableGenerator.java), but I couldn't find this implemented for Materialize.

@def- do you have any ideas?

def-: I think this might be a legitimate Mz product issue. I started seeing this too recently, but haven't investigated. I'll check if I can reproduce it and bisect the product change that caused this.

def-: Done: https://github.com/MaterializeInc/materialize/issues/19431

bajinsheng: Seems we can update the version of Materialize to solve the issue?

mrigger: @def-, great, thanks!

@ZhengLin-Li and @bajinsheng, perhaps we can use the pattern at https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#unfixed-bugs until the issue has been addressed?

ZhengLin-Li: @def- can we update Mz product version to adress this issue? If not, let me use the pattern at https://github.com/sqlancer/sqlancer/blob/master/CONTRIBUTING.md#unfixed-bugs

bajinsheng: @ZhengLin-Li In this case, perhaps you can use the bug pattern to bypass it?

ZhengLin-Li: > 

OK, let's try to update the Mz product version one week later(2023.05.30) to see if it is fixed. If not, I will use the bug pattern to bypass it.

def-: I think @ZhengLin-Li was arguing to see if this actually fixed the issue. But I think waiting for a new release is not required, since this change only lived in SQLancer and the CI run was green with it and is also green on master now (for Materialize): https://github.com/sqlancer/sqlancer/actions/runs/5063158641 So can be closed from my perspective.

mrigger: Sounds good, closing!

Issue #781, opened 2023-04-24
Update ubuntu version to latest in CI


bajinsheng:
GitHub stopped supporting Ubuntu 18.04 on April 1, 2023, see:

https://github.com/actions/runner-images/issues/6002

Updating to latest version of Ubuntu to avoid the error "Waiting for a runner to pick up this job" when running the CI.

bajinsheng:
Great!

PR #787, opened 2023-05-19
chore: upgrade versions in CI

ZhengLin-Li:
fix: #781

upgrade ubuntu to latest
upgrade actions/checkout@v3
upgrade actions/setup-java@v1.4.4
change mysql on ubuntu18 to ubuntu22

note that the CI error now with MySQL is resulted by the `Caused by: java.sql.SQLException: Memory capacity exceeded (capacity 8388608 bytes)`, I think this is not caused by this update and should be ignored?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -300,7 +300,8 @@ jobs:
       - name: Build SQLancer
         run: mvn -B package -DskipTests=true
       - name: Setup MySQL
-        uses: mirromutth/mysql-action@v1.1
+        run: |
+          sudo systemctl start mysql.service # MySQL 8.0.32-0ubuntu0.22.04.2
       - name: Create SQLancer user
         run: mysql -uroot -proot -e "CREATE USER 'sqlancer'@'localhost' IDENTIFIED BY 'sqlancer'; GRANT ALL PRIVILEGES ON * . * TO 'sqlancer'@'localhost';"
       - name: Run Tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: Setup JDK 11
+      - name: Set up JDK 11
         uses: actions/setup-java@v1.4.4
         with:
           java-version: 11
@@ -25,7 +25,7 @@ jobs:
         run: mvn -B verify -DskipTests=true
       - name: Misc Tests
         run: mvn -B '-Dtest=!sqlancer.dbms.**,!sqlancer.qpg.**' test
-      - name: Setup Python
+      - name: Set up Python
         uses: actions/setup-python@v4
         with:
           python-version: '3.10'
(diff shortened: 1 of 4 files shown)


src/sqlancer/mysql/gen/MySQLSetGenerator.java
```diff
@@ -66,8 +66,9 @@ private enum Action {
         MAX_SP_RECURSION_DEPTH("max_sp_recursion_depth", (r) -> r.getLong(0, 255), Scope.GLOBAL, Scope.SESSION), //
         MYISAM_DATA_POINTER_SIZE("myisam_data_pointer_size", (r) -> r.getLong(2, 7), Scope.GLOBAL), //
         MYISAM_MAX_SORT_FILE_SIZE("myisam_max_sort_file_size", (r) -> r.getLong(0, 9223372036854775807L), Scope.GLOBAL), //
-        MYISAM_REPAIR_THREADS("myisam_repair_threads", (r) -> r.getLong(1, Long.MAX_VALUE), Scope.GLOBAL,
-                Scope.SESSION), //
+        // MYISAM_REPAIR_THREADS("myisam_repair_threads", (r) -> r.getLong(1, Long.MAX_VALUE), Scope.GLOBAL,
```
mrigger, line 69: Was this removed from a recent MySQL version? If so, perhaps we can just remove the code rather than commenting it out.

.github/workflows/main.yml
```diff
@@ -14,11 +14,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: Set up JDK 11
-        uses: actions/setup-java@v1
+      - name: Setup JDK 11
```
mrigger, line 20: I believe the prior spelling was actually correct, given that that "set up" is used as a verb here rather than as a noun, in which case your spelling would have been correct? See https://www.merriam-webster.com/dictionary/setup.

mrigger:
> I think this is not caused by this update and should be ignored?

It's a bit surprising and might warrant further investigation. Perhaps it might be a new bug in MySQL? Can you reproduce this locally?

.github/workflows/main.yml
ZhengLin-Li: Oh yes, I agree.

ZhengLin-Li:
> > I think this is not caused by this update and should be ignored?
> 
> It's a bit surprising and might warrant further investigation. Perhaps it might be a new bug in MySQL? Can you reproduce this locally?

OK. Maybe we should test it on a machine with larger RAM. But I do not have such machine locally or on the cloud, could you reproduce it?

The Hardware specification for Linux virtual machines on which GitHub Action run is: 2-core CPU (x86_64); 7 GB of RAM; 14 GB of SSD space

refer: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

oraluben:
> But I do not have such machine locally or on the cloud

[tmate](https://github.com/marketplace/actions/debugging-with-tmate) would always be your best friend :)

mrigger:
> Maybe we should test it on a machine with larger RAM.

The error message suggests that 8 MB exceeds the limit, so I think it's not a hardware problem.

What about merging this PR for now, since it fixes the problem of outdated dependencies, and looking into this as part of a follow-up issue or PR?

> [tmate](https://github.com/marketplace/actions/debugging-with-tmate) would always be your best friend :)

Cool, I also need to check this out!

ZhengLin-Li:
ok，lets merge it first

PR #790, opened 2023-05-23
ClickHouse: set allow_suspicious_indices

qoega:
We just can generate constant primary key and similar suspicious indexes in schema. Just skip this check as it is not a bug in generator.

### DIFF
--- a/src/sqlancer/clickhouse/ClickHouseProvider.java
+++ b/src/sqlancer/clickhouse/ClickHouseProvider.java
@@ -143,8 +143,8 @@ public class ClickHouseProvider extends SQLProviderAdapter<ClickHouseGlobalState
         }
         con.close();
         con = DriverManager.getConnection(
-                String.format("jdbc:clickhouse://%s:%d/%s?socket_timeout=300000&allow_suspicious_indices=1%s", host,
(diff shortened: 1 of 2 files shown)


src/sqlancer/clickhouse/ClickHouseProvider.java
```diff
@@ -143,8 +143,8 @@ public SQLConnection createDatabase(ClickHouseGlobalState globalState) throws SQ
         }
         con.close();
         con = DriverManager.getConnection(
-                String.format("jdbc:clickhouse://%s:%d/%s?socket_timeout=300000%s", host, port, databaseName,
-                        clickHouseOptions.enableAnalyzer ? "&allow_experimental_analyzer=1" : ""),
+                String.format("jdbc:clickhouse://%s:%d/%s?allow_suspicious_indices=1&socket_timeout=300000%s%s", host,
```
ZhengLin-Li, line 146: why should we change `socket_timeout=300000%s` to `socket_timeout=300000%s%s`

mrigger: I think it's because of the additional argument `clickHouseOptions.enableAnalyzer ? "&allow_experimental_analyzer=1" : ""`, but not sure why the tests are currently failing.

qoega:
I think it is finally ready.

mrigger (approved):
LGTM, thanks!

PR #793, opened 2023-05-24
Add query logging for MySQL

arvind-murty:
This PR adds query logging for MySQL. I simply copied from `SQLite3TLPWhereOracle.java`, as modified by PR #648.

### DIFF
--- a/src/sqlancer/mysql/oracle/MySQLTLPWhereOracle.java
+++ b/src/sqlancer/mysql/oracle/MySQLTLPWhereOracle.java
@@ -12,6 +12,8 @@ import sqlancer.mysql.MySQLVisitor;
 
 public class MySQLTLPWhereOracle extends MySQLQueryPartitioningBase {
 
+    private String generatedQueryString;
+
     public MySQLTLPWhereOracle(MySQLGlobalState state) {
         super(state);
     }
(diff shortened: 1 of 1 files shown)


arvind-murty:
This does not actually fix anything. The log files are overwritten when an erroring query occurs (which I assume is intended behavior), and MySQL errors far more frequently than other DBMS's.

arvind-murty:
> > MySQL errors far more frequently than other DBMS's
> 
> During normal execution, no errors should appear unless SQLancer has found a bug. Perhaps it is related to https://github.com/sqlancer/sqlancer#i-am-running-sqlancer-on-the-latest-version-of-a-supported-dbms-is-it-expected-that-sqlancer-prints-many-assertionerrors?

Yes, most likely, since I am running MySQL 8.0.33.

I was also thinking about the logging behavior and I'm not sure if it's intended anymore. Essentially, if I exit the execution (with `--num-tries` set to 1) before an AssertionError is thrown, there is only one log file produced (`database0-cur.log`), which has all of the successfully executed queries.

If instead an AssertionError is thrown, then there are two log files produced (`database0-cur.log` and `database0.log`), both of which only have the unsuccessful query.

I only found this behavior on MySQL because the other DBMS's I tested (DuckDB and SQLite) never errored, which is why I thought MySQL had a logging error.

mrigger:
Yes, that sounds like a bug. Thanks for mentioning it! Do you want to open an issue copying the text? If not, I can create an issue. We should also at some point in time update the MySQL implementation (cc @ZhengLin-Li and @bajinsheng).

ZhengLin-Li:
> > Yes, that sounds like a bug. Thanks for mentioning it! Do you want to open an issue copying the text?
> 
> I have created the issue (#794).

ok, will have a look at it.

Issue #789, opened 2023-05-23
Yugabyte CI ERROR: cannot use tablegroups in a colocated database


ZhengLin-Li:
I have investigaed the Yugabyte CI error recently.

It seams our code violated the rules of Yugabute, resulting the failure of CI. Our code use use tablegroups in a colocated database which is not allow.

Please note the line 305 and line 324 in CI error log: https://github.com/sqlancer/sqlancer/actions/runs/5043922091/jobs/9046326765

Yugabute doc: https://github.com/yugabyte/yugabyte-db/blob/bc6682c0229447cfffc7bf8aa6c9d5deb7802ff2/architecture/design/ysql-tablegroups.md?plain=1#LL99C1-L99C1

I am considering fixing this currently and opened this issue for reference. Any comments on how to fix are welcome!

mrigger:
Yes, I think one possible fix would be to prevent that both features are used at the same time in the table generator of Yugabyte.

bajinsheng:
We can fix the commands to create databases in Yugabyte.

PR #795, opened 2023-05-25
Fix yugabyte tablegroups in colocated database

ZhengLin-Li:
fix: https://github.com/sqlancer/sqlancer/issues/789

In this PR, we fix the issue that SQLancer may create tablegroups in colocated database.

### DIFF
--- a/src/sqlancer/yugabyte/ysql/YSQLSchema.java
+++ b/src/sqlancer/yugabyte/ysql/YSQLSchema.java
@@ -160,7 +160,7 @@ public class YSQLSchema extends AbstractSchema<YSQLGlobalState, YSQLTable> {
         return columns;
     }
 
-    public boolean getDatabaseIsColocated(SQLConnection con) throws SQLException {
+    public boolean getDatabaseIsColocated(SQLConnection con) throws Exception {
         try (Statement s = con.createStatement()) {
             try (ResultSet rs = s.executeQuery("SELECT yb_is_database_colocated();")) {
                 if (!rs.next()) {
--- a/src/sqlancer/yugabyte/ysql/YSQLSchema.java
+++ b/src/sqlancer/yugabyte/ysql/YSQLSchema.java
@@ -166,7 +166,7 @@ public class YSQLSchema extends AbstractSchema<YSQLGlobalState, YSQLTable> {
                 if (!rs.next()) {
                     throw new IgnoreMeException();
                 }
-                String result = rs.getString(0);
+                String result = rs.getString(1);
                 // The query will result in a 'f' for a non-colocated database
                 return !"f".equals(result);
             }
--- a/src/sqlancer/yugabyte/ysql/YSQLSchema.java
+++ b/src/sqlancer/yugabyte/ysql/YSQLSchema.java
@@ -160,16 +160,17 @@ public class YSQLSchema extends AbstractSchema<YSQLGlobalState, YSQLTable> {
         return columns;
     }
 
-    public boolean getDatabaseIsColocated(SQLConnection con) throws Exception {
-        try (Statement s = con.createStatement()) {
-            try (ResultSet rs = s.executeQuery("SELECT yb_is_database_colocated();")) {
-                if (!rs.next()) {
-                    throw new IgnoreMeException();
-                }
-                String result = rs.getString(1);
-                // The query will result in a 'f' for a non-colocated database
-                return !"f".equals(result);
+    public boolean getDatabaseIsColocated(SQLConnection con) {
+        try (Statement s = con.createStatement(); ResultSet rs = s.executeQuery("SELECT yb_is_database_colocated();")) {
+            if (!rs.next()) {
+                throw new IgnoreMeException();
             }
+            String result = rs.getString(1);
+            // The query will result in a 'f' for a non-colocated database
+            return !"f".equals(result);
+
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
         }
     }
 
--- a/src/sqlancer/yugabyte/ysql/YSQLSchema.java
+++ b/src/sqlancer/yugabyte/ysql/YSQLSchema.java
@@ -168,7 +168,7 @@ public class YSQLSchema extends AbstractSchema<YSQLGlobalState, YSQLTable> {
             return !"f".equals(result);
 
         } catch (SQLException e) {
-            throw new RuntimeException(e);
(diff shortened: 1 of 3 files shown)


src/sqlancer/yugabyte/ysql/gen/YSQLCommon.java
```diff
@@ -159,8 +159,14 @@ public static void generateWith(StringBuilder sb, YSQLGlobalState globalState, E
             }
         } else if (Randomly.getBoolean()) {
             errors.add("Cannot use TABLEGROUP with TEMP table");
```
bajinsheng, line 161: Is this error caused by the same reason? Can we remove this error?

src/sqlancer/yugabyte/ysql/gen/YSQLTableGroupGenerator.java
```diff
@@ -15,6 +15,7 @@ private YSQLTableGroupGenerator() {
     }
 
     public static SQLQueryAdapter create(YSQLGlobalState globalState) {
+
```
bajinsheng, line 18: Remove this empty line

src/sqlancer/yugabyte/ysql/gen/YSQLCommon.java
```diff
@@ -159,8 +159,14 @@ public static void generateWith(StringBuilder sb, YSQLGlobalState globalState, E
...
-            sb.append(" TABLEGROUP tg").append(
-                    Randomly.getNotCachedInteger(1, (int) YSQLTableGroupGenerator.UNIQUE_TABLEGROUP_COUNTER.get()));
+            try {
```
bajinsheng, line 162: Why we need this try catch?

src/sqlancer/yugabyte/ysql/YSQLSchema.java
```diff
@@ -160,6 +160,19 @@ protected static List<YSQLColumn> getTableColumns(SQLConnection con, String tabl
         return columns;
     }
 
+    public boolean getDatabaseIsColocated(SQLConnection con) throws Exception {
+        try (Statement s = con.createStatement()) {
+            try (ResultSet rs = s.executeQuery("SELECT yb_is_database_colocated();")) {
+                if (!rs.next()) {
+                    throw new IgnoreMeException();
```
mrigger, line 167: Can it actually be that the database system returns an empty result, or would it be a bug in the database system? If it would be a bug, we should not ignore it.

src/sqlancer/yugabyte/ysql/gen/YSQLCommon.java
mrigger: I think a cleaner way would be to let `getDatabaseIsColocated` return `Optional` (in case an empty result set might be expected for the query that `getDatabaseIsColocated` executes).

src/sqlancer/yugabyte/ysql/YSQLSchema.java
ZhengLin-Li: I think actually it won't returns an empty result. See: https://www.yugabyte.com/blog/differences-primary-key-usage-colocated-noncolocated-tables/#:~:text=a%20colocated%20database%3A-,SELECT%20yb_is_database_colocated()%3B,-Primary%20key%20syntaxes

Also I test it with and without selecting a databse, and it turned out to return a result.

src/sqlancer/yugabyte/ysql/gen/YSQLCommon.java
ZhengLin-Li: I think it is not the same error. 

What we are tring to fix is: `ERROR: cannot use tablegroups in a colocated database`. see: https://github.com/sqlancer/sqlancer/issues/789

But here it is `Cannot use TABLEGROUP with TEMP table`

src/sqlancer/yugabyte/ysql/YSQLSchema.java
mrigger: In this case, I think you can remove the check, and thus also remove the exception-handling code in the caller.

src/sqlancer/yugabyte/ysql/gen/YSQLCommon.java
ZhengLin-Li: What does it mean by saying return `optional`. Could give a example?

src/sqlancer/yugabyte/ysql/YSQLSchema.java
ZhengLin-Li: Ok, make sense.

src/sqlancer/yugabyte/ysql/YSQLSchema.java
```diff
@@ -160,6 +160,20 @@ protected static List<YSQLColumn> getTableColumns(SQLConnection con, String tabl
         return columns;
     }
 
+    public boolean getDatabaseIsColocated(SQLConnection con) {
+        try (Statement s = con.createStatement(); ResultSet rs = s.executeQuery("SELECT yb_is_database_colocated();")) {
+            if (!rs.next()) {
+                throw new IgnoreMeException();
```
mrigger, line 166: I guess we can remove the check here, right?

src/sqlancer/yugabyte/ysql/YSQLSchema.java
```diff
@@ -160,6 +160,18 @@ protected static List<YSQLColumn> getTableColumns(SQLConnection con, String tabl
...
 
+    public boolean getDatabaseIsColocated(SQLConnection con) {
+        try (Statement s = con.createStatement(); ResultSet rs = s.executeQuery("SELECT yb_is_database_colocated();")) {
+            rs.next();
+            String result = rs.getString(1);
+            // The query will result in a 'f' for a non-colocated database
+            return !"f".equals(result);
+
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
```
mrigger, line 171: Typically, we throw an `AssertionError`, but I think this should be fine as well

mrigger:
LGTM, after fixing the PMD warning we can merge the PR.

bajinsheng (approved):
LGTM! But it seems this branch is out of date?

mrigger:
@bajinsheng and I briefly chatted about this - since the change is unlikely to break anything despite being outdated with respect to the main branch, we decided to merge the PR (currently, it seems only I have the necessary permissions).

PR #796, opened 2023-06-02
Fix the package declaration for the Materialize QPG test

mrigger:


def- (approved):
Oops, copy-paste error

PR #797, opened 2023-06-02
Remove the redundancy in comparing floating-point values

mrigger:
The canonicalization was first introduced as part of https://github.com/sqlancer/sqlancer/pull/497 and then copied to subsequent implementations.


kokrui (approved):
for what it's worth lgtm! this definitely makes more sense if this canonicalization is required for multiple DBMSes

mrigger:
Thanks for the prompt review!

Issue #799, opened 2023-06-07
SQLancer fails to load providers when using a newer JVM


mrigger:
When using SQLancer with the JVM below (e.g., using a command such as `java -jar sqlancer-2.0.0.jar duckdb`), everything seems to work fine.

```
openjdk 11.0.18 2023-01-17
OpenJDK Runtime Environment (build 11.0.18+10-post-Ubuntu-0ubuntu122.04)
OpenJDK 64-Bit Server VM (build 11.0.18+10-post-Ubuntu-0ubuntu122.04, mixed mode, sharing)
```

If I now use the JVM below, I get a `ParameterException` as shown farther below.
```
openjdk 17.0.6 2023-01-17
OpenJDK Runtime Environment Temurin-17.0.6+10 (build 17.0.6+10)
OpenJDK 64-Bit Server VM Temurin-17.0.6+10 (build 17.0.6+10, mixed mode)
```
```
Exception in thread "main" com.beust.jcommander.ParameterException: Was passed main parameter 'duckdb' but no main parameter was defined in your arg class
	at com.beust.jcommander.JCommander.initMainParameterValue(JCommander.java:961)
	at com.beust.jcommander.JCommander.parseValues(JCommander.java:762)
	at com.beust.jcommander.JCommander.parse(JCommander.java:363)
	at com.beust.jcommander.JCommander.parse(JCommander.java:342)
	at sqlancer.Main.executeMain(Main.java:467)
	at sqlancer.Main.main(Main.java:296)
```

This seems related to the `ServiceLoader` that we are using. Starting SQLancer with no options (`java -jar sqlancer-2.0.0.jar`) shows that no providers were loaded since only general options are shown:
```
Usage: SQLancer [options]
  Options:
    --constant-cache-size
      Specifies the size of the constant cache. This option only takes effect 
      when constant caching is enabled
      Default: 100
    --database-prefix
      The prefix used for each database created
      Default: database
...
    --use-constant-caching
      Specifies whether constants should be cached and re-used with a certain 
      probability 
      Default: true
    --use-reducer
      EXPERIMENTAL Attempt to reduce queries using a simple reducer
      Default: false
    --username
      The user name used to log into the DBMS
      Default: sqlancer
```

[...]

PR #800, opened 2023-06-07
Add a workaround for issue 799

mrigger:
Work around https://github.com/sqlancer/sqlancer/issues/799.


oraluben (approved):
LGTM as a workaround.

sqlancer doesn't work well when integrated with IntelliJ Idea as well. Might need further work on IDE integration.

PR #813, opened 2023-06-13
fix: rename master to main, add certain constrains

ZhengLin-Li:
fix: fix the ci not running bug

- rename master to main
- add certain constrains

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,9 +4,14 @@ on:
   push:
     branches:
       - master
+      - main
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -4,9 +4,14 @@ on:
   push:
     branches:
       - master
```
mrigger, line 6: I guess we could also drop `master` here and below?

mrigger (approved):

PR #815, opened 2023-06-13
feat: Add virtual db for statement reduction testing

ColinYoungTaro:
**Simplify the code of Statement Reducer**
The Statement Reducer implementation is basically the same as #757 , except for some simplification.

**Implement a fake engine prototype named `VirtualDB`**
`VirtualDB` is implemented at ` test/sqlancer/reducer ` so that The reducer could be tested automatically. A condition could be set to trigger a bug.

**Write 3 simple test cases in test/sqlancer/reducer/TestStatementReducer.java.**
+ 3 SQL statements, the select statement would trigger a bug.
+ 1000 fake statement in the format: `Statement_i;` one of which would trigger a bug.
+ 1000 fake statements and the combination of certain statements would trigger a bug.

**Add max steps and timeout ops**


src/sqlancer/StatementReducer.java
```diff
@@ -1,59 +1,95 @@
...
 
     public StatementReducer(DatabaseProvider<G, O, C> provider) {
         this.provider = provider;
     }
 
+    private boolean hasNotReachedLimit(long curr, long limit) {
+        if (limit > 0) {
+            return curr < limit;
+        }
+        return true;
```
mrigger, line 24: I think this would be easier to read if you first check for the special value `-1` and return `true` then. You could also define a constant in `MainOptions` for that, so that future changes keep in sync.

test/sqlancer/reducer/TestEnvironment.java
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     *
+     * @param queries:
+     *            List of Query<?>
+     *
+     * @return String of queries that appended together with '\n' separated (no '\n' at the last line)
+     */
+    public static String getQueriesString(List<Query<?>> queries) {
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < queries.size(); i++) {
```
mrigger, line 36: For these kinds of things, the streaming API seems quite handy.

test/sqlancer/reducer/VirtualDB/VirtualDBOptions.java
```diff
@@ -0,0 +1,27 @@
...
+public class VirtualDBOptions implements DBMSSpecificOptions<VirtualDBFactory> {
+
+    List<VirtualDBFactory> factories = new ArrayList<>();
+
+    @Override
+    public List<VirtualDBFactory> getTestOracleFactory() {
+        return factories;
+    }
+
+    public enum VirtualDBFactory implements OracleFactory<VirtualDBGlobalState> {
```
mrigger, line 19: Do we actually need this enum? Could we just return an empty list above?

test/sqlancer/reducer/TestEnvironment.java
```diff
@@ -0,0 +1,127 @@
...
+            newGlobalState.setDatabaseName(databaseName);
+            newGlobalState.setMainOptions(options);
+        }
+    }
+
+    public void setInitialStatementsFromStrings(List<String> statements) {
+        List<Query<?>> queries = new ArrayList<>();
+        for (String s : statements) {
+            Query<?> q;
+            if (s.contains("CREATE TABLE") && !s.startsWith("EXPLAIN")) {
```
mrigger, line 99: I assume we could just always use `q = new VirtualDBQuery(s);` without the check, or is it needed somewhere?

mrigger (commented):
Looks good to me overall. I added some minor potential improvement suggestions. Perhaps @oraluben has more.

test/sqlancer/reducer/VirtualDB/VirtualDBOptions.java
ColinYoungTaro: OK, I would change it to a dummy class.

src/sqlancer/StatementReducer.java
```diff
@@ -1,59 +1,95 @@
...
     private final DatabaseProvider<G, O, C> provider;
     private boolean observedChange;
+    private int partitionNum;
 
     public StatementReducer(DatabaseProvider<G, O, C> provider) {
         this.provider = provider;
     }
 
+    private boolean hasNotReachedLimit(long curr, long limit) {
+        if (limit == MainOptions.NO_REDUCE_LIMIT) {
```
ColinYoungTaro, line 21: create a constant NO_REDUCE_LIMIT 
make the code more readable

test/sqlancer/reducer/TestEnvironment.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     *
+     * @param queries:
+     *            List of Query<?>
+     *
+     * @return String of queries that appended together with '\n' separated (no '\n' at the last line)
+     */
+    public static String getQueriesString(List<Query<?>> queries) {
+        return queries.stream().map(Query::getQueryString).collect(Collectors.joining("\n"));
```
ColinYoungTaro, line 36: Use stream API to construct the string.

test/sqlancer/reducer/TestEnvironment.java
```diff
@@ -0,0 +1,115 @@
...
+            newGlobalState.setState(stateToReproduce);
+            newGlobalState.setDatabaseName(databaseName);
+            newGlobalState.setMainOptions(options);
+        }
+    }
+
+    public void setInitialStatementsFromStrings(List<String> statements) {
+        List<Query<?>> queries = new ArrayList<>();
+        for (String s : statements) {
+            queries.add(new VirtualDBQuery(s));
```
ColinYoungTaro, line 92: directly add `new VirtualDBQuery(s)`

test/sqlancer/reducer/VirtualDB/VirtualDBOptions.java
```diff
@@ -0,0 +1,25 @@
...
+
+    List<VirtualDBFactory> factories = new ArrayList<>();
+
+    @Override
+    public List<VirtualDBFactory> getTestOracleFactory() {
+        return factories;
+    }
+
+    public static class VirtualDBFactory implements OracleFactory<VirtualDBGlobalState> {
+        @Override
```
ColinYoungTaro, line 20: replace the `VirtualDBFactory` to a static class so that no enum is needed.

oraluben:
Looks promising! I'll take a closer look later.
Does this PR contain all the changes of the previous delta debugging one? In general, it would be a good idea to keep git history in a PR, rather than merge them into one (and use force push). This would allow other to understand what you've done and what's changed since last review.

ColinYoungTaro:
> Looks promising! I'll take a closer look later. Does this PR contain all the changes of the previous delta debugging one? In general, it would be a good idea to keep git history in a PR, rather than merge them into one (and use force push). This would allow other to understand what you've done and what's changed since the last review. Some projects do require the commits to be squashed before merging, but that's not the case during development.

Yes, all the changes of delta-debugging  are here. Timeout and max steps options are added. And I simplify the code by inlining the lambda function and use `subList.clear` to remove statements install of using for-loop. 
Sorry for the violation of rules, I'll pay attention to it.

src/sqlancer/StatementReducer.java
```diff
@@ -84,4 +123,4 @@ private void printQueries(List<Query<C>> statements) {
         System.out.println("===============================");
     }
 
-}
+}
```
oraluben, line 126: tiny issue: it's better to have an extra empty line at the end of file.

Detail: https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file

oraluben (approved):
LGTM

oraluben:
Now we have reducer tests!🎉 is this ready to merge? @ColinYoungTaro

ColinYoungTaro:
> Now we have reducer tests!🎉 is this ready to merge? @ColinYoungTaro 

I think so. 😀

mrigger:
It seems the style-check test is currently failing. You can check the warnings using mvn -B verify -DskipTests=true.

mrigger:
After fixing this, we can merge the PR.

oraluben:
I've fixed the format issue. Currently the "general" tests are still failing, because it runs the reducer tests, which violate some naming convention.

I suggest to skip the reducer tests in "general" (change `mvn -B '-Dtest=!sqlancer.dbms.**,!sqlancer.qpg.**' test` in `main.yml`), this could also reduce noise in the output.

mrigger:
>I suggest to skip the reducer tests in "general"

Sounds good - we can enable them later.

oraluben:
I think we're ready to go :)

PR #819, opened 2023-06-13
Refactor: improve debugging experience

ZhengLin-Li:
Often, I do not know which sentence belongs to which part when debugging, which is very annoying. 

This PR separates them out by adding an additional `\n` return.

### DIFF
--- a/src/sqlancer/ComparatorHelper.java
+++ b/src/sqlancer/ComparatorHelper.java
@@ -84,13 +84,13 @@ public final class ComparatorHelper {
         return resultSet;
     }
 
-    public static void assumeResultSetsAreEqual(List<String> firstResultSet, List<String> secondResultSet,
-            String originalQuery, List<String> combinedQuery, SQLGlobalState<?, ?> state) {
-        if (firstResultSet.size() != secondResultSet.size()) {
+    public static void assumeResultSetsAreEqual(List<String> resultSet, List<String> secondResultSet,
+            String originalQueryString, List<String> combinedString, SQLGlobalState<?, ?> state) {
+        if (resultSet.size() != secondResultSet.size()) {
             String queryFormatString = "-- %s;" + System.lineSeparator() + "-- cardinality: %d" + System.lineSeparator()
                     + System.lineSeparator() + System.lineSeparator();
-            String firstQueryString = String.format(queryFormatString, originalQuery, firstResultSet.size());
-            String combinedQueryString = String.join(";", combinedQuery);
+            String firstQueryString = String.format(queryFormatString, originalQueryString, resultSet.size());
+            String combinedQueryString = String.join(";", combinedString);
             String secondQueryString = String.format(queryFormatString, combinedQueryString, secondResultSet.size());
             state.getState().getLocalState()
                     .log(String.format("%s" + System.lineSeparator() + "%s", firstQueryString, secondQueryString));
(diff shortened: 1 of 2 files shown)


mrigger:
Could you show a before/after output? I noticed that the logging no longer works correctly, as newlines are currently printed as `\n`, rather than an actual linebreak.

src/sqlancer/ComparatorHelper.java
```diff
@@ -84,35 +84,42 @@ public static List<String> getResultSetFirstColumnAsString(String queryString, E
...
+                    firstResultSet.size(), secondResultSet.size(), firstQueryString, secondQueryString);
             throw new AssertionError(assertionMessage);
         }
 
-        Set<String> firstHashSet = new HashSet<>(resultSet);
+        Set<String> firstHashSet = new HashSet<>(firstResultSet);
         Set<String> secondHashSet = new HashSet<>(secondResultSet);
 
         if (!firstHashSet.equals(secondHashSet)) {
             Set<String> firstResultSetMisses = new HashSet<>(firstHashSet);
```
ZhengLin-Li, line 109: here, what firstResultSetMisses should be (secondHashSet-firstHashSet)?, e.g. 
```java
Set<String> firstResultSetMisses = new HashSet<>(secondHashSet);
firstResultSetMisses.removeAll(firstHashSet);
```

bajinsheng: Sounds good. But it is a name conversion issue, which is not related to the current issue according to your description. Can you create a separate PR to submit all your modifications about name conversions?

bajinsheng:
> Could you show a before/after output? I noticed that the logging no longer works correctly, as newlines are currently printed as `\n`, rather than an actual linebreak.

@ZhengLin-Li  Can you also look into Manuel's comment?

ZhengLin-Li:
> Could you show a before/after output? I noticed that the logging no longer works correctly, as newlines are currently printed as `\n`, rather than an actual linebreak.

before:

image




after:

image

src/sqlancer/ComparatorHelper.java
```diff
@@ -87,13 +87,19 @@ public static List<String> getResultSetFirstColumnAsString(String queryString, E
...
-            state.getState().getLocalState().log(String.format("%s\n%s", firstQueryString, secondQueryString));
-            String assertionMessage = String.format("the size of the result sets mismatch (%d and %d)!\n%s\n%s",
-                    resultSet.size(), secondResultSet.size(), firstQueryString, secondQueryString);
+            String combinedQueryString = String.join(";", combinedString);
+            String secondQueryString = String.format(queryFormatString, combinedQueryString, secondResultSet.size());
+            state.getState().getLocalState()
+                    .log(String.format("%s" + System.lineSeparator() + "%s", firstQueryString, secondQueryString));
+            String assertionMessage = String.format(
+                    "the size of the result sets mismatch (%d and %d)!" + System.lineSeparator()
+                            + "FirstQuery: \"%s\", whose cardinality is: %d" + System.lineSeparator()
```
mrigger, line 99: Since these are not variable names, I think we could just use "first query: " and "second query: "?

ZhengLin-Li: make sense, done.

mrigger (approved):
LGTM! Based on my understanding, you addressed @bajinsheng's comments as well, so I'm merging the PR.

Issue #822, opened 2023-06-15
Supporting the Oracle for Finding Performance Issues


bajinsheng:
We support the new oracle to find performance issues by testing cardinality estimation.
See more: [Arxiv](https://arxiv.org/abs/2306.00355), [Twitter](https://twitter.com/JinshengBa/status/1664445689060032512)

bajinsheng:
Quick Start:
```shell
java -jar sqlancer-2.0.0.jar --num-threads 1 cockroachdb --oracle CERT
```

```shell
java -jar sqlancer-2.0.0.jar --num-threads 1 tidb --oracle CERT
```


```shell
java -jar sqlancer-2.0.0.jar --num-threads 1 mysql --oracle CERT
```

PR #829, opened 2023-06-18
tidb: enable non-empty tables

bajinsheng:
Within https://github.com/sqlancer/sqlancer/issues/822

### DIFF
--- a/src/sqlancer/tidb/TiDBProvider.java
+++ b/src/sqlancer/tidb/TiDBProvider.java
@@ -5,6 +5,8 @@ import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import com.google.auto.service.AutoService;
 
@@ -21,6 +23,7 @@ import sqlancer.common.query.SQLQueryAdapter;
 import sqlancer.common.query.SQLQueryProvider;
 import sqlancer.common.query.SQLancerResultSet;
 import sqlancer.tidb.TiDBProvider.TiDBGlobalState;
+import sqlancer.tidb.TiDBSchema.TiDBTable;
 import sqlancer.tidb.gen.TiDBAlterTableGenerator;
 import sqlancer.tidb.gen.TiDBAnalyzeTableGenerator;
 import sqlancer.tidb.gen.TiDBDeleteGenerator;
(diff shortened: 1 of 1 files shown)


src/sqlancer/tidb/TiDBProvider.java
```diff
@@ -84,6 +87,7 @@ private static int mapActions(TiDBGlobalState globalState, Action a) {
         case CREATE_INDEX:
             return r.getInteger(0, 2);
         case INSERT:
+            return r.getInteger(0, globalState.getOptions().getMaxNumberInserts());
```
mrigger, line 90: Interesting, I didn't notice that we previously generated only up to two `INSERT` statements.

src/sqlancer/tidb/TiDBProvider.java
```diff
@@ -213,4 +217,15 @@ protected void executeMutator(int index, TiDBGlobalState globalState) throws Exc
         globalState.executeStatement(queryMutateTable);
     }
 
+    @Override
+    public boolean addRowsToAllTables(TiDBGlobalState globalState) throws Exception {
+        List<TiDBTable> tablesNoRow = globalState.getSchema().getDatabaseTables().stream()
+                .filter(t -> t.getNrRows(globalState) == 0).collect(Collectors.toList());
+        for (TiDBTable table : tablesNoRow) {
+            SQLQueryAdapter queryAddRows = TiDBInsertGenerator.getQuery(globalState, table);
```
mrigger, line 225: Based on my understanding, executing this might result in an `ExpectedError`, in which case no row is successfully inserted. Is this an issue? If so, perhaps we can check the number of rows in the table to decide whether to continue trying? In that case, we also need to pay attention that it's not possible to insert a row (e.g., in case of a `CHECK` constraint that always evaluates to `FALSE`). If it's not an issue, we can go ahead and merge.

bajinsheng: I think it should be fine as it is hard to make sure tables are not non-empty.
 
We have the same logic for other DBMSs, such as https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/sqlite3/SQLite3Provider.java#L350

Perhaps we can fix them in the future if needed.

mrigger: Could you open an issue for that so we can track this?

PR #831, opened 2023-06-18
tidb: cert init

bajinsheng:
This PR was tested locally, but I do not include it in CI because:
* CERT oracle generates more than 50 issues per minute. It makes CI always fail.
* The checking of query plan formats has been covered by the CI for QPG.

Within: https://github.com/sqlancer/sqlancer/issues/822

### DIFF
--- a/src/sqlancer/common/DBMSCommon.java
+++ b/src/sqlancer/common/DBMSCommon.java
@@ -1,5 +1,6 @@
 package sqlancer.common;
 
+import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -39,4 +40,26 @@ public final class DBMSCommon {
         }
         return maxIndex;
     }
+
+    public static int editDistance(List<String> list1, List<String> list2) {
+        int[][] dp = new int[list1.size() + 1][list2.size() + 1];
+        for (int i = 0; i <= list1.size(); i++) {
+            for (int j = 0; j <= list2.size(); j++) {
+                if (i == 0) {
+                    dp[i][j] = j;
+                } else if (j == 0) {
+                    dp[i][j] = i;
+                } else {
+                    dp[i][j] = Math.min(dp[i - 1][j - 1] + costOfSubstitution(list1.get(i - 1), list2.get(j - 1)),
+                            Math.min(dp[i - 1][j] + 1, dp[i][j - 1] + 1));
+                }
+            }
+        }
+        return dp[list1.size()][list2.size()];
+    }
+
+    private static int costOfSubstitution(String string, String string2) {
+        return string.equals(string2) ? 0 : 1;
+    }
+
 }
--- a/src/sqlancer/common/DBMSCommon.java
+++ b/src/sqlancer/common/DBMSCommon.java
@@ -41,7 +41,11 @@ public final class DBMSCommon {
         return maxIndex;
     }
 
-    public static int areQueryPlanSequencesSimilar(List<String> list1, List<String> list2) {
+    public static boolean areQueryPlanSequencesSimilar(List<String> list1, List<String> list2) {
+        return editDistance(list1, list2) <= 1;
+    }
+
+    public static int editDistance(List<String> list1, List<String> list2) {
         int[][] dp = new int[list1.size() + 1][list2.size() + 1];
         for (int i = 0; i <= list1.size(); i++) {
             for (int j = 0; j <= list2.size(); j++) {
--- a/src/sqlancer/common/DBMSCommon.java
+++ b/src/sqlancer/common/DBMSCommon.java
@@ -41,7 +41,7 @@ public final class DBMSCommon {
         return maxIndex;
     }
 
-    public static int editDistance(List<String> list1, List<String> list2) {
+    public static int areQueryPlanSequencesSimilar(List<String> list1, List<String> list2) {
         int[][] dp = new int[list1.size() + 1][list2.size() + 1];
         for (int i = 0; i <= list1.size(); i++) {
             for (int j = 0; j <= list2.size(); j++) {
--- a/src/sqlancer/common/ast/SelectBase.java
+++ b/src/sqlancer/common/ast/SelectBase.java
@@ -90,7 +90,7 @@ public class SelectBase<T> {
     public void clearHavingClause() {
         this.havingClause = null;
     }
-    
+
     public void setLimitClause(T limitClause) {
         this.limitClause = limitClause;
     }
--- a/src/sqlancer/common/ast/SelectBase.java
+++ b/src/sqlancer/common/ast/SelectBase.java
@@ -50,6 +50,10 @@ public class SelectBase<T> {
         this.groupByExpressions = groupByExpressions;
     }
 
+    public void clearGroupByExpressions() {
+        this.groupByExpressions = Collections.emptyList();
+    }
+
     public List<T> getGroupByExpressions() {
         assert groupByExpressions != null;
         return groupByExpressions;
@@ -83,6 +87,10 @@ public class SelectBase<T> {
         return havingClause;
     }
 
+    public void clearHavingClause() {
+        this.havingClause = null;
+    }
+    
     public void setLimitClause(T limitClause) {
         this.limitClause = limitClause;
     }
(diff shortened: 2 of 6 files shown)


src/sqlancer/common/ast/SelectBase.java
```diff
@@ -50,6 +50,10 @@ public void setGroupByExpressions(List<T> groupByExpressions) {
         this.groupByExpressions = groupByExpressions;
     }
 
+    public void clearGroupByExpressions() {
```
mrigger, line 53: I think we don't need a separate method here and below. We can just use `null` as an argument to clear the expression(s). For the `GROUP BY` expressions, we can check whether the argument is `null` and if so, set an empty list.

src/sqlancer/tidb/TiDBProvider.java
```diff
@@ -136,6 +138,22 @@ public void generateDatabase(TiDBGlobalState globalState) throws Exception {
                 throw new AssertionError(e);
             }
         }
+
+        if (globalState.getDbmsSpecificOptions().getTestOracleFactory().get(0)
```
mrigger, line 142: This only checks for one test oracle, while multiple ones can be passed, I think. Perhaps we can check all of them?

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
```diff
@@ -0,0 +1,285 @@
...
+
+        // Get the result of the first query
+        String queryString1 = TiDBVisitor.asString(select);
+        double rowCount1 = getRow(state, queryString1, queryPlan1Sequences);
+
+        // Mutate the query
+        boolean increase = false;
+        int mutator = (int) Randomly.getNotCachedInteger(0, 7);
+
+        switch (mutator) {
```
mrigger, line 94: We should use an enum for the different options here, which is less error prone.

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
```diff
@@ -0,0 +1,285 @@
...
+        default:
+            throw new AssertionError();
+        }
+
+        // Get the result of the second query
+        String queryString2 = TiDBVisitor.asString(select);
+        double rowCount2 = getRow(state, queryString2, queryPlan2Sequences);
+
+        // Check structural equivalence
+        if (DBMSCommon.editDistance(queryPlan1Sequences, queryPlan2Sequences) > 1) {
```
mrigger, line 125: I guess we could just have a method `DBMSCommon.areQueryPlanSequencesSimilar` or so, which might be less error-prone to use? In that case, we wouldn't need to expose that we are using the edit distance here.

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
```diff
@@ -0,0 +1,285 @@
...
+            if (Randomly.getBoolean()) {
+                select.setHavingClause(gen.generateHavingClause());
+            }
+        }
+        if (Randomly.getBoolean()) {
+            select.setLimitClause(gen.generateExpression());
+        }
+        if (Randomly.getBoolean()) {
+            select.setOffsetClause(gen.generateExpression());
+        }
```
mrigger, line 84: Could we extract this to a method that returns a random `SELECT` statement? Perhaps this would be something that the other test oracles might be able to use?

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
```diff
@@ -0,0 +1,285 @@
...
+        // Get the row count
+        SQLQueryAdapter q = new SQLQueryAdapter(explainQuery, errors);
+        try (SQLancerResultSet rs = q.executeAndGet(globalState)) {
+            if (rs != null) {
+                while (rs.next()) {
+                    double estRows = Double.parseDouble(rs.getString(2));
+                    if (row == -1) {
+                        row = estRows;
+                    }
+                    String operation = rs.getString(1).split("_")[0];
```
mrigger, line 272: perhaps a comment would make clearer what we are extracting here?

src/sqlancer/common/ast/SelectBase.java
bajinsheng: Directly setting `null` will leave some issues, such as `null pointer`. Here, I need to make it an empty `collection`, instead of `null.

src/sqlancer/tidb/TiDBProvider.java
bajinsheng: CERT is impossible to be passed with other oracles. In TiDB, only `WHERE` and `HAVING` oracles can be composed.
If we enforce to compose CERT with other oracles, it is incorrect to apply these settings to other oracles.

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
bajinsheng: Solved

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
bajinsheng: I think edit distance is more concrete and can be used in other code?

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
bajinsheng: Actually I have some customized code to generate satisfied `SELECT` statements, such as no natural join

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
bajinsheng: Solved

src/sqlancer/common/ast/SelectBase.java
mrigger: Right. Above, I meant, that in the setter, we can just check whether the argument is `null`, and if so, set an empty list.

src/sqlancer/tidb/TiDBProvider.java
mrigger: I think we can then validate that only one test oracle has been specified before executing this code?

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
mrigger: Given that you include a comment to explain what this does call does, I think it's clearer to just have a method with a similar name.

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
mrigger: I see. Then, perhaps, we can just extract a method but leave the other parts as is?

src/sqlancer/common/ast/SelectBase.java
bajinsheng: https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/common/ast/SelectBase.java#L48 If the argument is `null`, it would throw an assertion. I guess it may break other code that depends on this function.

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
bajinsheng: We can remove the comment.

bajinsheng: We can remove the comment.
However, I think changing the function name makes it less general and clear.
Anyway, updated.

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
bajinsheng: It will absolutely make the code more complicated, as we do not have clear application scenarios that how other oracles use it while we create a complicated architecture. Therefore, I prefer to leave it as it is to make the code simple.

src/sqlancer/common/DBMSCommon.java
```diff
@@ -39,4 +40,26 @@ public static int getMaxIndexInDoubleArray(double... doubleArray) {
         }
         return maxIndex;
     }
+
+    public static int areQueryPlanSequencesSimilar(List<String> list1, List<String> list2) {
```
mrigger, line 44: What I meant is that we have a separate `areQueryPlanSequencesSimilar` method that returns a boolean value, and the previous `private` method that is called by it.

src/sqlancer/common/ast/SelectBase.java
mrigger: Alright, then let's use the new `clear` methods that you introduced rather than changing the behavior of the `SelectBase`. I think it would unlikely, though, that existing code relies on an exception being thrown here.

src/sqlancer/common/DBMSCommon.java
bajinsheng: Sorry for the misunderstanding. Fixed now.

mrigger (approved):

PR #835, opened 2023-06-21
mysql: optimize code for CERT

bajinsheng:
Within: https://github.com/sqlancer/sqlancer/issues/822

### DIFF
--- a/src/sqlancer/mysql/ast/MySQLSelect.java
+++ b/src/sqlancer/mysql/ast/MySQLSelect.java
@@ -18,6 +18,10 @@ public class MySQLSelect extends SelectBase<MySQLExpression> implements MySQLExp
         this.setFromOptions(fromOptions);
     }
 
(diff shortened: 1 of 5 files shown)


src/sqlancer/mysql/ast/MySQLSelect.java
```diff
@@ -18,6 +18,10 @@ public void setSelectType(SelectType fromOptions) {
         this.setFromOptions(fromOptions);
     }
 
+    public SelectType getSelectType() {
```
mrigger, line 21: Isn't this duplicate to the method below?

bajinsheng: Aha yes. Will fix later.

bajinsheng:
Seems a bug in jdbc driver?

mrigger:
Yes, seems so, but likely not caused by the change.

PR #816, opened 2023-06-13
feat: initial support of StoneDB

ZhengLin-Li:
feat: initial support of StoneDB

PR #836, opened 2023-06-23
feat: stonedb follow up implementation

ZhengLin-Li:
This PR is the follow up implementation of StoneDB.

Change list:

- add serveral generator, including alterTableGenerator, DeleteGenerator, ExpressionGenerator, IndexGenerator, InsertGenerator
- add ToStringVisitor
- partly implement new ast


src/sqlancer/stonedb/StoneDBSchema.java
```diff
@@ -0,0 +1,382 @@
...
+                        isNullable, precision);
+                columns.add(c);
+            }
+        }
+        return columns;
+    }
+
+    private static StoneDBCompositeDataType getColumnCompositeDataType(String typeString) {
+        switch (typeString) {
+        case "tinyint":
```
mrigger, line 112: For this, I guess `StoneDBDataType.valueOf` would result in a one-line implementation?

src/sqlancer/stonedb/StoneDBSchema.java
```diff
@@ -0,0 +1,382 @@
...
+                size = -2;
+                break;
+            case MEDIUMBLOB:
+                size = -2;
+                break;
+            case LONGBLOB:
+                size = -2;
+                break;
+            default:
+                size = -1;
```
mrigger, line 360: I guess this assignment doesn't make sense, since you throw an exception right after?

src/sqlancer/stonedb/ast/StoneDBConstant.java
```diff
@@ -0,0 +1,192 @@
...
+import java.text.SimpleDateFormat;
+
+import sqlancer.common.ast.newast.Node;
+
+public class StoneDBConstant implements Node<StoneDBExpression> {
+
+    private StoneDBConstant() {
+    }
+
+    public static class DuckDBNullConstant extends StoneDBConstant {
```
mrigger, line 13: Check for other naming issues where you still refer to `DuckDB`

src/sqlancer/stonedb/StoneDBSchema.java
ZhengLin-Li: oh yes, make sense, I just want to let `-1` represents the abnormal value here.

src/sqlancer/stonedb/StoneDBSchema.java
ZhengLin-Li: Ohhh, yes, will change.

src/sqlancer/Randomly.java
```diff
@@ -452,6 +451,18 @@ public long getLong(long left, long right) {
         return getNextLong(left, right);
     }
 
+    public BigInteger getBigInteger(BigInteger left, BigInteger right) {
+        if (left.equals(right)) {
+            return left;
+        }
+        while (true) {
+            BigInteger result = new BigInteger(63, new Random());
+            if (result.compareTo(left) >= 0 && result.compareTo(right) <= 0) {
```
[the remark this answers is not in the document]
ZhengLin-Li: Yes, this is a problem.

I searched but did not found perfect implementation. So how about changing to this?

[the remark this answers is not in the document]
mrigger: This still seems rather complex. Perhaps we can just get a random long value for now? I think it's better to be sure that SQLancer doesn't get stuck, than being able to generate large random big integers within a certain range.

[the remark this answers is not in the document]
ZhengLin-Li: sounds good, changed to a simplier way.

src/sqlancer/Randomly.java
```diff
@@ -452,6 +451,13 @@ public long getLong(long left, long right) {
         return getNextLong(left, right);
     }
 
+    public BigInteger getBigInteger(BigInteger left, BigInteger right) {
+        if (left.equals(right)) {
+            return left;
+        }
+        return new BigInteger(String.valueOf(getInteger(left.intValue(),right.intValue())));
```
mrigger, line 458: I think this might cause some issues. According to https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html#intValue, this method truncates the integer, so it might return a number outside the range specified by the big integers. Perhaps we can first check whether the big integer can be precisely represented, and if not, throw an `IgnoreMeException`?

ZhengLin-Li: OK, IgnoreMeException  added

src/sqlancer/stonedb/StoneDBSchema.java
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    private static StoneDBCompositeDataType getColumnCompositeDataType(String typeString) {
+        if (Arrays.stream(StoneDBDataType.values()).noneMatch(e -> e.name().equals(typeString.toUpperCase()))) {
+            throw new AssertionError(typeString);
+        }
+        return new StoneDBCompositeDataType(StoneDBDataType.valueOf(typeString.toUpperCase()));
+    }
+
+    //
```
mrigger, line 179: I guess we could remove the commented-out code?

src/sqlancer/stonedb/ast/StoneDBConstant.java
```diff
@@ -0,0 +1,203 @@
...
+        private final Long value;
+
+        public StoneDBBigIntConstant(long value) {
+            this.value = value;
+        }
+
+        @Override
+        public String toString() {
+            if (value.equals(Long.MIN_VALUE)) {
+                return "BIGINT_NULL";
```
mrigger, line 54: I think that here, a brief comment would be useful to explain what this does. Does StoneDB not support Long.MIN_VALUE?

src/sqlancer/stonedb/gen/StoneDBExpressionGenerator.java
```diff
@@ -0,0 +1,276 @@
...
+    @Override
+    public Node<StoneDBExpression> generateConstant() {
+        if (Randomly.getBooleanWithSmallProbability()) {
+            return StoneDBConstant.createNullConstant();
+        }
+        StoneDBDataType type = StoneDBDataType.getRandomWithoutNull();
+        switch (type) {
+        case INT:
+            return StoneDBConstant
+                    .createIntConstant(globalState.getRandomly().getInteger(Integer.MIN_VALUE + 1, Integer.MAX_VALUE));
```
mrigger, line 57: I assume it's deliberate that we are not generating `Integer.MIN_VALUE` here?

src/sqlancer/stonedb/gen/StoneDBInsertGenerator.java
```diff
@@ -0,0 +1,88 @@
...
+import sqlancer.Randomly;
+import sqlancer.common.query.ExpectedErrors;
+import sqlancer.common.query.SQLQueryAdapter;
+import sqlancer.common.schema.AbstractTableColumn;
+import sqlancer.stonedb.StoneDBProvider.StoneDBGlobalState;
+import sqlancer.stonedb.StoneDBSchema.StoneDBColumn;
+import sqlancer.stonedb.StoneDBSchema.StoneDBTable;
+import sqlancer.stonedb.StoneDBToStringVisitor;
+
+public class StoneDBInsertGenerator {
```
mrigger, line 15: Would it be applicable to inhert from https://github.com/sqlancer/sqlancer/blob/1f80de16e38c4a144e8dabbcef00d07cd15470f9/src/sqlancer/common/gen/AbstractInsertGenerator.java here?

mrigger (commented):
Some minor comments, addressing which would be nice. Not sure if @bajinsheng has any additional comments (I assume you already addressed some in the other PR)?

src/sqlancer/stonedb/StoneDBSchema.java
ZhengLin-Li: will refactor this.

src/sqlancer/stonedb/ast/StoneDBConstant.java
ZhengLin-Li: Yes, accroding to https://stonedb.io/docs/SQL-reference/data-types

> For INT type: -2147483648 is reserved to indicate INT_NULL in Tianmu engine, Minimum Value Signed start from -2147483647.

ZhengLin-Li: will add this to the comment in the code

src/sqlancer/stonedb/gen/StoneDBExpressionGenerator.java
ZhengLin-Li: I original thought the "For INT type: -2147483648 is reserved to indicate INT_NULL in Tianmu engine, Minimum Value Signed start from -2147483647." So we need to start from `Integer.MIN_VALUE + 1`

I thought originally that we should generate valid non value. But it seams that we call generate null value also?

src/sqlancer/stonedb/gen/StoneDBInsertGenerator.java
ZhengLin-Li: will try

bajinsheng:
I have no more comment. This code is too huge for me to follow. Perhaps you can create a small PR next time.
Another question, what is the relation of this PR to #816 ?

mrigger:
OK, then merging now. I assume it supersedes the previous PR?

PR #837, opened 2023-06-23
mysql: cert init

bajinsheng:


src/sqlancer/mysql/oracle/MySQLCERTOracle.java
```diff
@@ -0,0 +1,254 @@
...
+
+public class MySQLCERTOracle implements TestOracle<MySQLGlobalState> {
+    private final MySQLGlobalState state;
+    private MySQLExpressionGenerator gen;
+    private final ExpectedErrors errors = new ExpectedErrors();
+    private MySQLSelect select;
+    private List<String> queryPlan1Sequences;
+    private List<String> queryPlan2Sequences;
+
+    public enum Mutator {
```
mrigger, line 37: Could the enum be shared among different implementations? If so, perhaps, this could be called `CertMutation`.

src/sqlancer/mysql/MySQLProvider.java
```diff
@@ -153,6 +156,31 @@ public void generateDatabase(MySQLGlobalState globalState) throws Exception {
...
+            // Enfore statistic collected for all tables
+            ExpectedErrors errors = new ExpectedErrors();
+            MySQLErrors.addExpressionErrors(errors);
+            for (MySQLTable table : globalState.getSchema().getDatabaseTables()) {
+                StringBuilder sb = new StringBuilder();
+                sb.append("ANALYZE TABLE ");
+                sb.append(table.getName());
+                sb.append(" UPDATE HISTOGRAM ON ");
+                boolean first = true;
+                for (MySQLColumn c : table.getColumns()) {
```
mrigger, line 172: I think this loop could be a one-liner with the streaming API? See https://stackoverflow.com/questions/37987051/how-to-concatenate-a-string-with-the-new-1-8-stream-api.

src/sqlancer/mysql/oracle/MySQLCERTOracle.java
```diff
@@ -0,0 +1,248 @@
...
+        String queryString1 = MySQLVisitor.asString(select);
+        int rowCount1 = getRow(state, queryString1, queryPlan1Sequences);
+
+        boolean increase = false;
+        // Mutate the query
+        // Disable limit due to its false positive
+        Mutator mutation = Mutator.getRandom();
+        switch (mutation) {
+        case JOIN:
+            // increase = mutateJoin();
```
mrigger, line 81: not sure if I understand the code here. In the case of a `JOIN`, you fall through to the `DISTINCT` because `JOIN` is not supported in the current implementation? I think it would be more readable to have a `getRandomExcept` method in the `Mutator` to select all enum values except those that are not supported.

I guess you could also just extract the switch statement to a new method in `CertOracleBase` then?

src/sqlancer/common/oracle/CERTOracleBase.java
```diff
@@ -0,0 +1,88 @@
...
+
+    protected final S state;
+    protected final ExpectedErrors errors;
+    protected List<String> queryPlan1Sequences;
+    protected List<String> queryPlan2Sequences;
+
+    protected enum Mutator {
+        JOIN, DISTINCT, WHERE, GROUPBY, HAVING, AND, OR, LIMIT;
+
+        public static Mutator getRandom(Mutator... exclude) {
```
mrigger, line 20: The method name is confusing. Could we name this `getRandomExcept`, `getRandomExcluding`, or something similar? Currently, I'd believe that we randomly select from one of the options.

mrigger (commented):
Could we still find a better name for the `getRandom` method? After that, I'll merge the PR.

mrigger (approved):

PR #839, opened 2023-06-26
CockroachDB: optimize code for CERT

bajinsheng:
Within #822 .

### DIFF
--- a/src/sqlancer/Randomly.java
+++ b/src/sqlancer/Randomly.java
@@ -135,6 +135,14 @@ public final class Randomly {
         return extractNrRandomColumns(columns, nr);
     }
 
+    public static <T> List<T> nonEmptySubsetLeast(List<T> columns, int min) {
+        int nr = getNextInt(min, columns.size() + 1);
+        if (nr > columns.size()) {
+            throw new AssertionError(columns + " " + nr);
+        }
+        return extractNrRandomColumns(columns, nr);
+    }
+
     public static <T> List<T> nonEmptySubsetPotentialDuplicates(List<T> columns) {
         List<T> arr = new ArrayList<>();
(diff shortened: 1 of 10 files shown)


src/sqlancer/Randomly.java
```diff
@@ -135,6 +135,14 @@ public static <T> List<T> nonEmptySubset(List<T> columns, int nr) {
         return extractNrRandomColumns(columns, nr);
     }
 
+    public static <T> List<T> nonEmptySubsetLeast(List<T> columns, int min) {
+        int nr = getNextInt(min, columns.size() + 1);
+        if (nr > columns.size()) {
+            throw new AssertionError(columns + " " + nr);
```
mrigger, line 141: I think we should throw an `AssertionError` only to protect against unexpected arguments. Here, perhaps we can use an `assert` statement? We could still protect against an invalid `min` argument at the beginning of the method and throw an `AssertionError` there.

mrigger (commented):
Looks good to me. Perhaps you can still address the minor comment?

src/sqlancer/Randomly.java
bajinsheng: Not sure I understand your comment. I followed your way to implement `nonEmptySubset`, which also writes AssertionError in this way, to implement this new function.

Why do we use `assert`? How can we throw `AssertionError` still here?

mrigger: In `nonEmptySubset`, we check whether the input argument passed is valid. Protecting APIs with an explicit check that remains also in production mode seems common. In the case above, it seems you are checking whether a random integer is in the expected range? In this case, I assume we'd only throw the error in case `getNextInt` is incorrectly implemented. In this case, it seems appropriate to only check this in debugging/development mode.

mrigger (approved):

PR #843, opened 2023-06-26
doc: add CERT

bajinsheng:
Within #822 .

### DIFF
--- a/src/sqlancer/cockroachdb/CockroachDBProvider.java
+++ b/src/sqlancer/cockroachdb/CockroachDBProvider.java
@@ -19,6 +19,7 @@ import sqlancer.Randomly;
 import sqlancer.SQLConnection;
 import sqlancer.SQLGlobalState;
 import sqlancer.SQLProviderAdapter;
+import sqlancer.cockroachdb.CockroachDBOptions.CockroachDBOracleFactory;
 import sqlancer.cockroachdb.CockroachDBProvider.CockroachDBGlobalState;
 import sqlancer.cockroachdb.CockroachDBSchema.CockroachDBTable;
 import sqlancer.cockroachdb.gen.CockroachDBCommentOnGenerator;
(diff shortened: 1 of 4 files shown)


SuriZhang (approved):
LGTM

src/sqlancer/cockroachdb/CockroachDBProvider.java
```diff
@@ -252,8 +252,9 @@ public void generateDatabase(CockroachDBGlobalState globalState) throws Exceptio
             total--;
         }
 
-        if (globalState.getDbmsSpecificOptions().getTestOracleFactory().size() == 1 && globalState
-                .getDbmsSpecificOptions().getTestOracleFactory().get(0).create(globalState) instanceof CockroachDBCERTOracle) {
+        if (globalState.getDbmsSpecificOptions().getTestOracleFactory().size() == 1
```
mrigger, line 255: What will this do if `CERT` and another oracle like `TLP` is used?

bajinsheng: Actually, I still think checking the number equals to 1 is redundant.

bajinsheng:
```shell
java -jar sqlancer-2.0.0.jar --num-threads 1 --username root --password "" cockroachdb --oracle NoREC,CERT

Exception in thread "main" com.beust.jcommander.ParameterException: Invalid value for --oracle parameter. Allowed values:[NOREC, AGGREGATE, GROUP_BY, HAVING, WHERE, DISTINCT, EXTENDED_WHERE, QUERY_PARTITIONING, CERT]
```

bajinsheng: The really confusing code is here:
[getTestOracleFactory](https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/cockroachdb/CockroachDBOptions.java#L122) returns a list of oracles, but the [oracle](https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/cockroachdb/CockroachDBOptions.java#L33) can only be one.

Not sure why it must return a list of oracles, so I updated the code as you requested.

PR #846, opened 2023-06-29
fix: SQL syntax errors and add expected errors in StoneDB implementation Part 1

ZhengLin-Li:
I'm starting to test our implementation, as well as StoneDB version 5.7, in my local development environment. This PR will fix errors in the StoneDB implementation.

Change list:
- add `StoneDBProvider()` in `Main.java`
- add one space between `CREATE TABLE` and `tableName`
- add one space between `INDEX` and `tableName`
- add one space between `RENAME COLUMN` and `columnName`
- add one space before and after `IF NOT EXISTS`
- support both `getRandomWithoutNull()` and `getRandom()` in `StoneDBDataType`
- implement `StoneDBToStringVisitor` 
- refactor anoyname function to lambda expression.
- add expected errors in all generators.
- change local new `randomly` to global one
- fix bugs in index generator according to the MySQL 5.7 document.
- support `insertValue` to let `StoneDBInsertGenerator` extend `AbstractInsertGenerator<StoneDBColumn>`
- fix bugs when we need `StoneDBPrimitiveDataType` rather than `StoneDBCompositeDataType`
- remove pre-added expected errors
- remove redentent identifier.
- define length after column type when creating tables.
- renaming `StoneDBDeleteGenerator ` to `StoneDBTableDeleteGenerator`
- remove `AS` caulse when deleting tables.
- fix bugs when new `StoneDBCastOperation`

### DIFF
--- a/src/sqlancer/stonedb/StoneDBProvider.java
+++ b/src/sqlancer/stonedb/StoneDBProvider.java
@@ -59,10 +59,18 @@ public class StoneDBProvider extends SQLProviderAdapter<StoneDBProvider.StoneDBG
     }
 
     private static int mapActions(StoneDBGlobalState globalState, Action a) {
-        globalState.getRandomly();
+        Randomly r = globalState.getRandomly();
         switch (a) {
         case SHOW_TABLES:
             return 1;
+        case ALTER_TABLE:
+            return Randomly.smallNumber();
+        case DELETE:
+            return Randomly.smallNumber();
+        case INDEX:
+            return Randomly.smallNumber();
+        case INSERT:
+            return r.getInteger(0, globalState.getOptions().getMaxNumberInserts());
         default:
             throw new AssertionError(a);
         }
--- a/src/sqlancer/stonedb/StoneDBProvider.java
+++ b/src/sqlancer/stonedb/StoneDBProvider.java
@@ -23,6 +23,7 @@ import sqlancer.stonedb.gen.StoneDBIndexGenerator;
 import sqlancer.stonedb.gen.StoneDBInsertGenerator;
 import sqlancer.stonedb.gen.StoneDBTableAlterGenerator;
 import sqlancer.stonedb.gen.StoneDBTableCreateGenerator;
+import sqlancer.stonedb.gen.StoneDBTableDeleteGenerator;
 
 @AutoService(DatabaseProvider.class)
 public class StoneDBProvider extends SQLProviderAdapter<StoneDBProvider.StoneDBGlobalState, StoneDBOptions> {
@@ -41,7 +42,7 @@ public class StoneDBProvider extends SQLProviderAdapter<StoneDBProvider.StoneDBG
     enum Action implements AbstractAction<StoneDBGlobalState> {
         SHOW_TABLES((g) -> new SQLQueryAdapter("SHOW TABLES")), //
         ALTER_TABLE(StoneDBTableAlterGenerator::generate), //
-        DELETE(sqlancer.stonedb.gen.StoneDBTableDeleteGenerator::generate), //
+        DELETE(StoneDBTableDeleteGenerator::generate), //
         INDEX(StoneDBIndexGenerator::generate), //
         INSERT(StoneDBInsertGenerator::generate); //
 
--- a/src/sqlancer/stonedb/StoneDBProvider.java
+++ b/src/sqlancer/stonedb/StoneDBProvider.java
@@ -19,7 +19,6 @@ import sqlancer.common.DBMSCommon;
 import sqlancer.common.query.Query;
 import sqlancer.common.query.SQLQueryAdapter;
 import sqlancer.common.query.SQLQueryProvider;
-import sqlancer.stonedb.gen.StoneDBDeleteGenerator;
 import sqlancer.stonedb.gen.StoneDBIndexGenerator;
 import sqlancer.stonedb.gen.StoneDBInsertGenerator;
 import sqlancer.stonedb.gen.StoneDBTableAlterGenerator;
@@ -42,7 +41,7 @@ public class StoneDBProvider extends SQLProviderAdapter<StoneDBProvider.StoneDBG
     enum Action implements AbstractAction<StoneDBGlobalState> {
         SHOW_TABLES((g) -> new SQLQueryAdapter("SHOW TABLES")), //
         ALTER_TABLE(StoneDBTableAlterGenerator::generate), //
-        DELETE(StoneDBDeleteGenerator::generate), //
+        DELETE(sqlancer.stonedb.gen.StoneDBTableDeleteGenerator::generate), //
         INDEX(StoneDBIndexGenerator::generate), //
         INSERT(StoneDBInsertGenerator::generate); //
 
--- a/src/sqlancer/stonedb/StoneDBProvider.java
+++ b/src/sqlancer/stonedb/StoneDBProvider.java
@@ -19,11 +19,13 @@ import sqlancer.common.DBMSCommon;
 import sqlancer.common.query.Query;
 import sqlancer.common.query.SQLQueryAdapter;
 import sqlancer.common.query.SQLQueryProvider;
-import sqlancer.stonedb.gen.StoneDBIndexGenerator;
-import sqlancer.stonedb.gen.StoneDBInsertGenerator;
+import sqlancer.stonedb.gen.StoneDBIndexCreateGenerator;
+import sqlancer.stonedb.gen.StoneDBIndexDropGenerator;
 import sqlancer.stonedb.gen.StoneDBTableAlterGenerator;
 import sqlancer.stonedb.gen.StoneDBTableCreateGenerator;
 import sqlancer.stonedb.gen.StoneDBTableDeleteGenerator;
+import sqlancer.stonedb.gen.StoneDBTableInsertGenerator;
+import sqlancer.stonedb.gen.StoneDBTableUpdateGenerator;
 
 @AutoService(DatabaseProvider.class)
 public class StoneDBProvider extends SQLProviderAdapter<StoneDBProvider.StoneDBGlobalState, StoneDBOptions> {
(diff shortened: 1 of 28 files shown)


src/sqlancer/stonedb/gen/StoneDBTableCreateGenerator.java
```diff
@@ -153,6 +153,8 @@ private void addCommonErrors(ExpectedErrors list) {
         list.add("A BLOB field is not allowed in partition function");
```
mrigger, line 153: Did you check whether the other expected errors are actually needed?

ZhengLin-Li: Will gradually add each expected errors by each commit

mrigger: My comment was about the existing errors that you have added without feedback by the DBMS. Are they actually needed? I'd suggest to remove them and only re-add those that are needed.

ZhengLin-Li: Oh, got it. Will remove the errors added before today's test and only add them back when I got one.

src/sqlancer/stonedb/gen/StoneDBTableCreateGenerator.java
```diff
@@ -222,8 +224,7 @@ private void appendColumnOption(StoneDBDataType type) {
                 }
                 break;
             case COMMENT:
-                // TODO: generate randomly
-                sb.append(String.format("COMMENT '%s' ", "asdf"));
+                sb.append(String.format("COMMENT '%s' ", new Randomly().getString()));
```
mrigger, line 227: Rather than creating new `Randomly` here, use the one from the global state object.

src/sqlancer/stonedb/StoneDBProvider.java
```diff
@@ -42,7 +41,7 @@ protected StoneDBSchema readSchema() throws Exception {
     enum Action implements AbstractAction<StoneDBGlobalState> {
         SHOW_TABLES((g) -> new SQLQueryAdapter("SHOW TABLES")), //
         ALTER_TABLE(StoneDBTableAlterGenerator::generate), //
-        DELETE(StoneDBDeleteGenerator::generate), //
+        DELETE(sqlancer.stonedb.gen.StoneDBTableDeleteGenerator::generate), //
```
SuriZhang, line 44: remove blank comment lines

src/sqlancer/stonedb/gen/StoneDBTableDeleteGenerator.java
```diff
@@ -37,23 +39,25 @@ public SQLQueryAdapter getQuery() {
         }
         sb.append(" FROM ");
         sb.append(randomTable.getName());
-        if (Randomly.getBoolean()) {
-            sb.append(" AS ");
-            sb.append(globalState.getSchema().getRandomTable().getName());
-        }
+        // do not need AS if does not use tbl_alias later
+        // if (Randomly.getBooleanWithSmallProbability()) {
```
SuriZhang, line 43: remove commented lines ( if you will need this in the future then keep it here should be fine for now)

src/sqlancer/stonedb/gen/StoneDBTableCreateGenerator.java
```diff
@@ -145,15 +149,14 @@ private void appendTableOptions() {
     }
 
     private void addCommonErrors(ExpectedErrors list) {
-        list.add("The storage engine for the table doesn't support");
+        // java.sql.SQLSyntaxErrorException: BLOB/TEXT column 'c0' used in key specification without a key length
```
SuriZhang, line 152: i think it's okay to remove those comments , but it's your call to keep them or remove

src/sqlancer/stonedb/StoneDBProvider.java
ZhengLin-Li: Blank comment lines are used to place each action on each line. If removed, two actions will be placed on one line by the formatter, which I think is not as good as each action on one line?

src/sqlancer/stonedb/gen/StoneDBTableCreateGenerator.java
ZhengLin-Li: How about leaving it here to know each expected error's source?

ZhengLin-Li:
Hi, @mrigger @SuriZhang @bajinsheng . This PR is now ready for review.  :blush:   :raised_hands:

src/sqlancer/stonedb/gen/StoneDBIndexDropGenerator.java
```diff
@@ -29,9 +29,13 @@ private SQLQueryAdapter getQuery() {
         sb.append(" ON ");
         sb.append(table.getName());
         appendAlgoOrLockOption();
+        addExpectedErrors();
         return new SQLQueryAdapter(sb.toString(), errors);
     }
 
+    private void addExpectedErrors() {
```
mrigger, line 36: I guess we don't know need this method.

src/sqlancer/stonedb/gen/StoneDBInsertGenerator.java
```diff
@@ -45,9 +45,25 @@ private SQLQueryAdapter getQuery() {
         sb.append(table.getName());
         appendPartition();
         appendColumnsAndValues(columns);
+        addExpectedErrors();
         return new SQLQueryAdapter(sb.toString(), errors);
     }
 
+    private void addExpectedErrors() {
```
mrigger, line 52: I just noticed that `private final StringBuilder sb = new StringBuilder();` in this class is redundant, since the field is already defined in the superclass.

src/sqlancer/stonedb/gen/StoneDBTableCreateGenerator.java
```diff
@@ -51,11 +55,29 @@ public SQLQueryAdapter getQuery() {
...
+    private void addExpectedErrors() {
+        // java.sql.SQLSyntaxErrorException: BLOB/TEXT column 'c0' used in key specification without a key length
+        errors.add("used in key specification without a key length");
+        // java.sql.SQLException: Tianmu engine does not support unique index.
+        errors.add("Tianmu engine does not support unique index");
+        // java.sql.SQLException: BLOB column 'c0' can't be used in key specification with the used table type
+        errors.add("can't be used in key specification with the used table type");
+        // java.sql.SQLException: Table storage engine for 't0' doesn't have this option
+        errors.add("doesn't have this option");
+        // java.sql.SQLSyntaxErrorException: Table 't0' uses an extension that doesn't exist in this MySQL version
```
mrigger, line 72: Why are we getting this error? Is this due to a feature that StoneDB does not support, but MySQL does? If there is no way to exercise the feature, we should probably not test it.

src/sqlancer/stonedb/gen/StoneDBTableUpdateGenerator.java
```diff
@@ -39,9 +39,13 @@ public SQLQueryAdapter getQuery() {
         if (Randomly.getBoolean()) {
             appendLimit();
         }
+        addExpectedErrors();
         return new SQLQueryAdapter(sb.toString(), errors);
     }
 
+    private void addExpectedErrors() {
```
mrigger, line 46: I think we don't need this method. Let's add it if we actually need it.

src/sqlancer/stonedb/StoneDBProvider.java
```diff
@@ -59,10 +59,18 @@ public Query<?> getQuery(StoneDBGlobalState globalState) throws Exception {
     }
 
     private static int mapActions(StoneDBGlobalState globalState, Action a) {
-        globalState.getRandomly();
+        Randomly r = globalState.getRandomly();
         switch (a) {
         case SHOW_TABLES:
             return 1;
+        case ALTER_TABLE:
+            return Randomly.smallNumber();
```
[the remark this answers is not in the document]
bajinsheng: ?

bajinsheng (changes requested):
Overall LGTM, except for a small comment about `Random`.

src/sqlancer/stonedb/StoneDBProvider.java
```diff
@@ -59,10 +59,18 @@ public Query<?> getQuery(StoneDBGlobalState globalState) throws Exception {
...
     private static int mapActions(StoneDBGlobalState globalState, Action a) {
-        globalState.getRandomly();
+        Randomly r = globalState.getRandomly();
         switch (a) {
         case SHOW_TABLES:
             return 1;
+        case ALTER_TABLE:
+            return r.getInteger(0, 2);
+        case DELETE:
+            return Randomly.smallNumber();
```
bajinsheng, line 69: it requires as well.

SuriZhang:
Overalll LGTM (other than the `Random` integer Jinsheng pointed out).  Once you fixed that, you can re-request review otherwise merging is blocked

bajinsheng:
Have you resolved my comments?

ZhengLin-Li:
> Have you resolved my comments?

Some comments need future clearification, not sure how to change. For example, what is required [here](https://github.com/sqlancer/sqlancer/pull/846#discussion_r1251413256)? Have I got the reason right [here](https://github.com/sqlancer/sqlancer/pull/846#discussion_r1246714469)?

bajinsheng:
It is the only comment I gave and I also indicated which functions you should use. You can check other dbms how they specify the number here for each action.

ZhengLin-Li:
> It is the only comment I gave and I also indicated which functions you should use. You can check other dbms how they specify the number here for each action.

done

src/sqlancer/stonedb/gen/StoneDBTableCreateGenerator.java
mrigger: @ZhengLin-Li, any updates on this?

ZhengLin-Li:
reply to: https://github.com/sqlancer/sqlancer/pull/846#discussion_r1254341280

I found the reason is the COMPRESSION option when creating tables. For example, CREATE TEMPORARY TABLE t0(c0 BLOB) COMPRESSION = 'LZ4'; TIANMU engine (aka StoneDB engine) does not support this option.

So let's remove this option when creating table?

BTW, seams others can not see my reply? it is strange.
image
image

mrigger:
> So let's remove this option when creating table?

If StoneDB does not support something, we shouldn't generate it, yes.

Yes, we cannot see your responses. Could it be that you started a review on your PR? I typically get the "Pending" status when I start a review.

bajinsheng (approved):

bajinsheng:
@mrigger Although StoneDB's CI still fails, @ZhengLin-Li has fixed several bugs in StoneDB's implementation. Considering fixing all bugs in single PR would make it too huge, we would like to merge this PR now, and create another PR to continue fix other bugs.

src/sqlancer/stonedb/gen/StoneDBTableDeleteGenerator.java
```diff
@@ -37,24 +39,25 @@ public SQLQueryAdapter getQuery() {
...
         if (Randomly.getBoolean()) {
             sb.append(" LIMIT ");
-            sb.append(new Randomly().getInteger());
+            sb.append(r.getInteger(0, (int) randomTable.getNrRows(globalState)));
         }
+        addExpectedErrors();
         return new SQLQueryAdapter(sb.toString(), errors);
     }
+
+    private void addExpectedErrors() {
```
mrigger, line 61: As mentioned before, I don't think we need to introduce such empty methods.

bajinsheng: @ZhengLin-Li Can you remove them?

ZhengLin-Li: done

PR #847, opened 2023-07-01
feat: Implement an AST based reducer prototype

ColinYoungTaro:
+ The reduer can run recursively visiting the ast-tree node and transforming the structure of the query. The reducing transformations includes:
  + remove some clauses such as where, having, etc.
  + remove some elements of a list. 
  + replace an binary expression with its left or right sub expression.
  + replace a constant long or double value with null or shorter literal value, e.g, `3.1415926535` -> `3.1`; `123431`->`0`;
+ Provide some test cases at `test/reducer/TestASTBasedReducer.java` demonstrating the effect of the reducer.
+ Currently the reducer focus on only `Select statements`. Could be extended later.
+ The code of this prototype might be hacky and ugly. Could be optimized on future commits.


.settings/org.eclipse.jdt.apt.core.prefs
```diff
@@ -0,0 +1,4 @@
+eclipse.preferences.version=1
```
mrigger, line 1: I guess we can revert updating the settings here and the file below?

src/sqlancer/ASTBasedReducer.java
```diff
@@ -0,0 +1,470 @@
...
+            // try removing a pair of brackets.
+            Parenthesis paren = (Parenthesis) expr;
+            return List.of(paren.getExpression());
+        } else if (expr instanceof BinaryExpression) {
+            return transformExpr((BinaryExpression) expr);
+        } else if (expr instanceof Between) {
+            return transformExpr((Between) expr);
+        } else if (expr instanceof LongValue) {
+            LongValue longValue = (LongValue) expr;
+            if (longValue.getValue() >= 1000) {
```
mrigger, line 48: Why are we checking for 1000? Would it not also make sense to reduce a value like 999?

src/sqlancer/ASTBasedReducer.java
```diff
@@ -0,0 +1,470 @@
...
+                return List.of(new NullValue(), new LongValue(10), new LongValue(0), new LongValue(1));
+            }
+            return new ArrayList<>();
+        } else if (expr instanceof DoubleValue) {
+            DoubleValue doubleValue = (DoubleValue) expr;
+            double literal = doubleValue.getValue();
+            if (String.valueOf(literal).length() <= 4) {
```
mrigger, line 55: Why are we checking for 4 here?

src/sqlancer/ASTBasedReducer.java
```diff
@@ -0,0 +1,470 @@
...
+            expressionReduce(selectExpressionItem, expr, (item, e) -> {
+                item.setExpression(e);
+                updateStatements();
+            });
+            super.visit(selectExpressionItem);
+        }
+    };
+
+    SelectVisitorAdapter selectReducerVisitor = new SelectVisitorAdapter() {
+        private final String[] removeList = { "Limit", "Offset", "Where", "Having", "GroupBy", "Distinct",
```
mrigger, line 270: I think it would be useful to have some comments for what these three arrays mean/how they are used

src/sqlancer/ASTBasedReducer.java
```diff
@@ -0,0 +1,470 @@
...
+                return "set" + astNodeName;
+            }
+        }
+
+        @Override
+        public void visit(PlainSelect plainSelect) {
+            // Remove Section
+            for (String astNodeName : removeList) {
+                try {
+                    Method nodeGetter = plainSelect.getClass().getMethod(getterName(astNodeName));
```
mrigger, line 292: Based on my understanding, we are using reflection because we cannot change the fields otherwise, right? If so, perhaps, we could also add a comment mentioning this somewhere

src/sqlancer/ASTBasedReducer.java
```diff
@@ -0,0 +1,470 @@
...
+                            try {
+                                nodeSetter.invoke(select, expression);
+                                updateStatements();
+                            } catch (IllegalAccessException | InvocationTargetException e) {
+                                e.printStackTrace();
+                            }
+                        });
+                    } else if (astNode instanceof List) {
+                        List<?> elms = (List<?>) astNode;
+                        if (elms.size() <= 1) {
```
mrigger, line 332: Can we sometimes reduce lists with one element to lists with zero elements? In some DBMSs, this is valid. See https://www.db-fiddle.com/f/vXcJLN119VekRuz1jJuyJ5/0

src/sqlancer/ASTBasedReducer.java
```diff
@@ -0,0 +1,470 @@
...
+                        GroupByElement groupByElement = (GroupByElement) astNode;
+                        ExpressionList expressionList = groupByElement.getGroupByExpressionList();
+                        if (expressionList != null) {
+                            expressionList.accept(expressionReducerVisitor);
+                        } else {
+                            // TODO: TO BE IMPLEMENTED
+                        }
+                    }
+                } catch (Exception e) {
+                    System.out.println("Reflection Exception:");
```
mrigger, line 393: perhaps we can just throw an `AssertionError` here instead, passing `e` to its constructor.

test/sqlancer/reducer/TestASTBasedReducer.java
```diff
@@ -0,0 +1,129 @@
...
+        String[] queriesStr = {
+                "SELECT DISTINCT * FROM v0 WHERE ((v0.rowid || ( (v0.c + v0.d) < 200 && v0.c >= 100) || 114514)OR(((v0.c0)||(1529686005)))) UNION SELECT DISTINCT * FROM v0 WHERE (NOT ((v0.rowid)OR(((v0.c0)||(1529686005))))) UNION SELECT DISTINCT * FROM v0 WHERE ((((v0.rowid)OR(((v0.c0)||(1529686005))))) IS NULL)" };
+        env.setInitialStatementsFromStrings(List.of(queriesStr));
+        env.setBugInducingCondition(statements -> {
+            String queriesString = TestEnvironment.getQueriesString(statements);
+            return queriesString.contains("&&");
+        });
+        env.runReduce();
+        List<Query<?>> reducedResult = env.getReducedStatements();
+        System.out.println(Arrays.toString(queriesStr));
```
mrigger, line 23: I guess we could already check against the result that this currently produces? We can still update the test case later when your implementation changes.

test/sqlancer/reducer/TestEnvironment.java
```diff
@@ -21,8 +21,26 @@ public class TestEnvironment {
     private VirtualDBProvider provider = null;
     private VirtualDBGlobalState state, newGlobalState;
 
-    public TestEnvironment() throws Exception {
+    private Reducer<VirtualDBGlobalState> reducer = null;
+
+    private static final int USING_STATEMENT_REDUCER = 1;
```
mrigger, line 26: Here, we should use a Java enum, but it's also fine to do this in a follow-up PR.

mrigger (commented):
Thanks a lot! As mentioned during our meeting, I think it's fine that the code is a bit hacky, given that it's an independent component that does not have a direct impact on the maintenance of the other DBMS implementations. I added some minor comments. @oraluben and @DerZc might have some more.

src/sqlancer/ASTBasedReducer.java
```diff
@@ -0,0 +1,470 @@
...
+            double literal = doubleValue.getValue();
+            if (String.valueOf(literal).length() <= 4) {
+                return new ArrayList<>();
+            }
+            double roundedValue = Math.round(literal * 10.0) / 10.0;
+            return List.of(new NullValue(), new DoubleValue(String.valueOf(roundedValue)));
+        } else if (expr instanceof CaseExpression) {
+            CaseExpression caseExpression = (CaseExpression) expr;
+            return List.of(caseExpression.getSwitchExpression(), caseExpression.getElseExpression());
+        } else
```
DerZc, line 63: Do we need to shorten other constant value like String?

src/sqlancer/ASTBasedReducer.java
```diff
@@ -0,0 +1,470 @@
...
+                updateStatements();
+            });
+            super.visit(selectExpressionItem);
+        }
+    };
+
+    SelectVisitorAdapter selectReducerVisitor = new SelectVisitorAdapter() {
+        private final String[] removeList = { "Limit", "Offset", "Where", "Having", "GroupBy", "Distinct",
+                "OrderByElements", "Joins" };
+        private final String[] pullUpList = { "Where", "Having", "FromItem", "SelectItems", "GroupBy", "Joins" };
```
DerZc, line 272: can we support common table expression (https://dev.mysql.com/doc/refman/8.0/en/with.html) here
and derived tables (https://dev.mysql.com/doc/refman/8.0/en/derived-tables.html)

src/sqlancer/ASTBasedReducer.java
ColinYoungTaro: In fact, I plan to shorten the value and let them be less then 4 digits. So the constant values (including integer and float values) which are shorter than that would not be checked and reduced. 
Also, There is another aspect that needs to be considered:
``` java
do {
    observeChange = false;
    List<Expression> candidates = ExpressionTransformer.candidateExpressions(subExpr);
    for (Expression candidate : candidates) {
        try {
            setter.accept(parent, candidate);
            if (bugStillTriggers()) {
                subExpr = candidate;
                observeChange = true;
            }
        } catch (Exception ignoredException) {
        }
    }
    setter.accept(parent, subExpr);
} while (observeChange);
```

For an expression, An array of candidate expressions would be generated. (e.g. for a long value, it would be [null, 0, 1]). 

In the first round of while-loop, the expression would become `1`, but in the next round of loop, null, 0, 1 would be tried again respectively. The result would also be 1 and the `observeChange` would be true. Thus we've gone into an infinite loop. 

Therefore, we have to ensure that the constant values in the candidate list would not be reduced again. (negative long value should also be checkd, which I didn't do in this version.) 

Shortening those values longer than a certain length could be a simple solution to avoid infinite loop. Maybe we could discuss and come up with some other approaches for this problem.

src/sqlancer/ASTBasedReducer.java
ColinYoungTaro: Relection is used to avoid repetitive code. For example, when trying removing clauses of an Select statement, removing `when`, `limit`, `offset` requires almost the same procedure. But the getter, setter method is not the same. I have to repeat the same code but replace the method name. e.g.: `getWhen, getLimit, getOffset`. Using reflection would make this procedure simpler.

src/sqlancer/ASTBasedReducer.java
ColinYoungTaro: OK, I think the 1-element constraint could be removed. :)

src/sqlancer/ASTBasedReducer.java
ColinYoungTaro: Yes, String value could be reduction. I would add it in the later PR.

oraluben:
LGTM in general, and @mrigger has already posted most of my comments : )

Besides future fixing and updates, I think another major thing we want is a tutorial about the reducer (both delta-debugging and AST-based), how they briefly work, how to use, what's expected, etc.

src/sqlancer/ASTBasedReducer.java
ColinYoungTaro: Theoretically both could be handled. 
For derived tables I've added support for it. :
Given a statement:
```sql 
SELECT AVG(sum_column1)
FROM (SELECT SUM(column1) AS sum_column1 FROM t1 GROUP BY column1 LIMIT 32 OFFSET 128) AS t1;
```
The result is:
``` sql 
SELECT AVG(sum_column1) FROM (SELECT SUM(column1) AS sum_column1 FROM t1) AS t1;
```
As for common table expression, it hasn't be supported yet.

oraluben (approved):

oraluben (changes requested):
Test's failing due to format issue.

mrigger:
After fixing this, I think we can go ahead and merge the PR as indicated by @oraluben.

mrigger (approved):

oraluben:
For future works, you could create a long-term issue and use [tasklist](https://docs.github.com/en/issues/tracking-your-work-with-issues/quickstart#adding-a-task-list) to track all the tasks related to the AST reducer. @ColinYoungTaro

PR #849, opened 2023-07-06
feat: add MySQLUpdateGenerator class

eltfshr:
This PR adds an update generator class for MySQL and adds update action to the MySQL Provider. It also extracts common expected errors from insert and update generators.

### DIFF
--- a/src/sqlancer/mysql/gen/MySQLUpdateGenerator.java
+++ b/src/sqlancer/mysql/gen/MySQLUpdateGenerator.java
@@ -38,14 +38,8 @@ public class MySQLUpdateGenerator extends AbstractUpdateGenerator<MySQLColumn> {
             MySQLErrors.addExpressionErrors(errors);
             sb.append(MySQLVisitor.asString(gen.generateExpression()));
         }
-        errors.add("doesn't have a default value");
-        errors.add("Data truncation");	
-        errors.add("Incorrect integer value");
-        errors.add("Duplicate entry");
-        errors.add("Data truncated for functional index");
-        errors.add("Data truncated for column");
-        errors.add("cannot be null");
-        errors.add("Incorrect decimal value");
+        MySQLErrors.addInsertUpdateErrors(errors);
+        errors.add("doesn't have this option");
 
         return new SQLQueryAdapter(sb.toString(), errors);
     }
--- a/src/sqlancer/mysql/gen/MySQLUpdateGenerator.java
+++ b/src/sqlancer/mysql/gen/MySQLUpdateGenerator.java
@@ -1,0 +1,64 @@
+package sqlancer.mysql.gen;
+
+import java.sql.SQLException;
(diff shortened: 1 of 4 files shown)


src/sqlancer/mysql/gen/MySQLUpdateGenerator.java
```diff
@@ -0,0 +1,64 @@
...
+        sb.append("UPDATE ");
+        sb.append(table.getName());
+        sb.append(" SET ");
+        updateColumns(columns);
+        if (Randomly.getBoolean()) {
+            sb.append(" WHERE ");
+            MySQLErrors.addExpressionErrors(errors);
+            sb.append(MySQLVisitor.asString(gen.generateExpression()));
+        }
+        errors.add("doesn't have a default value");
```
mrigger, line 41: Some of these seem similar to the ones in https://github.com/sqlancer/sqlancer/blob/5bcd9ce69898f2866094535610ab0b4c04778843/src/sqlancer/mysql/gen/MySQLInsertGenerator.java#L91-L98. Could we extract the common expected errors and move them to https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/mysql/MySQLErrors.java?

src/sqlancer/mysql/gen/MySQLUpdateGenerator.java
```diff
@@ -0,0 +1,64 @@
...
+        errors.add("Data truncation");	
+        errors.add("Incorrect integer value");
+        errors.add("Duplicate entry");
+        errors.add("Data truncated for functional index");
+        errors.add("Data truncated for column");
+        errors.add("cannot be null");
+        errors.add("Incorrect decimal value");
+
+        return new SQLQueryAdapter(sb.toString(), errors);
```
mrigger, line 50: Looking at the CI tests, it seems that some expected errors are still missing.

mrigger:
You can execute `mvn formatter:format` to fix the formatting issues, which currently causes the general tests to fail.

src/sqlancer/mysql/gen/MySQLUpdateGenerator.java
bajinsheng: @eltfshr As we discussed in the meeting, can you check whether these errors are necessary? You can remove some expected errors that will not happen in the UPDATE statement, instead of copying them from INSERT statement.

eltfshr: Yes, I checked each of errors, except "Data truncated for functional index". That one was a bug Manuel reported to MySQL, so I left it in. But on further analysis, I think it is not related to UPDATE. I will remove that one.

eltfshr: Common expected errors are now extracted.

bajinsheng (approved):

PR #854, opened 2023-07-10
fix & refactor: fix StoneDB implementation bugs and refactor code, Part 2 

ZhengLin-Li:
In this PR, we fix some implementation bugs and refactor code to achieve better code logic.

change list:
- change the appendType function, moving it from createTable to enum, to reused code when creating and altering tables;
- change `castOperationNode` to be compilable to MySQL syntax.
- disable `create temporary table`, because it caused some bugs so we plan to disable it now and support this feature in the future.
- delete `rename` column feature when altering tables because this is a new feature of MySQL 8.0, but we now support StoneDB 5.7.
- rename some variables


bajinsheng:
> com.mysql.cj.exceptions.CJCommunicationsException: Communications link failure appears frequently, but not sure the reason.

src/sqlancer/stonedb/gen/StoneDBTableAlterGenerator.java
```diff
@@ -27,19 +32,48 @@ public static SQLQueryAdapter generate(StoneDBGlobalState globalState) {
...
-                .setColumns(table.getColumns());
         sb.append(table.getName());
         sb.append(" ");
-        Action action = Randomly.fromOptions(Action.values());
+        appendAlterOptions();
+        addExpectedErrors();
+        return new SQLQueryAdapter(sb.toString(), errors, true);
+    }
+
+    private void addExpectedErrors() {
```
mrigger, line 42: For future expected errors: I think it would be better to inline the expected errors. They should be as localized as possible, so we don't miss any bugs that we could otherwise easily catch.

ZhengLin-Li: got it, will add errors immediately after the code that will cause such an error.

ZhengLin-Li:
> > com.mysql.cj.exceptions.CJCommunicationsException: Communications link failure appears frequently, but not sure the reason.
> 
> You can check the error before these communication errors. It is probably because your testing triggers a crash bug in StoneDB, and all following testing fail to connect to the server.

I found that the first statement resulted in the communication errors. Maybe it is the bug we previous found caused this carsh because the `delete from` statement was the ever first statement our tests ran.

image

bajinsheng:
For the crash issues that have not been fixed yet, you can consider restrict the corresponding SQL statements generation.

mrigger (approved):
LGTM! @bajinsheng, any comments?

bajinsheng (approved):
@mrigger I have no more comments, and this PR is probably able to be merged.

PR #864, opened 2023-08-01
feat: write log when running statement reducer

### LINKED ISSUE #865: add support of writing reduced result to log files.

ColinYoungTaro:
Closes #865 

+ write log when start reducing, finish reducing and when a reduction step is successfully performed
log is saved into `logs/<database_>/reduce/<database_name>-reduce.log`, for example: When running statement reducer for sqlite3, the log of databse0 is stored at `logs/sqlite3/reduce/database0-reduce.log`
The format is like:
```
Original queries:
====================
xxx original statements
====================
Current reduced result:
====================
xxx
====================
Current reduced result:
====================
xxx
...
====================
Current reduced result:
====================
xxx
====================
Reduced queries:
====================
CREATE VIRTUAL TABLE rt0 USING rtree_i32(c0, c1, c2);
CREATE TEMP TABLE t1 (c0 INTEGER );
INSERT OR IGNORE INTO rt0(c0, c1, c2) VALUES ('-1e500', '', x'');
INSERT OR FAIL INTO t1(c0) VALUES (NULL), (0.9692699099404161), ('');
INSERT OR IGNORE INTO rt0(c2) VALUES (''), (x''), ('');
====================

```
+ fix some bugs of the statement reducer.
Adjust the code about options of time limit and step limit, making it more reasonable

### DIFF
--- a/src/sqlancer/Main.java
+++ b/src/sqlancer/Main.java
@@ -5,6 +5,8 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.Writer;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
@@ -130,6 +132,7 @@ public final class Main {
                     reduceFileDir.mkdir();
                 }
                 this.reduceFile = new File(reduceFileDir, databaseName + "-reduce.log");
+
             }
             this.databaseProvider = provider;
         }
@@ -265,6 +268,16 @@ public final class Main {
             }
             try {
                 reduceFileWriter.write(sb.toString());
+
+                File symbolicLinkToReduceFile = new File(LOG_DIRECTORY, "current-reduce.log");
+                Path reduceFilePath = reduceFile.toPath().toAbsolutePath();
+                Path symbolicPath = symbolicLinkToReduceFile.toPath().toAbsolutePath();
+                if(!Files.readSymbolicLink(symbolicLinkToReduceFile.toPath()).toAbsolutePath().equals(reduceFilePath)) {
+                    Files.deleteIfExists(symbolicPath);
+                    Files.createSymbolicLink(symbolicPath, reduceFilePath);
+                    System.out.println("Create symbolic link from " + symbolicPath + " linked " + reduceFilePath);
+                }
+
             } catch (IOException e) {
                 throw new AssertionError(e);
             }finally {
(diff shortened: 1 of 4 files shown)


mrigger:
From a user perspective, I think it would be better to always overwrite the reduced file with the currently minimal version, since users are likely only interested in using that version. Perhaps we can include the additional information when a debugging flag is enabled?

oraluben:
The idea looks fine.

Besides @mrigger's comment, I think it's better to logger mechanism (`sqlancer.common.log.LoggableFactory`) for better scalability (e.g. print to stdout and file at same time).

> From a user perspective, I think it would be better to always overwrite the reduced file with the currently minimal version, since users are likely only interested in using that version. Perhaps we can include the additional information when a debugging flag is enabled?

When all outputs are kept, a common practice is to have a symbolic link to the latest file, so user can always use a shorter path for it, i.e. `logs/current.log`.

src/sqlancer/Main.java
```diff
@@ -265,6 +268,16 @@ public void logReduced(StateToReproduce state) {
             }
             try {
                 reduceFileWriter.write(sb.toString());
+
+                File symbolicLinkToReduceFile = new File(LOG_DIRECTORY, "current-reduce.log");
```
ColinYoungTaro, line 270: it works on windows. But I haven't tested it on Linux.
Currently, The symbolic link file refer to the latest modified file, which means, if a.log is created before b.log, but is modified later, the symlink file will link to `a.log`. I'm not sure if it's appropriate.
Also, I'm not sure if the concurrent issues should be considered here.

oraluben: > Currently, The symbolic link file refer to the latest modified file, which means, if `a.log` is created before `b.log`, but is modified later, the symlink file will link to `a.log`. I'm not sure if it's appropriate.
> Also, I'm not sure if the concurrent issues should be considered here.

I'd prefer to create the link when file created, not modified, i.e. it will link to the recent created file.

oraluben:
The code change looks fine. We should add some doc about how to understand the logs.

src/sqlancer/Main.java
ColinYoungTaro: > I'd prefer to create the link when file created, not modified, i.e. it will link to the recent created file.

In this way the previous log file should be deleted first. I think `ensureExistsAndIsEmpty` function should be altered a bit so that the log file in the sub dictionary `reduce` could be removed.

oraluben: I see, you're right. I missed the other logs. So we might not need to create a separate link for the reduce log?

PR #869, opened 2023-08-09
feat: add support of Fuzz oracle

ZhengLin-Li:
feat: add support of Fuzz oracle

The StoneDB is so weak to pass all our tests so that we need a oracle only to test it will not crash. Such oracle is Fuzz oracle. It will not care about the query result, but focus on not letting StoneDB crash.


bajinsheng:
It is strange that the CI does not run.

ZhengLin-Li:
> It is strange that the CI does not run.

will force push again and let CI run.

src/sqlancer/stonedb/oracle/StoneDBFuzzOracle.java
```diff
@@ -0,0 +1,82 @@
...
+        Node<StoneDBExpression> randomWhereCondition = gen.generateExpression();
+        List<StoneDBTable> tables = randomTables.getTables();
+        List<TableReferenceNode<StoneDBExpression, StoneDBTable>> tableList = tables.stream()
+                .map(t -> new TableReferenceNode<StoneDBExpression, StoneDBTable>(t)).collect(Collectors.toList());
+        List<Node<StoneDBExpression>> joins = StoneDBJoin.getJoins(tableList, globalState);
+
+        StoneDBSelect select = new StoneDBSelect();
+        Node<StoneDBExpression> asText = new NewPostfixTextNode<>(
+                new StoneDBCastOperation(
+                        new NewBinaryOperatorNode<>(new NewPostfixTextNode<>(randomWhereCondition, " IS NOT NULL "),
```
mrigger, line 56: Why do we need this specific expression in the fetch clause?

ZhengLin-Li: no special reasons. I just copied from the NoRec implementation. 😂

If it is not necessary, we can change it it `*` ?

mrigger: Yes, that, and perhaps selecting some random column subset with another probability.

ZhengLin-Li: that make sense. have done.

bajinsheng (approved):

src/sqlancer/stonedb/oracle/StoneDBFuzzOracle.java
```diff
@@ -0,0 +1,73 @@
...
+        select.setFetchColumns(
+                Randomly.fromOptions(List.of(new ColumnReferenceNode<>(new StoneDBColumn("*", null, false, false, 0))),
+                        allColumns, Randomly.subset(allColumns)));
+        select.setFromList(new ArrayList<>(tableList));
+        select.setJoinList(joins);
+        String queryString = StoneDBToStringVisitor.asString(select);
+        SQLQueryAdapter q = new SQLQueryAdapter(queryString, errors);
+        q.executeAndGetLogged(globalState);
+
+        select = new StoneDBSelect();
```
mrigger, line 60: Why do we create two queries here? This causes quite a bit of redundancy, and also causes the logging mechanism to show an incorrect number of executed tests.

ZhengLin-Li: I originally designed this oracle as a subset of the NoREC oracle, which means that they will have the same kinds of queries executed. However, the difference between Fuzz oracle and NoREC oracle is that Fuzz does not care about the query result, but NoREC does. 

Does this make sense? If we do not need to follow the statement structure of NoREC, we can have only one select statement here.

mrigger: Yes, some other DBMS implementations also have a fuzzer oracle. I think we want to just generate random queries, but I don't understand why we need to instantiate two queries for each fuzzing iteration. I think one would suffice, and result in cleaner code.

ZhengLin-Li:
Not sure why the general tests are failing.  I run `verify -DskipTests -f pom.xml` in my local computer and it's result meets expectation.

image

bajinsheng:
It is strange and I cannot reproduce the error in the CI as well,

From the CI:
> Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.14.0:check (pmd) on project sqlancer: You have 1 PMD violation. For more details see: /home/runner/work/sqlancer/sqlancer/target/pmd.xml -> [Help 1]
It is PMD violation.

@ZhengLin-Li Can you try to run all commands from the CI in the new docker image?

mrigger:
When running the respective test, it would also be possible to add a `| cat /home/runner/work/sqlancer/sqlancer/target/pmd.xml` or so to see about what PMD complains?

mrigger:
My guess is that PMD complains about the lack of capitalization in the enum (`Fuzz` rather than `FUZZ`), which is an antipattern (see https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html).

src/sqlancer/stonedb/StoneDBOptions.java
```diff
@@ -30,6 +31,12 @@ public class StoneDBOptions implements DBMSSpecificOptions<StoneDBOracleFactory>
     public List<StoneDBOracleFactory> oracles = List.of(StoneDBOracleFactory.NOREC);
 
     public enum StoneDBOracleFactory implements OracleFactory<StoneDBGlobalState> {
+        Fuzz {
```
mrigger, line 34: Perhaps this could be `FUZZER`, so we use the same name as the other implementations?

ZhengLin-Li: sure, will change

ZhengLin-Li:
> It is strange and I cannot reproduce the error from the CI as well,
> 
> From the CI:
> 
> > Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.14.0:check (pmd) on project sqlancer: You have 1 PMD violation. For more details see: /home/runner/work/sqlancer/sqlancer/target/pmd.xml -> [Help 1]
> 
> It is PMD violation.
> 
> @ZhengLin-Li Can you try to run all commands from the CI in a clean docker image?

I re-run it on a clean Ubuntu. And it's turned out the nested if statements is the cause. But accroding to https://github.com/sqlancer/sqlancer/pull/882#discussion_r1301805320, we will not need to change it?

image

mrigger:
>Accroding to https://github.com/sqlancer/sqlancer/pull/882#discussion_r1301805320, we will not need to change it? But if we do not change, the CI will still fail.

Thanks for spotting this. In this case, we'll need to undo the change, as in your PR (cc @ColinYoungTaro and @oraluben).

mrigger (approved):

PR #870, opened 2023-08-09
fix: drop features of StoneDB

ZhengLin-Li:
fix: drop features that StoneDB not supported or crash causing

In this PR, we delete some features that StoneDB does not support or some features that will lead to the crash of StoneDB.


bajinsheng:
Since there are still many other crashing issues, I think we do not need to bypass this crash issue anymore. We can close it.

PR #875, opened 2023-08-10
fix: avoid logging CREATE TABLE exception

JensonSung:
Current implement cannot log CREATE TABLE statements, exceptions are generated.
Fix this by setting the `couldAffectSchema` flag to true for CREATE TABLE statements.


src/sqlancer/common/log/SQLLoggableFactory.java
```diff
@@ -24,7 +24,11 @@ protected Loggable createLoggable(String input, String suffix) {
 
     @Override
     public SQLQueryAdapter getQueryForStateToReproduce(String queryString) {
-        return new SQLQueryAdapter(queryString);
+        if (queryString.contains("CREATE TABLE")) {
```
oraluben, line 27: IMO changes could take place in the default constructor of `SQLQueryAdapter`, not here. That class takes care of those stuffs, i.e. `sqlancer.common.query.SQLQueryAdapter#checkQueryString`.

test/sqlancer/TestLoggableFactory.java
```diff
@@ -0,0 +1,15 @@
...
+import org.junit.jupiter.api.Test;
+import sqlancer.common.log.SQLLoggableFactory;
+
+public class TestLoggableFactory {
+
+    @Test
+    public void testLogCreateTable() {
+        String query = "CREATE TABLE t1 (c1 INT)";
+        SQLLoggableFactory logger = new SQLLoggableFactory();
+        logger.getQueryForStateToReproduce(query);
```
oraluben, line 12: Current test didn't check for any oracles and cannot fail, could you add some assertion to test if a create statement is "loggable"? Maybe check `SQLQueryAdapter.couldAffectSchema` field.

oraluben (commented):
Thanks for the contribution! The change seems reasonable. I've left some comments that may improve the readability of this change.

src/sqlancer/common/log/SQLLoggableFactory.java
JensonSung: Agree. Such changes are more reasonable.

test/sqlancer/TestLoggableFactory.java
JensonSung: Agree. I make such changes.

src/sqlancer/common/log/SQLLoggableFactory.java
JensonSung: Does the following change confuse other users?
```Java
    public SQLQueryAdapter(String query) {
        this(query, new ExpectedErrors(), query.contains("CREATE TABLE") && !query.startsWith("EXPLAIN"));
    }
```

oraluben: I'd use something like:
```
    public SQLQueryAdapter(String query, ExpectedErrors expectedErrors) {
        this(query, expectedErrors, guessAffectSchemaFromQuery(query));
    }

    private static boolean guessAffectSchemaFromQuery(String query) {
        return query.contains("CREATE TABLE") && !query.startsWith("EXPLAIN");
    }

    private void checkQueryString() {
        if (guessAffectSchemaFromQuery(query) != couldAffectSchema) {
            throw new AssertionError("CREATE TABLE statements should set couldAffectSchema to true");
        }
    }
```

JensonSung: Agree. Thanks for help.
I change the committed code.
PTAL

oraluben (approved):

src/sqlancer/common/query/SQLQueryAdapter.java
```diff
@@ -46,7 +50,7 @@ private String canonicalizeString(String s) {
     }
 
     private void checkQueryString() {
-        if (query.contains("CREATE TABLE") && !query.startsWith("EXPLAIN") && !couldAffectSchema) {
+        if (guessAffectSchemaFromQuery(query) != couldAffectSchema) {
```
mrigger, line 53: The comparison here confuses me. I think the pattern before was more readable?

mrigger:
@oraluben, thanks for the ping, I've added one minor comment.

src/sqlancer/common/query/SQLQueryAdapter.java
oraluben: I move the `query.contains("CREATE TABLE") && !query.startsWith("EXPLAIN")` part to a separate method, for guess `couldAffectSchema` when not given, and to check here.

mrigger: I think extracting this part is fine, but the "!=" is confusing.

JensonSung: I think only when the flag `couldAffectSchema` is false, we need to check whether such value is proper according to the result of function `guessAffectSchemaFromQuery`. In this case, an exception could occur, such as when setting `couldAffectSchema` to true for ALTER statement. 
The following revision may be acceptable.
```java
    private void checkQueryString() {
        if (!couldAffectSchema) {
            if (guessAffectSchemaFromQuery(query)) {
                throw new AssertionError("CREATE TABLE statements should set couldAffectSchema to true");
            }
        }
    }
```

mrigger: Yes, I think that would be easy to understand, so I would prefer this.

JensonSung: Great, I have pushed the changes.
PTAL

mrigger (approved):

PR #876, opened 2023-08-10
fix: fix implementation bugs

ZhengLin-Li:
In this PR, we fix the implemnetation bugs by deleting all features that StoneDB does not support.

### DIFF
--- a/src/sqlancer/stonedb/gen/StoneDBExpressionGenerator.java
+++ b/src/sqlancer/stonedb/gen/StoneDBExpressionGenerator.java
@@ -168,7 +168,7 @@ public class StoneDBExpressionGenerator extends UntypedExpressionGenerator<Node<
 
     // https://stonedb.io/docs/SQL-reference/functions/aggregate-functions/
     public enum StoneDBAggregateFunction {
-        MAX(1), MIN(1), AVG(1), COUNT(1), FIRST(1), SUM(1);
+        MAX(1), MIN(1), AVG(1), COUNT(1), SUM(1);
 
         private int nrArgs;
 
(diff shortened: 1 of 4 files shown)


src/sqlancer/stonedb/gen/StoneDBExpressionGenerator.java
```diff
@@ -300,7 +300,7 @@ public String getTextRepresentation() {
      * Bitwise operators supported by StoneDB: https://stonedb.io/docs/SQL-reference/operators/bitwise-operators
      */
     public enum StoneDBBinaryBitwiseOperator implements Operator {
-        AND("&"), OR("|"), XOR("^"), LEFTSHIFT("<<"), RIGHTSHIFT(">>");
```
mrigger, line 303: XOR is supported, though? For unfixed bugs, we should adopt the pattern at https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#unfixed-bugs.

ZhengLin-Li: sorry for the confusion. I made a mistake in the impelmentation. `XOR` is not `StoneDBBinaryBitwiseOperator` so I delete it here. `XOR` should be `StoneDBBinaryLogicalOperator` and it already exists so it is not shown in the diff.

image
image
image

ZhengLin-Li: Ohh, Sure, I have made a mistask.

bajinsheng (approved):

PR #877, opened 2023-08-12
feat: bypass logical XOR

ZhengLin-Li:
feat: bypass logical XOR 

It seams that the expected supported operator is not actually supported so let's bypass it.

image


bajinsheng:
If it is a bug, we can bypass it in this way. If the XOR is not supported, we can directly remove the XOR.

bajinsheng (approved):

bajinsheng:
@mrigger Zhenglin explained to me it is a bug because the document shows the XOR is supported, so I think we can merge this PR.

mrigger:
Okay, sounds good. There is a failing Checkstyle check. After addressing it, we can merge the PR.

ZhengLin-Li:
> Okay, sounds good. There is a failing Checkstyle check. After addressing it, we can merge the PR.

Done

PR #879, opened 2023-08-17
feat: refactor AST-based reducer

ColinYoungTaro:
* different kinds of transformers are defined separately, making the reducer more extenable.
  * To define different types of transformations, you simply need to create classes that inherit from `Transformation` and implement specific logic. The `Transformation` class defines some common APIs such as `tryRemove`, `tryReduce`.
  * Currently, the implemented transformations are based on JSQLParser, so they all inherit from `JSQLParserBasedTransformation`, which interacts with JSQLParser. However, It is possible to define transformations not based on this tool since the `Transformation` class doesn't rely on it.
* add options to enable AST Based Reducer

### DIFF
--- a/src/sqlancer/MainOptions.java
+++ b/src/sqlancer/MainOptions.java
@@ -132,7 +132,13 @@ public class MainOptions {
     @Parameter(names = "--statement-reducer-max-steps", description = "EXPERIMENTAL Maximum steps the statement reducer will do")
     private long maxStatementReduceSteps = NO_REDUCE_LIMIT; // NOPMD
 
-    @Parameter(names = "--statement-reducer-max-time", description = "EXPERIMENTAL Maximum time duration (secs) the statement reducer will do")
+    @Parameter(names = "--statement-reducer-max-time", description = "EXPERIMENTAL Maximum time duration (secs) the AST-based reducer will do")
+    private long maxASTReduceTime = NO_REDUCE_LIMIT; // NOPMD
+
+    @Parameter(names = "--ast-reducer-max-steps", description = "EXPERIMENTAL Maximum steps the AST-based reducer will do")
+    private long maxASTReduceSteps = NO_REDUCE_LIMIT; // NOPMD
+
+    @Parameter(names = "--ast-reducer-max-time", description = "EXPERIMENTAL Maximum time duration (secs) the statement reducer will do")
     private long maxStatementReduceTime = NO_REDUCE_LIMIT; // NOPMD
 
     public int getMaxExpressionDepth() {
@@ -307,4 +313,13 @@ public class MainOptions {
     public long getMaxStatementReduceTime() {
         return maxStatementReduceTime;
     }
+
+    public long getMaxASTReduceSteps() {
+        return maxASTReduceSteps;
+    }
+
+    public long getMaxASTReduceTime() {
+        return maxASTReduceTime;
+    }
+
 }
--- a/src/sqlancer/MainOptions.java
+++ b/src/sqlancer/MainOptions.java
@@ -126,7 +126,7 @@ public class MainOptions {
     @Parameter(names = "--use-reducer", description = "EXPERIMENTAL Attempt to reduce queries using a simple reducer")
     private boolean useReducer = false; // NOPMD
 
-    @Parameter(names = "--reduce-AST", description = "EXPERIMENTAL perform AST reduction after statement reduction")
+    @Parameter(names = "--reduce-ast", description = "EXPERIMENTAL perform AST reduction after statement reduction")
     private boolean reduceAST = false; // NOPMD
 
     @Parameter(names = "--statement-reducer-max-steps", description = "EXPERIMENTAL Maximum steps the statement reducer will do")
--- a/src/sqlancer/MainOptions.java
+++ b/src/sqlancer/MainOptions.java
@@ -126,6 +126,9 @@ public class MainOptions {
     @Parameter(names = "--use-reducer", description = "EXPERIMENTAL Attempt to reduce queries using a simple reducer")
     private boolean useReducer = false; // NOPMD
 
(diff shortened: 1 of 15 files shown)


src/sqlancer/transformations/RemoveRowsOfInsert.java
```diff
@@ -0,0 +1,39 @@
...
+                    if (selBody instanceof ValuesStatement) {
+                        ValuesStatement valuesStatement = (ValuesStatement) selBody;
+                        if (valuesStatement.getExpressions() instanceof ExpressionList) {
+                            ExpressionList itemsList = (ExpressionList) valuesStatement.getExpressions();
+                            tryRemoveElms(itemsList, itemsList.getExpressions(), ExpressionList::setExpressions);
+                        }
+                    }
+                }
+            }
+        }
```
oraluben, line 37: This looks too nested, how about: 

```suggestion
        if (!(statement instanceof Insert)) {
            return;
        }

        SelectBody selectBody = ((Insert) statement).getSelect().getSelectBody();

        if (!(selectBody instanceof SetOperationList)) {
            return;
        }

        SetOperationList insertingList = (SetOperationList) selectBody;
        for (SelectBody selBody : insertingList.getSelects()) {
            if (!(selBody instanceof ValuesStatement)) {
                continue;
            }

            ValuesStatement valuesStatement = (ValuesStatement) selBody;
            ItemsList itemsList = valuesStatement.getExpressions();

            if (!(itemsList instanceof ExpressionList)) {
                continue;
            }

            tryRemoveElms((ExpressionList) itemsList, ((ExpressionList) itemsList).getExpressions(), ExpressionList::setExpressions);
        }
```

src/sqlancer/transformations/SimplifyConstant.java
```diff
@@ -0,0 +1,105 @@
...
+import net.sf.jsqlparser.statement.insert.Insert;
+import net.sf.jsqlparser.statement.select.Select;
+import net.sf.jsqlparser.util.deparser.ExpressionDeParser;
+import net.sf.jsqlparser.util.deparser.SelectDeParser;
+
+/**
+ * Shorten the constant of a statement e.g. "a_very_long_str" -> "", 12341234->1;
+ * <p>
+ * Note: Since the API of JSQLParser may have some problems with double values: when `setValue` is called, it doesn't
+ * change the string of the statement, double values are not handled here.
```
mrigger, line 20: not sure if I understand the comment here. how do double values relate to string values?

src/sqlancer/transformations/Transformation.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * The base class of transformations Defines APIs to remove, replace, remove elements of a list.
+ */
+public class Transformation {
+
+    private static Supplier<Boolean> bugJudgement;
+    protected boolean isChanged;
+    String current;
+    Statement statement;
+    // public enum StatusCode {
```
mrigger, line 19: Do we need this commented-out code?

test/sqlancer/reducer/TestASTBasedReducer.java
```diff
@@ -111,19 +151,51 @@ void testConstantVar() throws Exception {
...
+                    CCJSqlParserUtil.parse(s.getQueryString());
+                }
+            } catch (JSQLParserException e) {
+                return false;
+            }
+            return queriesString.contains("cte2");
+        });
+        env.runReduce();
+        List<Query<?>> reducedResult = env.getReducedStatements();
+        System.out.println(Arrays.toString(queriesStrs));
```
mrigger, line 175: For test cases, it might be better if we compare against the expected result here? we can still update it if the reducer becomes more effective in the future.

src/sqlancer/transformations/SimplifyConstant.java
ColinYoungTaro: In JSQLParser, The `DoubleValue` class contains two members: `String stringValue` and `double value`. When `setValue` is called, It just modifies `value` while `getString()` returns `stringValue`. When the code below is executed:
``` java
DoubleValue doubleValue = new DoubleValue("2.0");
doubleValue.setValue(1.0);
System.out.println(doubleValue);
```
The result is still `2.0`.  For more details, we can refer to `DoubleValue.class`

While for `LongValue` objects, its StringValue could be modified by calling `setStringValue`;

src/sqlancer/transformations/Transformation.java
ColinYoungTaro: > Do we need this commented-out code?

Currently not.
`name` is originally designed to identify the Transformation. (But seems not necessary at present).
`StatusCode` is designed to identify the status (if it succeeded or failed) after applying transformation. But it seems that `boolean` can already handle it.

mrigger:
Looks good to me (apart from the failing PMD test). @oraluben, any additional comments?

oraluben (approved):
Also LGTM

src/sqlancer/MainOptions.java
```diff
@@ -126,10 +126,19 @@ public class MainOptions {
     @Parameter(names = "--use-reducer", description = "EXPERIMENTAL Attempt to reduce queries using a simple reducer")
     private boolean useReducer = false; // NOPMD
 
+    @Parameter(names = "--reduce-AST", description = "EXPERIMENTAL perform AST reduction after statement reduction")
```
oraluben, line 129: There seems to be a check against the parameter format's failing (see "General tests"), looks like parameters are not expected to contain capital letters: https://github.com/sqlancer/sqlancer/blob/97b8850ed697359d91f77950741e5bd4a0681449/test/sqlancer/TestParameterFormat.java#L20
Maybe we can use `ast` here?

oraluben (changes requested):

src/sqlancer/MainOptions.java
ColinYoungTaro: ok

oraluben (approved):

PR #886, opened 2023-08-26
feat: add support of 8.0 version

ZhengLin-Li:
feat: add support of 8.0 version

In this PR, we add a flag to let user select which version of StoneDB to test.


bajinsheng:
1. CI fails.
2. Only adding a flag? No code generation needs to be changed?

ZhengLin-Li:
1. Yes, the CI fials. not sure the reason. The error message is: `org.opentest4j.AssertionFailedError: --test-8.0-version ==> expected: <true> but was: <false>`. Why it's expected value is true?
2. Basiclly yes, from my previous testing result, because the generation syntax is the same, only expected error message need to be changed or added.

bajinsheng:
1. You can search Google to look for a solution.
2. Really? If so, why initially you show us there are some differences between the two versions?

bajinsheng:
I guess it should be some new features or grammar in version 8.0

ZhengLin-Li:
> I guess it should be some new features or grammar in version 8.0

absolutely yes. I still remember there is a difference in `alter table` statement. I will add supported sytax.

bajinsheng:
Ok. LGTM.

bajinsheng:
@mrigger The PR works for me.

mrigger:
Thanks for the ping. Overlooked that we can merge the PR.

PR #892, opened 2023-09-04
Reducer updates.

oraluben:
This PR:
1. Add information when reproducer (and reducer) is not available. This will help the following reducer implementations;
2. Illustrate how to add reducer to a oracle class (`TiDBTLPWhereOracle`);
3. Add doc about reproducer.

Some minor changes:
1. Bump `actions/setup-java` to v3 to support caching maven artifacts;
2. Remove non-LTS JDK tests. This currently blockes merging because those tests are marked as required, which is no longer necessary IMO. I suggest to remove them from required tests.


src/sqlancer/Main.java
```diff
@@ -449,9 +450,11 @@ public void run() throws Exception {
                 if (options.reduceAST() && !options.useReducer()) {
                     throw new AssertionError("To reduce AST, use-reducer option must be enabled first");
                 }
-                if (reproducer != null && options.useReducer()) {
-                    System.out.println("EXPERIMENTAL: Trying to reduce queries using a simple reducer.");
-                    // System.out.println("Reduced query will be output to stdout but not logs.");
+                if (options.useReducer()) {
+                    if (reproducer == null) {
+                        logger.getQueryPlanFileWriter().write("current oracle do not support experimental reducer.");
```
mrigger, line 455: I think this should be "does not support" since the message is about a specific oracle. Other than that, the PR looks good to me.

mrigger (approved):
LGTM! @bajinsheng, could you perhaps have a quick look at the TiDB reducer?

src/sqlancer/Main.java
```diff
@@ -449,9 +449,11 @@ public void run() throws Exception {
                 if (options.reduceAST() && !options.useReducer()) {
                     throw new AssertionError("To reduce AST, use-reducer option must be enabled first");
                 }
-                if (reproducer != null && options.useReducer()) {
-                    System.out.println("EXPERIMENTAL: Trying to reduce queries using a simple reducer.");
-                    // System.out.println("Reduced query will be output to stdout but not logs.");
+                if (options.useReducer()) {
+                    if (reproducer == null) {
+                        logger.getQueryPlanFileWriter().write("current oracle does not support experimental reducer.");
```
bajinsheng, line 454: Why write this error information into the query plan logging file?

src/sqlancer/tidb/oracle/TiDBTLPWhereOracle.java
```diff
@@ -42,11 +45,29 @@ public void check() throws SQLException {
                 thirdQueryString, combinedString, !orderBy, state, errors);
         ComparatorHelper.assumeResultSetsAreEqual(resultSet, secondResultSet, originalQueryString, combinedString,
                 state);
+        reproducer = globalState -> {
```
bajinsheng, line 48: Can you give more documents about how to support reproducers for a new DBMS? For this code, I am not very sure how it works, and it seems different from the reproducer in the NoREC oracle.

bajinsheng: Can you change it to a unified format as the reproducer in NoREC?

mrigger:
Let me know when/if I can merge the PR. On vacation next week.

oraluben:
I think it's good to go, let me know if I've answered all your concerns @bajinsheng

bajinsheng:
@oraluben Please see my comment above. I am just wondering can you make the coding style consistent?

mrigger:
Hmm, should we merge the code to have the extra functionality? Perhaps we can refactor the other code to use a lambda as well as part of a subsequent PR.

oraluben:
> I am just wondering can you make the coding style consistent?

Do you mean use lambda v.s. define new class for each reducer? I think there's no fundamental difference between them as reducer implementation, and can be chosen accordingly to type of reproducer.

bajinsheng:
@oraluben These are definitely semantically equivalent implements, but it is not a good coding style.

@mrigger  I am fine to merge it since the PR has been open for a quite long time.

oraluben:
I think we can merge it first, and unify the style later.

mrigger:
Currently, the style checks are failing. Could you please execute `mvn formatter:format`?

mrigger:
Checkstyle is failing (see `mxn verify -DskipTests=true`). After fixing this, we can merge the PR. Thanks!

mrigger:
Sorry, one last thing: I noticed that you accidentally checked in some binary files (in the `databases` subdirectory). Could you remove those?

oraluben:
> Sorry, one last thing: I noticed that you accidentally checked in some binary files (in the `databases` subdirectory). Could you remove those?

Oops, sorry for that, now it should be ok

Issue #895, opened 2023-10-18
CERT support/implementation for PostgreSQL


destrex271:
**Proposal**: To implement the CERT testing technique presented in the following research paper: [https://arxiv.org/abs/2306.00355](https://arxiv.org/abs/2306.00355)  for PostgreSQL as well alongside the existing implementations for MySQL, Cockroach DB and TiDB.

destrex271:
Can this issue be assigned to me? I have started the work on a patch for this feature.

bajinsheng:
Sure. That would be great! Looking forward to your PRs!

PR #896, opened 2023-10-21
CERT Implementation for PostgreSQL 

destrex271:
This PR references the Issue #895 . This is an initial PR, would be great if someone can review and validate its working. 
@bajinsheng can you please verify the expected behavior? Just want to cross check if the tests are working as expected.


destrex271:
Came across some bugs, will update the pr once fixed.

.settings/org.eclipse.jdt.core.prefs
```diff
@@ -111,4 +111,4 @@ org.eclipse.jdt.core.compiler.problem.unusedTypeParameter=ignore
 org.eclipse.jdt.core.compiler.problem.unusedWarningToken=warning
 org.eclipse.jdt.core.compiler.problem.varargsArgumentNeedCast=warning
 org.eclipse.jdt.core.compiler.release=disabled
-org.eclipse.jdt.core.compiler.source=1.8
+org.eclipse.jdt.core.compiler.source=1.8
```
bajinsheng, line 114: whitespace?

src/sqlancer/postgres/oracle/PostgresCERT.java
```diff
@@ -0,0 +1,197 @@
...
+            return false;
+        } else {
+            PostgresExpression newWhere = new PostgresBinaryLogicalOperation(select.getWhereClause(),
+                    gen.generateExpression(0), PostgresBinaryLogicalOperation.BinaryLogicalOperator.OR);
+            select.setWhereClause(newWhere);
+            return true;
+        }
+    }
+
+    private int getRow(SQLGlobalState<?, ?> globalState, String selectStr, List<String> queryPlanSequences)
```
bajinsheng, line 162: This function should retrieve the cardinality estimation from the query plans of PostgreSQL, and is not exactly the same the one as MySQL.

You can check here: https://www.postgresql.org/docs/current/using-explain.html to know the query plan formats, and redesign this function to extract the correct cardinality.

bajinsheng (changes requested):
Thanks for your PR! The overall workflow is fine, but the code still requires some effort to make the code work.  `PostgresCERTOracle.java` is similar to `MySQLCERTOracle.java`, but should not be the exact same. 

Additionally, the file name should be `PostgresCERTOracle.java`.
Before pushing PRs, can you run `mvn formatter:format` and `mvn verify -DskipTests=true`?

src/sqlancer/postgres/oracle/PostgresCERT.java
destrex271: Thanks! 
Was looking for this only. 
I realized it latter that Postgres implementation will be a little different than the MySQL one. Will update the PR as I fix these issues.

destrex271: I did come across some problematic queries that were being generated by the [PostgresExpressionGenerator](https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/postgres/gen/PostgresExpressionGenerator.java) with errors mostly related to default function formats(eg RPAD) and errors related to data types.
Should I follow a similar approach that was taken in MySQL by adding the expected errors in  [MySQLErrors.java](https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/mysql/MySQLErrors.java)?

destrex271:
Added the following
  - Rows extracted from top level operation
  - Query plan divided into steps for similarity checks.

What is left
  - Mutating JOIN

Please review. Meanwhile I'll keep looking for possible bugs

src/sqlancer/postgres/oracle/PostgresCERTOracle.java
```diff
@@ -0,0 +1,216 @@
...
+            }
+        }
+
+        // Get row count
+        SQLQueryAdapter q = new SQLQueryAdapter(explainQuery, errors);
+        try (SQLancerResultSet rs = q.executeAndGet(globalState)) {
+            if (rs != null) {
+                while (rs.next()) {
+                    String content = rs.getString(1).trim();
+                    if (content.contains("rows")) {
```
bajinsheng, line 189: "row="

.settings/org.eclipse.jdt.core.prefs
bajinsheng: this modification is not necessary

src/sqlancer/postgres/PostgresErrors.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.regex.Pattern;
+
+import sqlancer.common.query.ExpectedErrors;
+
+public final class PostgresErrors {
+
+    private PostgresErrors() {
+    }
+
+    public static void addExpressionErrors(ExpectedErrors errors) {
```
bajinsheng, line 12: This file is not necessary. You can consider adding the expected errors in `PostgresCommon.java` or the `PostgreCERTOracle.java`.

src/sqlancer/postgres/PostgresErrors.java
```diff
@@ -0,0 +1,16 @@
...
+        errors.addRegex(Pattern.compile("ERROR: *"));
```
bajinsheng, line 13: "ERROR: *" is too general and would hide unexpected errors.

src/sqlancer/postgres/oracle/PostgresCERTOracle.java
```diff
@@ -0,0 +1,216 @@
...
+import sqlancer.postgres.PostgresSchema.PostgresDataType;
+import sqlancer.postgres.PostgresSchema.PostgresTables;
+import sqlancer.postgres.PostgresVisitor;
+import sqlancer.postgres.ast.PostgresBinaryLogicalOperation;
+import sqlancer.postgres.ast.PostgresBinaryLogicalOperation.BinaryLogicalOperator;
+import sqlancer.postgres.ast.PostgresColumnReference;
+import sqlancer.postgres.ast.PostgresExpression;
+import sqlancer.postgres.ast.PostgresSelect;
+import sqlancer.postgres.ast.PostgresTableReference;
+// import sqlancer.postgres.ast.PostgresBinaryArithmeticOperation.PostgresBinaryOperator;
```
bajinsheng, line 28: Please remove unused code

bajinsheng (commented):
Looks good now. Will check again after you support JOIN and LIMIT.

src/sqlancer/postgres/PostgresErrors.java
destrex271: Ok!

src/sqlancer/postgres/oracle/PostgresCERTOracle.java
destrex271:
```                                     QUERY PLAN
-------------------------------------------------------------------​------------------
 Limit  (cost=0.29..14.48 rows=2 width=244)
   ->  Index Scan using tenk1_unique2 on tenk1  (cost=0.29..71.27 rows=10 width=244)
         Index Cond: (unique2 > 9000)
         Filter: (unique1 < 100)
```
According to these query plans I think going with "rows=" || "row=" would be better.

bajinsheng: Sorry, my typo. You can just use "rows=" I think

destrex271:
> Looks good now. Will check again after you support JOIN and LIMIT.

I have worked on the LIMIT mutation but in almost all the cases the tests are throwing inconsistent results like the following:

```
Inconsistent result for query: EXPLAIN SELECT ALL t1.c0 FROM t3, t1 WHERE FALSE; --0
--EXPLAIN SELECT ALL t1.c0 FROM t3, t1 WHERE FALSE LIMIT 947; --1
```

Also to implement the mutations for Join we require a Join implementation similar to TiDBJoin with two table references instead of one, can I modify the PostgresJoin file according to the requirements by adding new constructors and functions or should I keep it completely separate?

bajinsheng:
> In almost all the mutations with limit, the query plan returns 1 if the result without limit is 0. I think this can be a default behaviour of somewhat for LIMIT.

For this case, I think we can ignore LIMIT for PostgreSQL.

> can I modify the PostgresJoin file according to the requirements by adding new constructors and functions or should I keep it completely separate?

We can modify the PostgreJoin file to make the PostgreSQL implementation consistent with other implementations.

destrex271:
> > In almost all the mutations with limit, the query plan returns 1 if the result without limit is 0. I think this can be a default behaviour of somewhat for LIMIT.
> 
> For this case, I think we can ignore LIMIT for PostgreSQL.
> 
> > can I modify the PostgresJoin file according to the requirements by adding new constructors and functions or should I keep it completely separate?
> 
> We can modify the PostgreJoin file to make the PostgreSQL implementation consistent with other implementations.

I also noticed that in some queries the estimated rows are 0, these queries are the ones that usually throw errors so how about we skip those queries altogether?

bajinsheng:
If a query throws an error, it should return nothing? Can you give an example? Typically, if a query throws an expected error for some known reasons, we can filter out these queries by https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/postgres/oracle/PostgresNoRECOracle.java#L46

destrex271:
Sorry my bad, I meant those queries are mostly flagged as Inconsistent by our logic.

bajinsheng:
If it is detected by our logic, it implies a potential bug and you should output it. Like this: https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/tidb/oracle/TiDBCERTOracle.java#L104

destrex271:
> If it is detected by our logic, it implies a potential bug and you should output it. Like this: https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/tidb/oracle/TiDBCERTOracle.java#L104

Ok! will do

destrex271:
I have tried to implement a version of Join mutation but I am constantly running into some bugs because of which the queries being generated do not have the correct syntax. Most of the syntax errors are related to the absence of required tables after FROM or missing clauses for WHERE and GROUP BY.  Has something similar been encountered before that I can refer to pinpoint the actual problem?

destrex271:
Added JOIN mutations to the tests, please review.

bajinsheng:
Can you check the CI? All failed

destrex271:
> Can you check the CI? All failed

I checked the logs and there are some warnings that are being considered as errors. However, I am a little confused since the problematic files in the CI were not modified in this PR and I was also able to build it successfully using `mvn verify -DskipTests=true` locally. I have restored the .settings folder content for now,  maybe that can help with the warnings.

src/sqlancer/postgres/PostgresErrors.java
bajinsheng: As commented, can you remove this file?

destrex271: Done.

bajinsheng (approved):
Thanks for your contribution! Looks good to me!  @mrigger I think we can merge it if there is no more comment from you.

destrex271:
> Thanks for your contribution! Looks good to me!  @mrigger I think we can merge it if there is no more comment from you.

Thanks for helping me out with the codebase! Happy to contribute.

src/sqlancer/postgres/oracle/PostgresCERTOracle.java
```diff
@@ -0,0 +1,266 @@
...
+            try {
+                globalState.getLogger().getCurrentFileWriter().flush();
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+
+        // Get row count
+        SQLQueryAdapter q = new SQLQueryAdapter(explainQuery, errors);
+        try (SQLancerResultSet rs = q.executeAndGet(globalState)) {
```
mrigger, line 235: It would be great to subsequently factor out some of this functionality and write some tests so that debugging any breaking changes in the query plan format are easier to debug.

mrigger (approved):
Thanks a lot for your contribution!

src/sqlancer/postgres/oracle/PostgresCERTOracle.java
destrex271: Ok, I'll look into this.

PR #901, opened 2023-12-05
Fix some bug in reducer

oraluben:
cc @suyuz

### DIFF
--- a/src/sqlancer/Main.java
+++ b/src/sqlancer/Main.java
@@ -204,7 +204,7 @@ public final class Main {
             }
             if (reduceFileWriter == null) {
                 try {
-                    reduceFileWriter = new FileWriter(reduceFile, false);
+                    reduceFileWriter = new FileWriter(reduceFile, true);
                 } catch (IOException e) {
                     throw new AssertionError(e);
                 }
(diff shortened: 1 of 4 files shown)


src/sqlancer/Main.java
```diff
@@ -204,7 +204,7 @@ public FileWriter getReduceFileWriter() {
             }
             if (reduceFileWriter == null) {
                 try {
-                    reduceFileWriter = new FileWriter(reduceFile, false);
+                    reduceFileWriter = new FileWriter(reduceFile, true);
```
suyZhong, line 207: I suppose we'd better use false here. Otherwise, the reduce log would keep increasing.

suyZhong: The actual issue is at L495-497. I think the logger there is not for the `newGlobalState`

suyZhong (changes requested):

suyZhong:
Just a few suggestions,

1.  We could replace the `currentFileWriter` when reducing since we don't need to write to `-cur.log` anymore
2.  Currently the exception log （`database.log`） only shows DDL&DML statements after reducing. I hope except for the general "check reduce.log", I could also see the reduced query in it. We could somehow keep the error messages (Assertion when finding a bug) in the reproducer and add this message at here: https://github.com/sqlancer/sqlancer/blob/201fd4c24aba5378447dc6b521e86d3b85ee485f/src/sqlancer/Main.java#L501

PR #902, opened 2023-12-08
Upgrade CockroachDB to v23.2.0-beta.1

mgartner:
I've updated the CockroachDB tests to use v23.2.0-beta.1.


mrigger (approved):
Thanks a lot! Sorry for noticing the PR only now.

PR #908, opened 2024-01-16
Make --test-collations=false more robust.

### LINKED ISSUE #907: Collation errors for postgres clusters created with different locale
When changing my test machine, I mistakenly created a postgres cluster with `SQL_ASCII` encoding (C). This caused SQLancer to throw a lot of errors around collation.

```
$ java -jar sqlancer-*.jar --username ubuntu postgres 2>&1 | grep "^Caused by" | cut --bytes=-95 | sort | uniq -c | sort -nr
     18 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_AT.utf8" for encoding "SQL_A
     15 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_IT.utf8" for encoding "SQL_A
     13 Caused by: org.postgresql.util.PSQLException: ERROR: collation "en_US.utf8" for encoding "SQL_A
     13 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_BE.utf8" for encoding "SQL_A
     12 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_DE.utf8" for encoding "SQL_A
     11 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_LI.utf8" for encoding "SQL_A
     10 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_LU.utf8" for encoding "SQL_A
      8 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_CH.utf8" for encoding "SQL_A
```

However, it was interesting to note that disabling collations only reduce the error, and not completely make it go away.

```
$ java -jar sqlancer-*.jar --username ubuntu postgres --test-collations=false 2>&1 | grep "^Caused by" | cut --bytes=-95 | sort | uniq -c | sort -nr
     67 Caused by: org.postgresql.util.PSQLException: ERROR: invalid byte sequence for encoding "UTF8":
      7 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_IT.utf8" for encoding "SQL_A
      5 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_BE.utf8" for encoding "SQL_A
      4 Caused by: org.postgresql.util.PSQLException: ERROR: collation "en_US.utf8" for encoding "SQL_A
      4 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_LI.utf8" for encoding "SQL_A
      4 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_CH.utf8" for encoding "SQL_A
      3 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_LU.utf8" for encoding "SQL_A
      3 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_DE.utf8" for encoding "SQL_A
      2 Caused by: org.postgresql.util.PSQLException: ERROR: collation "de_AT.utf8" for encoding "SQL_A
```

robins:
With `--test-collations=false`, database creation would now enforce using `UTF8` and `template0` database. 

Fixes #907 .

This ensures that when run with `--test-collations=false` on postgres clusters created in other locales, SQLancer still tries to create UTF8 encoded databases and test other aspects of the database engine.

No change made to scenarios where --test-collations=true (default).

`mvn verify` passes. 

With patch, stops throwing collation errors:
```
$ cd $lancerpth && java -jar sqlancer-*.jar --username ubuntu postgres --test-collations=false
[2024/01/16 01:23:49] Executed 13705 queries (2740 queries/s; 3.20/s dbs, successful statements: 42%). Threads shut down: 0.
[2024/01/16 01:23:54] Executed 37912 queries (4843 queries/s; 0.00/s dbs, successful statements: 41%). Threads shut down: 0.
[2024/01/16 01:23:59] Executed 60715 queries (4561 queries/s; 0.00/s dbs, successful statements: 41%). Threads shut down: 0.
[2024/01/16 01:24:05] Executed 82813 queries (4385 queries/s; 0.00/s dbs, successful statements: 40%). Threads shut down: 0.
[2024/01/16 01:24:09] Executed 108197 queries (5116 queries/s; 0.00/s dbs, successful statements: 40%). Threads shut down: 0.
^C
```

### DIFF
--- a/src/sqlancer/postgres/PostgresProvider.java
+++ b/src/sqlancer/postgres/PostgresProvider.java
@@ -325,17 +325,24 @@ public class PostgresProvider extends SQLProviderAdapter<PostgresGlobalState, Po
     private String getCreateDatabaseCommand(PostgresGlobalState state) {
         StringBuilder sb = new StringBuilder();
         sb.append("CREATE DATABASE " + databaseName + " ");
-        if (Randomly.getBoolean() && ((PostgresOptions) state.getDbmsSpecificOptions()).testCollations) {
+        if (((PostgresOptions) state.getDbmsSpecificOptions()).testCollations) {
             if (Randomly.getBoolean()) {
-                sb.append("WITH ENCODING '");
-                sb.append(Randomly.fromOptions("utf8"));
-                sb.append("' ");
-            }
-            for (String lc : Arrays.asList("LC_COLLATE", "LC_CTYPE")) {
-                if (!state.getCollates().isEmpty() && Randomly.getBoolean()) {
-                    sb.append(String.format(" %s = '%s'", lc, Randomly.fromList(state.getCollates())));
+                if (Randomly.getBoolean()) {
+                    sb.append("WITH ENCODING '");
+                    sb.append(Randomly.fromOptions("utf8"));
+                    sb.append("' ");
                 }
+                for (String lc : Arrays.asList("LC_COLLATE", "LC_CTYPE")) {
+                    if (!state.getCollates().isEmpty() && Randomly.getBoolean()) {
+                        sb.append(String.format(" %s = '%s'", lc, Randomly.fromList(state.getCollates())));
+                    }
+                }
+                sb.append(" TEMPLATE template0");
             }
+        } else {
+            sb.append("WITH ENCODING '");
+            sb.append(Randomly.fromOptions("utf8"));
+            sb.append("' ");
             sb.append(" TEMPLATE template0");
         }
         return sb.toString();
--- a/src/sqlancer/postgres/PostgresProvider.java
+++ b/src/sqlancer/postgres/PostgresProvider.java
@@ -340,10 +340,7 @@ public class PostgresProvider extends SQLProviderAdapter<PostgresGlobalState, Po
                 sb.append(" TEMPLATE template0");
             }
         } else {
-            sb.append("WITH ENCODING '");
-            sb.append(Randomly.fromOptions("utf8"));
-            sb.append("' ");
-            sb.append(" TEMPLATE template0");
+            sb.append("WITH ENCODING 'UTF8' TEMPLATE template0");
         }
         return sb.toString();
     }


src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -325,17 +325,24 @@ protected void prepareTables(PostgresGlobalState globalState) throws Exception {
...
+                for (String lc : Arrays.asList("LC_COLLATE", "LC_CTYPE")) {
+                    if (!state.getCollates().isEmpty() && Randomly.getBoolean()) {
+                        sb.append(String.format(" %s = '%s'", lc, Randomly.fromList(state.getCollates())));
+                    }
                 }
+                sb.append(" TEMPLATE template0");
             }
+        } else {
+            sb.append("WITH ENCODING '");
+            sb.append(Randomly.fromOptions("utf8"));
```
mrigger, line 344: Here, you can directly use "utf8" rather than `Randomly.fromOptions("utf8")`.

src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -325,17 +325,24 @@ protected void prepareTables(PostgresGlobalState globalState) throws Exception {
     private String getCreateDatabaseCommand(PostgresGlobalState state) {
         StringBuilder sb = new StringBuilder();
         sb.append("CREATE DATABASE " + databaseName + " ");
-        if (Randomly.getBoolean() && ((PostgresOptions) state.getDbmsSpecificOptions()).testCollations) {
```
mrigger, line 328: Does the omission of `Randomly.getBoolean()` change the behavior of this code? The original idea of this was that only with some chance will we create collation-related things, even if the collation-option is enabled (i.e., the option allows creating collation settings, but it doesn't mandate them).

mrigger (commented):
Thanks a lot for the PR!

src/sqlancer/postgres/PostgresProvider.java
robins: Correct - For the default setting (`--test-collations=true`) that behavior should not change.

The way I ensured current default behavior is that if `--test-collation` flag is enabled, there would be a `Randomly.getBoolean()` to check if we're setting the collation options... and within that `if`, there's a second `Randomly.getBoolean()` to check if we're also doing `WITH ENCODING`. IIUC, this is how things were before this patch.

src/sqlancer/postgres/PostgresProvider.java
robins: Done. I was initially tempted to conflate the options, but saw another place use `fromOptions("utf8")` and erred to did that instead.

Anyway, have pushed another commit with conflated options.

mrigger (approved):

robins:
Interesting to see quite unrelated engines fail. This is my first PR and now reading `CONTRIBUTING.md` to see that `mvn verify` was not enough. Would retry `mvn test` and resubmit.

PR #914, opened 2024-03-05
[WIP] Generic TLP oracle

malwaregarry:
This PR introduces a new common implementation for test oracles that can be used by all DBMS. 
This is done by decoupling test oracles and expression generators. 

This is part of the ongoing effort to reduce code duplication in SQLancer.

### DIFF
--- a/src/sqlancer/ComparatorHelper.java
+++ b/src/sqlancer/ComparatorHelper.java
@@ -195,14 +195,27 @@ public final class ComparatorHelper {
     }
 
     public static String runQuery(String queryString, ExpectedErrors errors, GlobalState<?, ?, SQLConnection> state) {
+        if (state.getOptions().logEachSelect()) {
+            // TODO: refactor me
+            state.getLogger().writeCurrent(queryString);
+            try {
+                state.getLogger().getCurrentFileWriter().flush();
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+
         SQLQueryAdapter query = new SQLQueryAdapter(queryString, errors);
         try (SQLancerResultSet result = query.executeAndGet(state)) {
             if (result == null) {
                 throw new IgnoreMeException();
             }
+            if (!result.next()) {
+                return null;
+            }
             return result.getString(1);
-        } catch (Exception e) {
-            throw new IgnoreMeException();
+        } catch (SQLException e) {
+            throw new AssertionError(queryString, e);
         }
     }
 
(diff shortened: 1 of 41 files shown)


src/sqlancer/sqlite3/SQLite3Errors.java
```diff
@@ -16,57 +17,45 @@ public static void addDeleteErrors(ExpectedErrors errors) {
         errors.add("ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint");
     }
 
+    public static List<String> getExpectedExpressionErrors() {
```
mrigger, line 20: For the `getExpectedExpressionErrors` refactorings, do you think we can extract this to a separate PR, which we could quickly merge?

mrigger: Btw, `List.of` creates an immutable list. Do we want this? I'd expect that at some locations in the code, we might get an initial list of errors and then expand it (e.g., first some expression errors and than some errors for `GROUP BY` classes if we use this).

mrigger: Given that you've started refactoring this class. Perhaps we could also change the interface of these utility classes to return `List<ExpectedErrors>`? Just an idea for a potential follow-up PR.

malwaregarry: > Given that you've started refactoring this class. Perhaps we could also change the interface of these utility classes to return List<ExpectedErrors>? Just an idea for a potential follow-up PR.


I don't think this would work because ExpectedErrors is a collection of errors, ie. it contains a set and a list.

malwaregarry: > Btw, List.of creates an immutable list. Do we want this? I'd expect that at some locations in the code, we might get an initial list of errors and then expand it (e.g., first some expression errors and than some errors for GROUP BY classes if we use this).

My idea was to keep them immutable so that it becomes easier to reason about. 

If we need  to have more errors we could always concat the lists together. Though the syntax would be more verbose which is a downside.

malwaregarry: I've created a new pr for the errors #922

mrigger: > Edit: We can return ExpectedErrors, but I chose not to do this so that it is easier to compose errors

I think we could subsequently also have a way to do so in the `ExpectedErrors` class. A list concatenation might, for example, not be optimal, as we might have duplicate error  messages and we could subsequently also optimize things (e.g., remove error messages that match a regex error template).

> My idea was to keep them immutable so that it becomes easier to reason about.

Sounds good to me, assuming that it doesn't break the current code.

> I've created a new pr for the errors https://github.com/sqlancer/sqlancer/pull/922

Sounds good. Let me know once I should review it - I assume we can merge it quite quickly since it's not a major change.

Issue #24, opened 2020-06-07
Add a testing implementation for TimescaleDB


mrigger:
Mike Freedman, CTO and Co-founder of Timescale, [expressed interest in having TimescaleDB tested](https://twitter.com/michaelfreedman/status/1264752236644761605):

> Hey Manuel: would welcome any testing to @TimescaleDB , which is implemented as extension to Postgres and thus probably already fits your model.

The implementation could probably be based on top of SQLancer's PostgreSQL implementation.

Ethan-DB:
What is it different from PostgreSQL? Whether it can share the same set of tests with Postgres? And add the `time travel` function on this Postgres test. @mrigger

GoswamiTushar:
Hi, can I work on this issue?

amangupta201:
Hey could you please assign me this issue?

PR #916, opened 2024-03-06
Remove PQS

malwaregarry:
Removed PQS test oracles as they are currently unmaintained and different from the other test oracles

Also removed TimeScaleDB as it was only making use of the PQS test oracle (might need to re-open #24)


mrigger:
@def-, is PQS actively being used within Materialize? We are considering removing it from the codebase as part of efforts to improve SQLancer's architecture and codebase in general. If it's being used, we should not remove it (at least for Materialize).

def-:
Yes, we use it in Materialize.

malwaregarry:
ok sure, i can add back materialize

mrigger:
Hmm, I wonder if we should perhaps just leave PQS in the source base. It might be that also other companies are still using it, and I guess we cannot simplify the architecture as soon as we have one implementation that uses PQS. What do you think, @malwaregarry?

malwaregarry:
yup that makes sense. In that case,  instead of consolidating an oracle for PQS, we can move most of the logic into the generator and the oracle just executes and compares. How does this sound , @mrigger?

PR #917, opened 2024-03-12
Update SQLite3 tests

malwaregarry:
Originally, the test did not include query partitioning oracles as the default was NoREC.

Also, `--num-queries` was previously set to 0, and is now changed to what is specified in `TestConfig`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,12 +6,14 @@ on:
       - main
     paths:
       - 'src/**'
+      - 'test/**'
       - '.github/workflows/*.yml'
   pull_request:
     branches:
       - main
     paths:
       - 'src/**'
+      - 'test/**'
       - '.github/workflows/*.yml'
 
 jobs:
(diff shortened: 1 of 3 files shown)


mrigger (approved):

malwaregarry:
yup, i ran both the tests for the full 5min without any errors. The problem with checks not running should be because they only run when something in `src/` is updated and this PR only edits `test/`. Is there any way to trigger it manually? If not should I update the workflow file to include `test/`? @mrigger

malwaregarry:
i ran the both tests for 5 minutes and didn't get any errors

mrigger:
Could you check whether you can reproduce this issue on the latest development build of SQLite (i.e., building it yourself)? If so, this likely is a bug in SQLite.

malwaregarry:
ok i figured out what is causing this errror. It's  not a bug in sqlite.

It's because the first test deleted the db file while the second test is still using it. This could be why all other tests run are separated into their own class

mrigger:
I see. Thanks for investigating!

malwaregarry:
@mrigger i think you closed this PR instead of merging it.

Issue #918, opened 2024-03-14
DQP Oracle


bajinsheng:
We will support a new oracle Differential Query Plans (DQP), introduced in the SIGMOD'24 paper: `Keep It Simple: Testing Databases via Differential Query Plans`.

bajinsheng:
Quick Start:
```shell
java -jar sqlancer-2.0.0.jar --num-threads 1 mariadb --oracle DQP
```

```shell
java -jar sqlancer-2.0.0.jar --num-threads 1 tidb --oracle DQP
```


```shell
java -jar sqlancer-2.0.0.jar --num-threads 1 mysql --oracle DQP
```

PR #919, opened 2024-03-14
Support TiFlash for TiDB

bajinsheng:
Part of #918


mrigger (approved):

mrigger:
It seems PMD is failing (see `mvn -B verify -DskipTests=true`).

PR #921, opened 2024-03-18
MySQL Join implementation

bajinsheng:
Part of #918

### DIFF
--- a/src/sqlancer/mysql/MySQLProvider.java
+++ b/src/sqlancer/mysql/MySQLProvider.java
@@ -63,11 +63,6 @@ public class MySQLProvider extends SQLProviderAdapter<MySQLGlobalState, MySQLOpt
         SELECT_INFO((g) -> new SQLQueryAdapter(
                 "select TABLE_NAME, ENGINE from information_schema.TABLES where table_schema = '" + g.getDatabaseName()
                         + "'")), //
-        CREATE_TABLE((g) -> {
-            // TODO refactor
-            String tableName = DBMSCommon.createTableName(g.getSchema().getDatabaseTables().size());
-            return MySQLTableGenerator.generate(g, tableName);
-        }), //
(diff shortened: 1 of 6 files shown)


src/sqlancer/mysql/MySQLProvider.java
```diff
@@ -148,7 +140,7 @@ private static int mapActions(MySQLGlobalState globalState, Action a) {
 
     @Override
     public void generateDatabase(MySQLGlobalState globalState) throws Exception {
-        while (globalState.getSchema().getDatabaseTables().size() < Randomly.smallNumber() + 1) {
+        while (globalState.getSchema().getDatabaseTables().size() < Randomly.getNotCachedInteger(1, 6)) {
```
mrigger, line 143: 6 tables seem like a lot here and might have a detrimental effect on performance. Could we perhaps set the limit depending on the test oracle here?

src/sqlancer/mysql/MySQLProvider.java
```diff
@@ -148,7 +140,7 @@ private static int mapActions(MySQLGlobalState globalState, Action a) {
 
     @Override
     public void generateDatabase(MySQLGlobalState globalState) throws Exception {
-        while (globalState.getSchema().getDatabaseTables().size() < Randomly.smallNumber() + 1) {
+        while (globalState.getSchema().getDatabaseTables().size() < Randomly.getNotCachedInteger(1, 2)) {
```
mrigger, line 143: I think we could just leave `Randomly.smallNumber() + 1`, which might sometimes generate more tables than 2?

bajinsheng: No. Small number returns (-1, 1) with a Gaussian distribution, which means in most situations, there is only one table.

mrigger: Thanks for the clarification!

PR #922, opened 2024-03-18
Update expected errors

malwaregarry:
This PR simply decouples the definition of error strings and regexes with adding them to  ExpectedErrors.

### DIFF
--- a/src/sqlancer/clickhouse/ClickHouseErrors.java
+++ b/src/sqlancer/clickhouse/ClickHouseErrors.java
@@ -1,5 +1,7 @@
 package sqlancer.clickhouse;
 
+import java.util.List;
+
 import sqlancer.common.query.ExpectedErrors;
 
 public final class ClickHouseErrors {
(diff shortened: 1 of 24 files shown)


mrigger:
Some more thoughts:
* We could also implement a static factory method `ExpectedErrors.of`, similar to `List.of`.
* Since we don't want to have duplicate errors, we should actually use a `Set` rather than a `List` (I see that the `ExpectedErrors` class already does that).

malwaregarry:
@mrigger hmm, there is a static  factory method `from` in ExpectedErrors. Do I stick with that naming or should I rename it?

mrigger:
I think either would work, as you prefer.

src/sqlancer/common/query/ExpectedErrors.java
```diff
@@ -54,14 +56,35 @@ public static ExpectedErrors from(String... errors) {
         return expectedErrors;
     }
 
+    public static ExpectedErrors from(Pattern... errors) {
```
mrigger, line 59: Btw, I think we could also provide convenience classes that make adding regular expressions less tedious (e.g., `fromRegexes(String... errors)` that takes the strings and instantiates `Pattern` objects.

malwaregarry:
I added a new ExpectedErrorBuilder class to handle the creation of new errors. I kept the original functions because I think some dependencies require the fact that ExpectedErrors is mutable. 

> we don't want to have duplicate errors, we should actually use a Set rather than a List (I see that the ExpectedErrors class already does that).

I changed Regexes to use a set as well. 

I think this PR can be merged if you think it's good.

mrigger (approved):

mrigger:
> Edit: just saw the comment. Don't merge yet

Okay, let me know when.

It would btw be nice if we could also subsequently add some unit tests to test such core functionality.

malwaregarry:
@mrigger Clickhouse is now failing. I checked that it doesn't use the methods that were changed, and doesn't use regexes, so the  change to a hashset isn't the problem. And since the commit 3 days back passed the CI, it shouldn't have anything to do with this PR.

I just checked that there was an update to the clickhouse docker image 5 hours ago, so it could be anew bugs.

https://hub.docker.com/r/clickhouse/clickhouse-server/tags

mrigger:
It seems also PostgreSQL started failing, which I think still worked before. Do you think this is also unrelated?

mrigger:
The ClickHouse "java.io.IOException: std::exception. Code: 1001, type: std::bad_variant_access, e.what() = bad_variant_access (version 24.3.1.1823 (official build))" issue seems like a potential bug. @qoega, any input on this?

malwaregarry:
hmm, in that case I think i messed up somewhere. let me run locally.

qoega:
I've seen it our ci recently. It is already fixed!

https://github.com/ClickHouse/ClickHouse/pull/61654

malwaregarry:
The checks pass now. 

Postgres was due to an error on my side. Thanks for spotting it.

src/sqlancer/clickhouse/ClickHouseErrors.java
```diff
@@ -1,85 +1,65 @@
...
+                "is not under aggregate function and not in GROUP BY", "is not under aggregate function",
+                "is violated at row 1. Expression:", // TODO: check constraint on table creation
+                "is violated, because it is a constant expression returning 0. It is most likely an error in table definition",
+                "there are only columns", "there are columns", "in block. (NOT_FOUND_COLUMN_IN_BLOCK)",
+                "Missing columns", "Ambiguous column",
+                "Must be one unsigned integer type. (ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER)",
+                "Floating point partition key is not supported", "Cannot get JOIN keys from JOIN ON section",
+                "ILLEGAL_DIVISION", "DECIMAL_OVERFLOW",
+                "Cannot convert out of range floating point value to integer type",
+                "Unexpected inf or nan to integer conversion", "No such name in Block::erase", // https://github.com/ClickHouse/ClickHouse/issues/42769
```
mrigger, line 43: Just noticed. Some of these are known bugs, for which we should the pattern explained at https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#unfixed-bugs (e.g., see the Citus implementation). I guess it's not in scope of the PR, but if you are interested, @malwaregarry, you could address those (perhaps some of them have also been fixed by now).

src/sqlancer/sqlite3/SQLite3Errors.java
```diff
@@ -9,11 +11,75 @@ public final class SQLite3Errors {
...
+        // fts5 functions
+        errors.add("unable to use function highlight in the requested context");
+        errors.add("no such cursor");
+
+        // INDEXED BY
+        errors.add("no query solution");
+        errors.add("no such index");
+
+        // UNION/INTERSECT ...
+        errors.add("ORDER BY term does not match any column in the result set");
```
mrigger, line 73: For the set operators, window functions, INDEXED by, etc: these are actually not expression errors, so we'd like to add those only if the statement might use them (adding expected errors that are not relevant might otherwise shadow some bugs). Where did you extract those?

malwaregarry: I got them from the `addExpectedExpressionErrors()` function below. Just realised i forgot to update `addExpectedExpressionErrors()` to call this function

mrigger: Okay, I see. After remove this from the old one, we can merge the PR.

src/sqlancer/clickhouse/ClickHouseErrors.java
malwaregarry: thanks for pointing this out. I can work on this after I'm done with NoREC

src/sqlancer/sqlite3/SQLite3Errors.java
malwaregarry: @mrigger I made the change and checks pass.

mrigger (approved):

PR #924, opened 2024-03-19
MariaDB Join implementation

bajinsheng:
Part of #918


src/sqlancer/mariadb/ast/MariaDBConstant.java
```diff
@@ -1,6 +1,10 @@
 package sqlancer.mariadb.ast;
 
-public class MariaDBConstant extends MariaDBExpression {
+public class MariaDBConstant implements MariaDBExpression {
+
+    private MariaDBConstant() {
+        throw new UnsupportedOperationException("This is a utility class and cannot be instantiated");
```
mrigger, line 6: If it's private, you don't need the `throw` here.

mrigger (approved):

PR #928, opened 2024-03-26
[WIP] Generic NoREC oracle

malwaregarry:
Based off the sqlite3  NoREC implementation.

Not implemented for any DBMS yet.
Will extract DBMS specific select and generator implementations from #914 

@mrigger i think this might be easier to review since TLP has 5 sub-oracles while NoREC is just one.

### DIFF
--- a/src/sqlancer/common/ast/newast/Select.java
+++ b/src/sqlancer/common/ast/newast/Select.java
@@ -44,7 +44,5 @@ public interface Select<J extends Join<E, T, C>, E extends Expression<C>, T exte
 
     Expression<C> getHavingClause();
 
-    Expression<C> toSum();
-
     String asString();
 }
--- a/src/sqlancer/common/ast/newast/Select.java
+++ b/src/sqlancer/common/ast/newast/Select.java
@@ -44,5 +44,7 @@ public interface Select<J extends Join<E, T, C>, E extends Expression<C>, T exte
 
     Expression<C> getHavingClause();
 
+    Expression<C> toSum();
+
     String asString();
 }
--- a/src/sqlancer/common/gen/NoRECGenerator.java
+++ b/src/sqlancer/common/gen/NoRECGenerator.java
@@ -1,0 +1,27 @@
+package sqlancer.common.gen;
+
+import java.util.List;
+
+import sqlancer.common.ast.newast.Expression;
+import sqlancer.common.ast.newast.Join;
+import sqlancer.common.ast.newast.Select;
+import sqlancer.common.schema.AbstractTable;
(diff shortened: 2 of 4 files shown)


src/sqlancer/common/oracle/NoRECOracle.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private final G state;
+
+    private NoRECGenerator<J, E, T, C> gen;
+    private final ExpectedErrors errors;
+
+    private String lastQueryString;
+
+    public NoRECOracle(G state, NoRECGenerator<J, E, T, C> gen, ExpectedErrors expectedErrors) {
+        this.state = state;
```
mrigger, line 32: Given that this class will be widely used, perhaps we can implement it quite defensively? For example, here, we could throw an `IllegalArgumentException` if any of the arguments is `null` (I checked for other opportunities below, but didn't find any).

src/sqlancer/common/ast/newast/Select.java
```diff
@@ -44,5 +44,7 @@ public interface Select<J extends Join<E, T, C>, E extends Expression<C>, T exte
 
     Expression<C> getHavingClause();
 
+    Expression<C> toSum();
```
mrigger, line 47: Perhaps a comment would help here to understand the method and its purpose.

src/sqlancer/common/gen/NoRECGenerator.java
```diff
@@ -0,0 +1,27 @@
...
+
+    E generateBooleanExpression();
+
+    Select<J, E, T, C> generateSelect();
+
+    List<J> getRandomJoinClauses();
+
+    List<E> getTableRefs();
+
+    E generateOptimizedFetchColumn(boolean shouldUseAggregate);
```
mrigger, line 24: Also here, a comment would be helpful.

mrigger (approved):

src/sqlancer/common/oracle/NoRECOracle.java
malwaregarry: good idea! I've made the changes

src/sqlancer/common/ast/newast/Select.java
malwaregarry: i removed it after reviewing since it makes more sense for the generator to handle this

malwaregarry:
I was originally going to add sqlite3 generator implementation in this PR but I think it is better to do so in another PR. @mrigger if that's ok with you then this PR is ready for review

mrigger (approved):

PR #929, opened 2024-03-26
[DuckDB] downgrade version and add NoREC test

malwaregarry:
Version downgraded to the last known version that passed the unit test


mrigger (approved):
Thanks! We can merge this one, but it would be more ideal to update DuckDB to the latest version (which typically involves updating the expected errors and removing keywords that are no longer recognized.

PR #931, opened 2024-03-26
Remove the TiDB incremental ANALYZE feature

mrigger:
This feature currently fails the CI: "the incremental analyze feature has already been removed in TiDB v7.5.0, so this will have no effect."


malwaregarry:
I was looking into this.

There are 4 other errors after this change:

1. `Incorrect string value: '18446744073709551615' for function inet_aton` caused by an invalid parameter for inet_aton
2. `Got error 'Empty pattern is invalid' from regexp` caused by regexp applied to an empty string ie. `REGEX('')`
3. `Result of space() was larger than max_allowed_packet (67108864)` caused by space being called with a large number
4. `Column 'c0' has a partitioning function dependency and cannot be dropped or renamed`. Not sure about this one but should be because of the change here https://github.com/pingcap/tidb/issues/38739 

Adding all four to the insert errors fixes the CI. If you like, I can create a new PR which does so and pin the TIDB version to the current latest version v7.5.1

mrigger:
>Adding all four to the insert errors fixes the CI. If you like, I can create a new PR which does so and pin the TIDB version to the current latest version v7.5.1

That would be great! Do you also want to include the `ANALYZE` case, or should I merge this one?

malwaregarry:
you can merge this. I rebased from this branch

PR #932, opened 2024-03-27
[TIDB] Add expected insert errors for v7.5.1

malwaregarry:
rebased from #931 

Adding the four to the insert errors fixes the CI. 
1. `Incorrect string value: '18446744073709551615' for function inet_aton` caused by an invalid parameter for inet_aton
2. `Got error 'Empty pattern is invalid' from regexp` caused by regexp applied to an empty string ie. `REGEX('')`
3. `Result of space() was larger than max_allowed_packet (67108864)` caused by space being called with a large number
4. `Column 'c0' has a partitioning function dependency and cannot be dropped or renamed`. Not sure about this one but could be because of the change here https://github.com/pingcap/tidb/issues/38739 

Also pinned TiDB version in checks to the current latest (v7.5.1)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -459,7 +459,7 @@ jobs:
       - name: Set up TiDB
         run: |
           docker pull pingcap/tidb:latest
-          docker run --name tidb-server -d -p 4000:4000 pingcap/tidb:latest
+          docker run --name tidb-server -d -p 4000:4000 pingcap/tidb:v7.5.1
           sleep 10
       - name: Create SQLancer user
         run: sudo mysql -h 127.0.0.1 -P 4000 -u root -D test -e "CREATE USER 'sqlancer'@'%' IDENTIFIED WITH mysql_native_password BY 'sqlancer'; GRANT ALL PRIVILEGES ON *.* TO 'sqlancer'@'%' WITH GRANT OPTION; FLUSH PRIVILEGES;"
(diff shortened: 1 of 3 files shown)


mrigger:
It seems there is one remaining error that needs to be added:
```
Caused by: java.sql.SQLException: Column 'c0' has a partitioning function dependency and cannot be dropped or renamed
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:129)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
	at com.mysql.cj.jdbc.StatementImpl.executeInternal(StatementImpl.java:763)
	at com.mysql.cj.jdbc.StatementImpl.execute(StatementImpl.java:648)
	at sqlancer.common.query.SQLQueryAdapter.execute(SQLQueryAdapter.java:100)
	... 13 more
--java.lang.AssertionError: ALTER TABLE t0  DROP c0;
```

The tests also trigger a bug in TiDB. We can merge the PR without addressing this, but it would be nice to (potentially later) report it and prevent the bug-inducing pattern from being generated (see https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#unfixed-bugs).

malwaregarry:
> It seems there is one remaining error that needs to be added
Not sure why I wasn't  getting this error earlier. I ran again and discovered the bug, and also error 3 happening for delete statements as well. I moved the errors to the corresponding  locations in generation and the CI passes now.

> The tests also trigger a bug in TiDB. We can merge the PR without addressing this, but it would be nice to (potentially later) report it and prevent the bug-inducing pattern from being generated (see https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#unfixed-bugs).

What bug are you referring to? I don't see any other bugs in CI

malwaregarry:
Thanks, I see it now, didn't check the qpg tests.

mrigger (approved):
Merging this first, since addressing the TiDB bug might require more work.

PR #934, opened 2024-03-28
Add NoREC test for Postgres

malwaregarry:
* Add NoREC test for Postgres

* Split up tests so that there is no need for workaround

### DIFF
--- a/test/sqlancer/dbms/TestPostgresNoREC.java
+++ b/test/sqlancer/dbms/TestPostgresNoREC.java
@@ -7,29 +7,17 @@ import org.junit.jupiter.api.Test;
 
 import sqlancer.Main;
 
-public class TestPostgres {
+public class TestPostgresNoREC {
 
     String postgresAvailable = System.getenv("POSTGRES_AVAILABLE");
     boolean postgresIsAvailable = postgresAvailable != null && postgresAvailable.equalsIgnoreCase("true");
 
     @Test
-    public void testPostgres() {
+    public void testNoREC() {
         assumeTrue(postgresIsAvailable);
         assertEquals(0,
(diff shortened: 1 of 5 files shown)


test/sqlancer/dbms/TestPostgresNoREC.java
```diff
@@ -0,0 +1,23 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+import org.junit.jupiter.api.Test;
+
+import sqlancer.Main;
+
+public class TestPostgresNoREC {
+
+    String postgresAvailable = System.getenv("POSTGRES_AVAILABLE");
```
mrigger, line 12: Do you think it might be worthwhile to create a PostgreSQL test base class in which we can move the `postgresIsAvailable` logic?

malwaregarry: maybe instead of just for postgres, we can maybe add a utility function that can be used by all databses into TestConfig?

malwaregarry: i added a commit which shows the change. If this is good then I can also make the change for the other databases, either in this pr or a separate one.

mrigger: Looks good. Do you think we should make "POSTGRES_AVAILABLE" a string constant in `TestConfig`? This would help to avoid any issues if, at some point in time, we rename the constant. Either is fine to me - as you prefer.

malwaregarry: I think that's a good idea. I've made the change.

Also, I think I will make the changes to other databases in another PR

mrigger (approved):
LGTM! Merging after the CI checks have passed.

PR #935, opened 2024-03-29
[Postgres & Citus] Use generic NoREC oracle  

malwaregarry:
Changes
* Implemented expression interfaces for Postgres (not sure why the verification checks required updating Oceanbase and Mariadb with `@Override` for certain functions)
* Changed NoREC generator to use specific `Select` type
* Updated NoREC oracle for Postgres
* Updated NoREC oracle for Citus as it depended on Postgres

Note: The reproducer didn't seem specific to SQLite so I enabled them for Postgres and Citus as well

### DIFF
--- a/src/sqlancer/citus/oracle/CitusNoRECOracle.java
+++ b/src/sqlancer/citus/oracle/CitusNoRECOracle.java
@@ -1,14 +1,49 @@
 package sqlancer.citus.oracle;
 
-import sqlancer.citus.gen.CitusCommon;
-import sqlancer.postgres.PostgresGlobalState;
-import sqlancer.postgres.oracle.PostgresNoRECOracle;
+import java.sql.SQLException;
+import java.util.regex.Pattern;
 
-public class CitusNoRECOracle extends PostgresNoRECOracle {
+import sqlancer.Reproducer;
+import sqlancer.citus.gen.CitusCommon;
+import sqlancer.common.oracle.NoRECOracle;
+import sqlancer.common.oracle.TestOracle;
+import sqlancer.common.query.ExpectedErrors;
+import sqlancer.postgres.PostgresGlobalState;
+import sqlancer.postgres.PostgresSchema;
+import sqlancer.postgres.PostgresSchema.PostgresColumn;
+import sqlancer.postgres.PostgresSchema.PostgresTable;
+import sqlancer.postgres.ast.PostgresExpression;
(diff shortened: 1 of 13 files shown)


src/sqlancer/citus/oracle/CitusNoRECOracle.java
```diff
@@ -1,14 +1,49 @@
...
-public class CitusNoRECOracle extends PostgresNoRECOracle {
+public class CitusNoRECOracle implements TestOracle<PostgresGlobalState> {
+    private final NoRECOracle<PostgresSelect, PostgresJoin, PostgresExpression, PostgresSchema, PostgresTable, PostgresColumn, PostgresGlobalState> oracle;
 
     public CitusNoRECOracle(PostgresGlobalState globalState) {
-        super(globalState);
-        CitusCommon.addCitusErrors(errors);
+        PostgresExpressionGenerator gen = new PostgresExpressionGenerator(globalState);
+        ExpectedErrors errors = ExpectedErrors.newErrors()
+                .with(PostgresCommon.getCommonExpressionErrors().toArray(new String[0]))
```
mrigger, line 27: What about overloading the `with` method to accept a `Collection`? This way, we might save a bit of boilerplate code at various places (I remember seeing this pattern before).

mrigger: Other than that, the change looks good to me.

mrigger: Btw, feel free to click on "Request review" at the top right so I get a notification when I should review something.

malwaregarry: I wanted to change this but I couldn't overload `with` with both `Collection<String>` and `Collection<Pattern>` due to type erasure, which is a little annoying. 

The alternative is to have separate methods `with` and `withRegex`. Do you think this is better?

mrigger: Could you just specify a generic type parameter `<T>` for the method and then a type-unsafe assignment (might need to add a warning suppression to the method)?

mrigger: Anway, merging this one first, as we could do this refactoring as part of a follow-up PR.

mrigger (approved):

PR #936, opened 2024-03-31
Update `with` in `ExpectedErrorsBuilder` to support Collections

malwaregarry:
Used the trick stated in https://stackoverflow.com/a/74411632.

Opted to use this method so that we still have compile-time type checks, but we can still change to the generic method if this is too unorthodox

### DIFF
--- a/src/sqlancer/citus/oracle/CitusNoRECOracle.java
+++ b/src/sqlancer/citus/oracle/CitusNoRECOracle.java
@@ -1,7 +1,6 @@
 package sqlancer.citus.oracle;
 
 import java.sql.SQLException;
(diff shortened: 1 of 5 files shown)


malwaregarry:
@mrigger

mrigger:
Hmm, as a user of these methods, I think I would be confused by this. According to the principle of least surprise, perhaps we can pick a simpler option (e.g., just having a separate `withRegex` method might also work).

malwaregarry:
@mrigger I've made the changes and the checks are done.

No reviewers are showing up in the Reviewer section for some reason.

mrigger (approved):

PR #937, opened 2024-04-01
[Clickhouse] Update expected errors and pin version to latest LTS

malwaregarry:
PR to fix checks for Clickhouse which has been failing due to an updated error message.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,8 +109,8 @@ jobs:
         run: mvn -B package -DskipTests=true
       - name: Set up ClickHouse
         run: |
-          docker pull clickhouse/clickhouse-server:head
-          docker run --ulimit nofile=262144:262144 --name clickhouse-server -p8123:8123 -d clickhouse/clickhouse-server:head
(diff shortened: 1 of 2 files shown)


mrigger:
@qoega, do you think it's okay to pin the version? We are struggling with many failing CI tests, and one way to address this is to pin the version to the latest supported release of each database system.

qoega:
Do you see more ClickHouse failures? Pinning is ok, we test master by ourselves

malwaregarry:
That was the only error I saw

mrigger:
Merging this one. Btw, @qoega, it seems the just-merged PR triggered a potential issue (see https://github.com/sqlancer/sqlancer/actions/runs/8505272902/job/23293479849):

> Code: 10. DB::Exception: Not found column __table1.c0: in block __table2.c0 Int32 Int32(size = 0). (NOT_FOUND_COLUMN_IN_BLOCK) (version 24.4.1.183 (official build))

mrigger (approved):

mrigger:
@qoega, if you think of it, it would be great if you could update the ClickHouse version from time to time (e.g., if we forget to do so).

PR #938, opened 2024-04-01
[MySQL] Add bugs 114533 and 114544

malwaregarry:
The checks for MySQL were detecting an error caused by Java parsing an invalid string. I fixed this by passing the exception up as an `SQLException` and adding a new `ExpectedError`

~~There is still a potential bug reported, which I will investigate after this PR is merged.~~
https://bugs.mysql.com/bug.php?id=114533
https://bugs.mysql.com/bug.php?id=114534

### DIFF
--- a/src/sqlancer/common/query/SQLancerResultSet.java
+++ b/src/sqlancer/common/query/SQLancerResultSet.java
@@ -35,7 +35,11 @@ public class SQLancerResultSet implements Closeable {
     }
 
     public String getString(int i) throws SQLException {
-        return rs.getString(i);
+        try {
+            return rs.getString(i);
+        } catch (NumberFormatException e) {
+            throw new SQLException(e);
+        }
     }
 
     public boolean isClosed() throws SQLException {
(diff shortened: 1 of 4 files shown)


mrigger:
This looks like a bug in the JDBC driver to me. I think it would be ideal to reduce this case, report it to MySQL (if we believe it's indeed a bug), and then add it as a known bug, rather than an expected error.

malwaregarry:
```
DROP DATABASE IF EXISTS database3;
CREATE DATABASE database3;
USE database3;
CREATE TABLE IF NOT EXISTS t0(c0 FLOAT ZEROFILL);
INSERT INTO t0(c0) VALUES("-0");
SELECT t0.c0 FROM t0;  --0000000000-0
```

I reduced the error to this.  Seems to be  because of a column with ZEROFILL and a negative 0.

malwaregarry:
@mrigger the known bugs have been added

mrigger (approved):

mrigger:
It seems there is still another location where the JDBC issue is triggered: https://github.com/sqlancer/sqlancer/actions/runs/8505272902/job/23293481262.

PR #945, opened 2024-05-18
[MariaDB] Update expected errors and pin to version 11.3.2

malwaregarry:
* Added expected errors for new restrictions  introduced
* Prevent a newly discovered bug from being generated 
https://jira.mariadb.org/browse/MDEV-33893

related to #834


mrigger (approved):
LGTM, thanks!

PR #946, opened 2024-05-18
[Databend] Pinned version to v1.2.452

malwaregarry:
* Updated error messages for new restrictions
* New bugs reported:
https://github.com/datafuselabs/databend/issues/15568
https://github.com/datafuselabs/databend/issues/15569
https://github.com/datafuselabs/databend/issues/15570
https://github.com/datafuselabs/databend/issues/15572

There is still the problem where the client disconnects randomly and can't reconnect but I'm not sure how to fix that

### DIFF
--- a/src/sqlancer/databend/DatabendErrors.java
+++ b/src/sqlancer/databend/DatabendErrors.java
@@ -42,6 +42,9 @@ public final class DatabendErrors {
         if (DatabendBugs.bug9806) {
             errors.add("segment pruning failure");
         }
+        if (DatabendBugs.bug15568) {
+            errors.add("Decimal overflow at line : 723 while evaluating function `to_decimal");
+        }
 
         /*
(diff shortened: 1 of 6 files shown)


src/sqlancer/databend/DatabendErrors.java
```diff
@@ -42,6 +44,9 @@ public static List<String> getExpressionErrors() {
         if (DatabendBugs.bug9806) {
             errors.add("segment pruning failure");
         }
+        if (DatabendBugs.bug15568) {
+            errors.add("Decimal overflow at line : 723 while evaluating function `to_decimal");
```
mrigger, line 48: That's fine for this PR. Perhaps it would make the code more future proof to use regular expressions rather than line numbers.

mrigger (approved):

PR #954, opened 2024-07-08
[DataFusion] Initial implementation of SQLancer

2010YOUY01:
I have to say `SQLancer`'s idea is brilliant
With minimal implementation, we have identified ~ten new bugs https://github.com/apache/datafusion/issues?q=is%3Aissue+label%3Abug+sqlancer+

Could you take a look and give some suggestions when you have time?

This PR includes:
SQL features `SELECT ... FROM ... WHERE`
Typed expression generation of several numeric expressions/functions
`NoREC` and `TLP-Where` oracles
(CI should be deterministic, datafusion engine version is fixed and different runs produce the same result.)


mrigger:
Thanks a lot for the PR and very happy that it has helped making DataFusion more reliable! Will review asap.

src/sqlancer/datafusion/DataFusionProvider.java
```diff
@@ -0,0 +1,135 @@
...
+@AutoService(DatabaseProvider.class)
+public class DataFusionProvider extends SQLProviderAdapter<DataFusionGlobalState, DataFusionOptions> {
+
+    public DataFusionProvider() {
+        super(DataFusionGlobalState.class, DataFusionOptions.class);
+    }
+
+    @Override
+    public void generateDatabase(DataFusionGlobalState globalState) throws Exception {
+        int tableCount = Randomly.fromOptions(1, 2, 3, 4, 5, 6, 7);
```
mrigger, line 39: Based on my experience, having more than 3 tables might cause throughput to become very low (e.g., below 100 queries per second). Did you encounter this?

src/sqlancer/datafusion/DataFusionProvider.java
```diff
@@ -0,0 +1,135 @@
...
+
+        // Now only `INSERT` DML is supported
+        // If more DMLs are added later, should use`StatementExecutor` instead
+        // (see DuckDB's implementation for reference)
+
+        globalState.updateSchema();
+        List<DataFusionTable> allTables = globalState.getSchema().getDatabaseTables();
+        List<String> allTablesName = allTables.stream().map(t -> t.getName()).collect(Collectors.toList());
+        if (allTablesName.isEmpty()) {
+            System.out.println("Generate database failed");
```
mrigger, line 55: Could we make this an `AssertionError` or so instead? I.e., `throw new AssertionError("Generate database failed")`. Like this, some potential clean-up code might still be executed.

src/sqlancer/datafusion/DataFusionProvider.java
```diff
@@ -0,0 +1,135 @@
...
+            exit(1);
+        }
+
+        // Randomly insert some data into existing tables
+        for (DataFusionTable table : allTables) {
+            int nInsertQuery = globalState.getRandomly().getInteger(0, 8); // [0, 10)
```
mrigger, line 61: Here, you could consider using the parameter that also other SQLancer implementations use (see https://github.com/sqlancer/sqlancer/blob/852c1e4fecc853dc6246b45a3cc3045da9d8e920/src/sqlancer/MainOptions.java#L30). It's fine to do this as a follow-up PR if you want, perhaps also while adopting the `StatementExecutor` class used by other implementations).

src/sqlancer/datafusion/DataFusionProvider.java
```diff
@@ -0,0 +1,135 @@
...
+                SQLQueryAdapter insertQuery = null;
+                try {
+                    insertQuery = DataFusionInsertGenerator.getQuery(globalState, table);
+                } catch (IgnoreMeException e) {
+                    // Only for special case: table has 0 column
+                    continue;
+                }
+
+                insertQuery.execute(globalState);
+                globalState.dfLogger.appendToLog(DML, insertQuery.toString() + "\n");
```
mrigger, line 73: I noticed that you do not only use SQLancer to log the executed statements, but also a DataFusion logger. Just to double check, is there any logic that prevents that the DataFusion logger files exceed the local storage?

src/sqlancer/datafusion/DataFusionProvider.java
mrigger: Or perhaps using `dfAssert` to be consistent with the other code?

src/sqlancer/datafusion/DataFusionUtil.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        return resultStringBuilder.toString();
+    }
+
+    public static void dfAssert(boolean condition, String message) {
+        if (!condition) {
+            String methodName = Thread.currentThread().getStackTrace()[2].getMethodName();
+            System.err.println("DataFusion assertion failed in function '" + methodName + "': " + message);
+            exit(1);
```
mrigger, line 73: Could this be a `new AssertionError()` with the above error message or so? It would be best to avoid direct calls to `exit` (also in other instances).

src/sqlancer/datafusion/server/datafusion_server/src/main.rs
```diff
@@ -0,0 +1,463 @@
+#[cfg(feature = "datafusion_dev")]
```
mrigger, line 1: Would it be possible to add a one-line comment to explain the purpose of this file?

mrigger (commented):
Thanks a lot for the PR. Sorry that it took a while to review - I've been traveling. Overall, the code looks good to me. I have some minor nitpicks, and it would be great if you could check whether they make sense to you and whether you could address them.

With respect to long-term maintenance, it's not ideal that the implementation patterns deviate from the other implementations (e.g., using your custom logger or custom way to control the generation of statements). It would be great if, potentially as part of a follow-up PR, you could either adopt the existing patterns or implement the functionality so that also other implementations can use it.

Thanks a lot again, and very excited by your PR!

2010YOUY01:
> Thanks a lot for the PR. Sorry that it took a while to review - I've been traveling. Overall, the code looks good to me. I have some minor nitpicks, and it would be great if you could check whether they make sense to you and whether you could address them.
> 
> With respect to long-term maintenance, it's not ideal that the implementation patterns deviate from the other implementations (e.g., using your custom logger or custom way to control the generation of statements). It would be great if, potentially as part of a follow-up PR, you could either adopt the existing patterns or implement the functionality so that also other implementations can use it.
> 
> Thanks a lot again, and very excited by your PR!

Thank you for the review!

The feedbacks have been addressed.
(Except for totally getting rid of `custom*` stuff: I agree with your opinion about `custom*` thing, it's just some functionality I want but the `SQLancer` core hasn't been implemented yet, so I instead implemented it in a hacked way, I will try to fix and avoid them in the future.)

For the future plan, we will keep a separate branch for development and sync with upstream once in a while, if there is anything that requires modifying `SQLancer` common part, we will try to send small PRs for such modifications early.

Thank you again for such a great tool, please let me know if there is any further change required.

src/sqlancer/datafusion/DataFusionProvider.java
2010YOUY01: Yes, so our query-generating policy is to choose up to 3 tables from all existing tables for the general case.

src/sqlancer/datafusion/DataFusionProvider.java
2010YOUY01: Yes, it has error handling inside (if write fails it will throw `AssertionError`)

mrigger:
Great, thanks! I've merged the PR.

PR #956, opened 2024-07-15
Avoid storage parameters during CREATE TABLE for partitioned tables

robins:
In Postgres, CREATE TABLE WITH() allows storage parameters, but partitioned tables emit an error (given below) if attempted. This is because partitioned non-leaf tables are virutal tables [1] and don't accept storage parameters.

Sample Error:
"ERROR: cannot specify storage parameters for a partitioned table"

This patch skips storage parameter generation for partitioned tables.

Ref:
1. https://www.postgresql.org/docs/current/ddl-partitioning.html


mrigger (approved):
LGTM, thanks!

PR #959, opened 2024-07-19
Materialize: Use different database name

def-:
Since materialize already exists by default and is owned by mz_system:
```
org.postgresql.util.PSQLException: ERROR: must be owner of DATABASE materialize
	at sqlancer.qpg.materialize.TestMaterializeQueryPlan.testMaterializeQueryPlan(TestMaterializeQueryPlan.java:33)
```


mrigger (approved):
Thanks!

PR #962, opened 2024-08-21
Update new AST to support common test oracles

malwaregarry:
Update new AST implementations to not use `Node<T>` so that existing common test oracles can be used

Affected DBMS:
1. Databend
2. DataFusion
3. Doris
4. DuckDB
5. H2
6. HSQLDB
7. Presto
8. QuestDB
9. Yugabyte YCQL


src/sqlancer/common/ast/newast/NewBinaryOperatorNode.java
```diff
@@ -5,10 +5,10 @@
 public class NewBinaryOperatorNode<T> implements Node<T> {
 
     protected final Operator op;
-    protected final Node<T> left;
-    protected final Node<T> right;
+    protected final T left;
```
mrigger, line 8: I'm wondering about the semantics of this change (as well as the current implementation). Conceptually, I think that an AST consists of nodes that in turn have nodes as children. Based on this, should the type of this node not be the same as its children? The same would apply also to the other nodes.

PR #965, opened 2024-08-28
[DuckDB] Use common NoREC oracle

malwaregarry:


src/sqlancer/duckdb/DuckDBErrors.java
```diff
@@ -21,6 +21,7 @@ public static List<String> getExpressionErrors() {
         errors.add("can't be cast because the value is out of range for the destination type");
         errors.add("field value out of range");
         errors.add("Unimplemented type for cast");
+        errors.add("Duplicate alias");
```
mrigger, line 24: Is this a legitimate expected error? Just checking since it seems we didn't include this error before.

malwaregarry: you're right, I should have investigated further. It was caused by the refactoring resulting in duplicate tables in select statements

mrigger (approved):
LGTM, thanks!

PR #970, opened 2024-09-02
[CockroachDB] Use common NoREC oracle

malwaregarry:
Also updated the test to include TLP. 
Did not add CERT as a lot of errors were thrown and there seems to be some problem with the implementation.


mrigger (approved):
LGTM! For CERT, it might be expected. If you want, you could follow up in a separate PR and ping @bajinsheng to check.

PR #977, opened 2024-09-06
[Presto] Add CI & tests

malwaregarry:
Added CI for Presto and some expected errors but tests are still failing


mrigger:
It seems there are some overflow errors, which should probably be added as expected errors. Do you want to do that, or should I? Either works for me.

mrigger:
@branimir-vujicic, also CC'ing you here in case you want to have a look. Having a CI for Presto will help us not break anything in upcoming refactorings.

malwaregarry:
strange that there are already expected errors for overflow https://github.com/sqlancer/sqlancer/blob/326a8d52e5b6e2f24a19f7de83d98ad20f81bb88/src/sqlancer/presto/PrestoErrors.java#L27-L30

mrigger:
It's surprising that the query throughput is that low, which might explain the subsequent timeouts. @branimir-vujicic, do you remember what the rough query throughput was when you tested?

branimir-vujicic:
> It's surprising that the query throughput is that low, which might explain the subsequent timeouts. @branimir-vujicic, do you remember what the rough query throughput was when you tested?

Hi @mrigger  rough query throughput was 15 - 20  queries/s

tdcmeehan:
Folks, I am happy that this PR is being added.  For what it's worth, I recently ran SQLancer and discovered another bug that you might want to consider adding to your blocklist (we are working on it): https://github.com/prestodb/presto/issues/23324

On the throughput question, is it that we expect higher throughput from Presto, or we expect SQLAncer to generate queries faster?

mrigger:
> Hi @mrigger rough query throughput was 15 - 20 queries/s

That's quite low. On my machine, I remember that the typical rate is more like 1000 to 20,000 queries per second depending on the DBMS.

> On the throughput question, is it that we expect higher throughput from Presto, or we expect SQLAncer to generate queries faster?

It might be either or both. Typically, we try to balance this by adding only few tables and rows (typically around 30 `INSERT` statements). I checked the code and did not spot anything unusual. For example, only up to 2 tables are generated with the default number of `INSERT`s: https://github.com/sqlancer/sqlancer/blob/646315fa22022a6b864c712e1b9c72f27b369682/src/sqlancer/presto/PrestoProvider.java#L53

> Folks, I am happy that this PR is being added.

Also excited about @malwaregarry's contributions in making SQLancer more maintainable!

> I recently ran SQLancer and discovered another bug that you might want to consider adding to your blocklist (we are working on it): https://github.com/prestodb/presto/issues/23324

Thanks! We should indeed ignore this bug for now.

malwaregarry:
After some changes, tests seems to pass now, though could still fail

mrigger:
Okay, let's merge it for now and check any subsequent failures if we observe any.

mrigger (approved):

PR #978, opened 2024-09-06
[CnosDB] Add CI & tests

malwaregarry:
Added tests but current implementation results in errors. The database also crashes sporadically

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -92,6 +92,32 @@ jobs:
       - name: Run Tests
         run: CITUS_AVAILABLE=true mvn -Dtest=TestCitus test
 
+  cnosdb:
+    name: DBMS Tests (CnosDB)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up JDK 11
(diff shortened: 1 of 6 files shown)


mrigger:
Does it still result in errors/crashes? I see the current CI check passes. In case the crashes can be prevented by setting the number of queries to 0, that might also be an option.

malwaregarry:
let me investigate it some more tomorrow first before merging

malwaregarry:
I couldn't figure out what is causing the queries to crash the database. Can we leave the number of queries to be at 0?

malwaregarry:
> With the number of queries being set to 0, does the database system still crash?
No crashes with 0 queries.

I've added a comment in the test files

mrigger:
Sounds good, then let's merge it. It seems after the added comment, the formatting test is failing now.

Should we also open an issue in CnosDB to point out that SQLancer still triggers crashes in their system?

malwaregarry:
> Should we also open an issue in CnosDB to point out that SQLancer still triggers crashes in their system?

There are some similar issues open reporting crashes: https://github.com/cnosdb/cnosdb/issues/1801  https://github.com/cnosdb/cnosdb/issues/1445

PR #979, opened 2024-09-09
[HSQLDB] Add CI & tests

malwaregarry:


mrigger:
It seems that the CI tests pass. Something seems off, though:

```
[2024/09/09 06:44:45] Executed 0 queries (0 queries/s; 2.80/s dbs, successful statements:  3%). Threads shut down: 0.
```
The successful statement rate is low and the query throughput rate is zero. Not sure what's wrong. Do you want to investigate or should we merge this first, as having some CI tests is already better that having none for HSQLDB?

malwaregarry:
seems to only affect NoREC.  How about we merge this first, and I can look into it while updating it to use the common oracle?

mrigger (approved):
LGTM!

PR #988, opened 2024-09-16
[MySQL] Use common TLP-where oracle

malwaregarry:


src/sqlancer/mysql/MySQLErrors.java
```diff
@@ -32,6 +32,8 @@ public static List<Pattern> getExpressionRegexErrors() {
                                                                        // "00000000000000000000-0"
         }
 
+        errors.add(Pattern.compile("Unknown column '.*' in 'order clause'"));
```
mrigger, line 35: That's a bit surprising. Should the column not always exist?

malwaregarry: Columns will always exits. Error is thrown because `generateOrderBys()` can generate constants

mrigger (approved):

PR #999, opened 2024-09-18
[QuestDB] Update to 8.1.1 and add tests

malwaregarry:


src/sqlancer/questdb/QuestDBErrors.java
```diff
@@ -21,12 +22,34 @@ public static List<String> getExpressionErrors() {
         errors.add("too few arguments for 'in'");
         errors.add("cannot compare TIMESTAMP with type"); // WHERE column IN with nonTIMESTAMP arg
         errors.add("constant expected");
+        errors.add("Invalid date");
+        errors.add("invalid timestamp");
+        errors.add("invalid column");
+        errors.add("expression type mismatch");
+        errors.add("inconvertible value");
+        errors.add("not implemented: dynamic pattern would be very slow to execute");
+        errors.add("Cannot invoke \"java.lang.CharSequence.length()\" because \"seq\" is null");
```
mrigger, line 31: This seems like a potential bug, since Java error messages should presumably never surface to the user?

malwaregarry:
i'm not too sure how to "disable" the first error since it could surface from expressions containing null

puzpuzpuz:
> @puzpuzpuz, any chance that the above-referenced issues will be fixed soon, so we can test (and subsequently expand) SQLancer support for QuestDB?

Thanks for the heads-up! We'll check the issues and try to fix them as soon as possible. I can't name any ETAs though.

PR #1000, opened 2024-09-20
[Citus] Use common TLP Where oracle

malwaregarry:


mrigger (approved):

mrigger:
I guess we can address the "ERROR: Foreign keys and AFTER ROW triggers are not supported for columnar tables" error by adding this as an expected error?

mrigger (approved):

PR #1014, opened 2024-10-01
[CockroachDB] Add CERT tests

malwaregarry:
* Updated invariant for JOINs
* Reported bugs: https://github.com/cockroachdb/cockroach/issues/131640, https://github.com/cockroachdb/cockroach/issues/131647, https://github.com/cockroachdb/cockroach/issues/131875


mrigger (commented):
Given that you reported issues here, should we also disable some mutators for now?

mrigger (approved):

PR #1016, opened 2024-10-02
[TiDB] Add CERT tests

malwaregarry:


src/sqlancer/tidb/oracle/TiDBCERTOracle.java
```diff
@@ -82,8 +83,16 @@ public void check() throws SQLException {
...
+        excludes.add(Mutator.DISTINCT);
+        if (TiDBBugs.bug51525) {
+            excludes.add(Mutator.OR);
+        }
+        if (TiDBBugs.bug38319) {
+            excludes.add(Mutator.GROUPBY);
+        }
         // Mutate the query
-        boolean increase = mutate(Mutator.DISTINCT);
+        boolean increase = mutate(excludes.toArray(new Mutator[0]));
```
mrigger, line 95: I was briefly confused by this, as it is somewhat surprising that `mutate`'s argument do not use the arguments to mutate, but excludes those mutators. I guess a more helpful name could be something like `mutateExcept`. However, this is not an issue in this PR, but could be a follow-up refactoring?

mrigger (approved):

src/sqlancer/tidb/oracle/TiDBCERTOracle.java
malwaregarry: I can fix this when working on the common oracle

PR #1035, opened 2024-11-13
Add influx db

timbytimmy:
add more class to influxdb


test/sqlancer/dbms/TestMySQLPQS.java
```diff
@@ -11,12 +11,16 @@ public class TestMySQLPQS {
```
mrigger, line 11: I guess that's an unrelated change. Can we undo it?

src/sqlancer/influxdb/InfluxDBProvider.java
```diff
@@ -0,0 +1,112 @@
...
+
+@AutoService(DatabaseProvider.class)
+public class InfluxDBProvider extends SQLProviderAdapter<InfluxDBProvider.InfluxDBGlobalState, InfluxDBOptions> {
+
+    public InfluxDBProvider() {
+        super(InfluxDBGlobalState.class, InfluxDBOptions.class);
+    }
+
+    public enum Action implements AbstractAction<InfluxDBGlobalState> {
+        WRITE_POINT(sqlancer.influxdb.gen.InfluxDBInsertGenerator::getQuery),
```
mrigger, line 26: Is it intended that both of this refers to the `InsertGenerator`? I think it would also be good to stick to the existing naming conventions, which use `INSERT` as part of the action name.

src/sqlancer/influxdb/InfluxDBProvider.java
```diff
@@ -0,0 +1,112 @@
...
+        public SQLQueryAdapter getQuery(InfluxDBGlobalState state) throws Exception {
+            return sqlQueryProvider.getQuery(state);
+        }
+    }
+
+    private static int mapActions(InfluxDBGlobalState globalState, Action a) {
+        Randomly r = globalState.getRandomly();
+        switch (a) {
+            case WRITE_POINT:
+                return r.getInteger(0, globalState.getOptions().getMaxNumberWrites());
```
mrigger, line 46: We have an option related to `INSERT`s. Is there some overlap with this method? If not, do we use it for `INSERT`s?

https://github.com/sqlancer/sqlancer/blob/04f0b49ccbcee1054ef93f27b0055f2aa976ca05/src/sqlancer/MainOptions.java#L30

src/sqlancer/influxdb/InfluxDBProvider.java
```diff
@@ -0,0 +1,112 @@
...
+public class InfluxDBProvider extends SQLProviderAdapter<InfluxDBProvider.InfluxDBGlobalState, InfluxDBOptions> {
+
+    public InfluxDBProvider() {
+        super(InfluxDBGlobalState.class, InfluxDBOptions.class);
+    }
+
+    public enum Action implements AbstractAction<InfluxDBGlobalState> {
+        WRITE_POINT(sqlancer.influxdb.gen.InfluxDBInsertGenerator::getQuery),
+        CREATE_DATABASE(sqlancer.influxdb.gen.InfluxDBInsertGenerator::getQuery),
+        InfluxDBWritePointGenerator();
```
mrigger, line 28: What does `InfluxDBWritePointGenerator();` do here?

src/sqlancer/influxdb/InfluxDBOptions.java
```diff
@@ -0,0 +1,37 @@
...
+@Parameters(separators = "=", commandDescription = "InfluxDB (default port: " + InfluxDBOptions.DEFAULT_PORT
+        + " default host: " + InfluxDBOptions.DEFAULT_HOST + ")")
+public class InfluxDBOptions implements DBMSSpecificOptions<InfluxDBOracleFactory> {
+    public static final String DEFAULT_HOST = "localhost";
+    public static final int DEFAULT_PORT = 8086;
+
+    @Parameter(names = "--oracle")
+    public List<InfluxDBOracleFactory> oracle = Arrays.asList(InfluxDBOracleFactory.DEFAULT);
+
+    @Parameter(names = "--username", description = "The user name used to log into InfluxDB")
```
mrigger, line 18: We have options for user name and password in the main option class, so I think we should avoid duplicating them.

https://github.com/sqlancer/sqlancer/blob/04f0b49ccbcee1054ef93f27b0055f2aa976ca05/src/sqlancer/MainOptions.java#L70

src/sqlancer/common/oracle/PivotedQuerySynthesisBase.java
```diff
@@ -50,6 +50,9 @@ public final void check() throws Exception {
         if (!pivotRowIsContained) {
             reportMissingPivotRow(pivotRowQuery);
         }
+        else{
+            System.out.println(isContainedQuery.getQueryString());
```
mrigger, line 54: Could we omit this? I guess it was included for debugging?

pom.xml
```diff
@@ -483,7 +491,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
-            <version>3.3.0</version>
+            <version>4.0.0</version>
```
mrigger, line 494: I think we can upgrade these packages as part of a separate PR to ensure that this does not break anything in the other implementations.

df
mrigger, line 1: I think this was included by accident?

mrigger (changes requested):
Thanks a lot for the PR and sorry for the late review. I've added comments to improve the code as well as some questions.

Before we can merge the PR, we also need to make sure that all warnings are fixed (see `mvn verify -DskipTests=true`). In addition, we should add a workflow that runs the CI (see `.github/workflows/main.yml`).

@malwaregarry recently refactored some of the relevant code. @malwaregarry, do you have any suggestions or requests?

Issue #1044, opened 2024-12-09
Add support for Postgres v16


robins:
Creating a placeholder issue for SQLancer to run on Postgres v16.

Alike support for v15 Issue ( #1043 ) - I intend to let SQLancer run on the target version (here `REL_16_STABLE`) and indicate errors - and if any, should ideally be fixed one-by-one.

robins:
@PradeepFSTdhane123  yes - Please see this comment for more context https://github.com/sqlancer/sqlancer/issues/1045#issuecomment-2716334071

robins:
A few things that changed in v16:

1. JSON related changes:
  a. ABSENT ON NULL / NULL ON NULL
  b. WITH UNIQUE
  c. RETURNING
  d. IS JSON [ARRAY / OBJECT] etc..
2. underscores (_) as thousands separator
3. EXPLAIN (GENERIC_PLAN)
4. [VACUUM / ANALYSE] BUFFER_USAGE_LIMIT
5. SQL syntax changes
  a. GRANT ... WITH INHERIT / WITH ADMIN / SET ROLE
  b. ALTER GROUP ADD USER ADMIN
6. (Ensure) GENERATED works with inherited / partition tables
7. force_parallel_mode rename to debug_parallel_query 
8. (Possibly No-Op but) Many new functions added to the catalog

PR #1046, opened 2024-12-09
[Postgres][v16] Rename force_parallel_mode to debug_parallel_query

robins:
This is working towards #1044 - Ensuring that SQLancer works well with Postgres 16

In Postgres v16 [1] `force_parallel_mode` was renamed to `debug_parallel_query`. This causes avoidable errors during SQLancer runs. 

One way here is to put in `Errors` for both `debug_parallel_query` AND `force_parallel_mode` - which reduces irrespective of old or newer version. However IMHO this would reduce efficiency unnecessarily and thus this patch that renames the parameter, since now 2 Major versions (v16 and v17) already are on the newer parameter.

Reference:
1. https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=5352ca22e


robins:
Given the discussion happening here, I am willing to re-submit this PR using Approach 3 - if required.

PR #1054, opened 2025-01-05
Support CODDTest in sqlancer

DerZc:


src/sqlancer/cockroachdb/CockroachDBOptions.java
```diff
@@ -30,6 +30,9 @@ public class CockroachDBOptions implements DBMSSpecificOptions<CockroachDBOracle
     @Parameter(names = { "--max-num-indexes" }, description = "The maximum number of indexes that can be created")
     public int maxNumIndexes = 20;
 
+    @Parameter(names = { "--coddtest-model" }, description = "Apply CODDTest on expression, subquery, or random")
+    public String coddTestModel = "random";
```
mrigger, line 34: Perhaps we could use a similar strategy as the TLP oracles where we have separate enums?

src/sqlancer/cockroachdb/CockroachDBToStringVisitor.java
```diff
@@ -231,4 +249,155 @@ public void visit(CockroachDBMultiValuedComparison comp) {
...
+            for (CockroachDBColumn c: vs.keySet()) {
+                sb.append(vs.get(c).get(i).toString());
+                if (!(vs.get(c).get(i) instanceof CockroachDBNullConstant)) {
+                    if (c.getType() != null) {
+                        sb.append("::" + c.getType().toString());
+                    }
+                }
+                sb.append(", ");
+            }
+            sb.deleteCharAt(sb.length() - 1);
```
mrigger, line 285: Deleting previous parts seems quite complicated. Can we simplify this (and similar logic in other files)?

src/sqlancer/cockroachdb/ast/CockroachDBConstant.java
```diff
@@ -38,6 +44,53 @@ public long getValue() {
...
+            return this.val;
+        }
+
+        @Override
+        public String toString() {
+            return String.valueOf(val);
+        }
+
+        @Override
+        public String toStringForComparison() {
```
mrigger, line 73: This seems quite complicated. Can we simplify and/or document this?

src/sqlancer/cockroachdb/ast/CockroachDBWithClasure.java
```diff
@@ -0,0 +1,24 @@
+package sqlancer.cockroachdb.ast;
+
+public class CockroachDBWithClasure implements CockroachDBExpression {
```
mrigger, line 3: What does the name mean?

src/sqlancer/duckdb/DuckDBErrors.java
```diff
@@ -63,6 +63,15 @@ public static List<String> getExpressionErrors() {
                     "FATAL Error: Failed: database has been invalidated because of a previous fatal error. The database must be restarted prior to being used again");
         }
 
+        // added by CODDTest
+        errors.add("must appear in the GROUP BY clause or be used in an aggregate function");
+        errors.add("must appear in the GROUP BY clause or must be part of an aggregate function");
+        errors.add("GROUP BY term out of range - should be between");
+        errors.add("INTERNAL Error: Failed to bind column reference");
```
mrigger, line 70: Internal errors are bugs, so we should not ignore them.

src/sqlancer/duckdb/ast/DuckDBConstantWithType.java
```diff
@@ -0,0 +1,31 @@
+package sqlancer.duckdb.ast;
+
+import sqlancer.duckdb.DuckDBSchema.DuckDBCompositeDataType;
+import sqlancer.duckdb.ast.DuckDBConstant.DuckDBNullConstant;
+
+public class DuckDBConstantWithType implements DuckDBExpression {
```
mrigger, line 6: Can we achieve the same with a cast expression to which we pass a constant?

src/sqlancer/sqlite3/ast/SQLite3Expression.java
```diff
@@ -1552,4 +1563,192 @@ public boolean omitBracketsWhenPrinting() {
...
+            return this.columns;
+        }
+
+        @Override
+        public SQLite3CollateSequence getExplicitCollateSequence() {
+            return null;
+        }
+    }
+
+    public static class SQLite3ExpressionBag extends SQLite3Expression {
```
mrigger, line 1674: For some of the newly added classes, it's not immediately clear what this looks like at a SQL level. would it be possible to add a comment for such cases?

mrigger (commented):
Thanks for the PR! Added some initial comments. I think it's quite challenging to review such a large PR. Do you think we could split it into multiple smaller ones (referencing this as the main PR)? For example, as a first step, we could add some of the new expression nodes and the visit methods.

PR #1064, opened 2025-01-15
[Databend] Support Date and Timestamp type

b41sh:
This PR support generate Date and Timestamp type values for Databend

### DIFF
--- a/src/sqlancer/databend/ast/DatabendConstant.java
+++ b/src/sqlancer/databend/ast/DatabendConstant.java
@@ -48,6 +48,13 @@ public abstract class DatabendConstant implements DatabendExpression {
         throw new UnsupportedOperationException(this.toString());
     }
 
+    protected Timestamp truncateTimestamp(long val) {
+        // Databend supports `date` and `timestamp` type where the year cannot exceed `9999`,
+        // the value is truncated to ensure generate legitimate `date` and `timestamp` value.
+        long t = val % 253380000000000L;
+        return new Timestamp(t);
+    }
+
     public abstract DatabendConstant isEquals(DatabendConstant rightVal);
 
     public abstract DatabendConstant isLessThan(DatabendConstant rightVal);
@@ -322,10 +329,7 @@ public abstract class DatabendConstant implements DatabendExpression {
         public String textRepr;
 
         public DatabendDateConstant(long val) {
-            // Databend supports `date` type where the year cannot exceed `9999`,
-            // the value is truncated to ensure generate legitimate `date` value.
-            long t = val % 253380000000000L;
-            Timestamp timestamp = new Timestamp(t);
+            Timestamp timestamp = truncateTimestamp(val);
             SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd");
             textRepr = dateFormat.format(timestamp);
         }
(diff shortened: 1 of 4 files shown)


mrigger:
Thanks! It seems some expected errors need to be declared (and bugs reported and marked as known bugs). See https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#expected-errors and https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#unfixed-bugs.

b41sh:
> Thanks! It seems some expected errors need to be declared (and bugs reported and marked as known bugs). See https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#expected-errors and https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#unfixed-bugs.

@mrigger I have fixed the error report, PTAL

src/sqlancer/databend/ast/DatabendConstant.java
```diff
@@ -318,7 +322,8 @@ public static class DatabendDateConstant extends DatabendConstant {
         public String textRepr;
 
         public DatabendDateConstant(long val) {
-            Timestamp timestamp = new Timestamp(val);
+            long t = val % 250000000000000l;
```
mrigger, line 325: The formatting tests fail (see `mvn verify -DskipTests=true`):
> Error:  src/sqlancer/databend/ast/DatabendConstant.java:[325,28] (misc) UpperEll: Should use uppercase 'L'.
> Error:  src/sqlancer/databend/ast/DatabendConstant.java:[361,28] (misc) UpperEll: Should use uppercase 'L'.

While fixing this, could you also extract the constant to a named one that explains its purpose? I guess it's some kind of truncation factor?

After this, we can merge the change. Thank you!

b41sh: I have add some comments, thanks for your review @mrigger

src/sqlancer/databend/ast/DatabendConstant.java
```diff
@@ -318,7 +322,10 @@ public static class DatabendDateConstant extends DatabendConstant {
         public String textRepr;
 
         public DatabendDateConstant(long val) {
-            Timestamp timestamp = new Timestamp(val);
+            // Databend supports `date` type where the year cannot exceed `9999`,
+            // the value is truncated to ensure generate legitimate `date` value.
+            long t = val % 253380000000000L;
```
mrigger, line 327: Thanks, for merging this after the tests pass. As a minor nitpick: I think a named constant would have been better as we have the same comment and truncation factor twice in the code.

b41sh:
https://github.com/databendlabs/databend/issues/17337
Found a Databend bug, sqlancer is so helpful, thank you @mrigger

mrigger:
Thanks a lot for performing the changes! Should I go ahead and merge the PR now?

I noticed that the Databend CI test is currently failing, as an SQL statement causes the server to exit. If it's because of the above bug, do you think we can avoid triggering this bug for now so that the CI tests stay green? See here: https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#unfixed-bugs. This could also be part of a follow-up PR.

b41sh:
> Thanks a lot for performing the changes! Should I go ahead and merge the PR now?
> 
> I noticed that the Databend CI test is currently failing, as an SQL statement causes the server to exit. If it's because of the above bug, do you think we can avoid triggering this bug for now so that the CI tests stay green? See here: https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#unfixed-bugs. This could also be part of a follow-up PR.

We can merge this PR for now, and I'll continue to add some more features and fix the CI issue in the meantime!

mrigger:
@b41sh any updates on the CI fix issue? This would be important, as we can currently no longer check whether the Databend implementation works.

PR #1070, opened 2025-02-01
Minor tweaks to CockroachDBProvider

shailendra-patel:
* use SecureRandom
* support multi-host pgurl
* temporarily disable COLLATE (known bug)
* temporarily disable JSONB and INTERVAL types (known bugs)


src/sqlancer/Main.java
```diff
@@ -629,7 +631,7 @@ private String formatInteger(long intValue) {
             final String databaseName = options.getDatabasePrefix() + i;
             final long seed;
             if (options.getRandomSeed() == -1) {
-                seed = System.currentTimeMillis() + i;
+                seed = SECURE_PRNG.nextLong();
```
mrigger, line 634: What's the rationale behind this change?

src/sqlancer/cockroachdb/CockroachDBProvider.java
```diff
@@ -273,10 +307,14 @@ public SQLConnection createDatabase(CockroachDBGlobalState globalState) throws S
             port = CockroachDBOptions.DEFAULT_PORT;
         }
         String databaseName = globalState.getDatabaseName();
-        String url = String.format("jdbc:postgresql://%s:%d/test", host, port);
+        String url = buildPGUrl(host, port, databaseName);
+        System.out.println("JDBC url="+ url);
```
mrigger, line 311: It would be great if we could remove such print statements before merging the change.

src/sqlancer/cockroachdb/CockroachDBSchema.java
```diff
@@ -24,7 +25,12 @@ public enum CockroachDBDataType {
         TIMETZ, ARRAY;
 
         public static CockroachDBDataType getRandom() {
-            return Randomly.fromOptions(values());
+            // return Randomly.fromOptions(values());
+            // INTERVAL and JSONB can trigger a number of known issues, e.g., [1], so we skip them.
+            // [1] https://github.com/cockroachdb/cockroach/issues/45993
```
mrigger, line 30: Would it be possible to adopt this scheme? https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#unfixed-bugs

mrigger: (also for some of the other cases)

mrigger (commented):
Thanks a lot for the PR! Added some suggestions and comments. Please tag me when I should review the PR again.

PR #1071, opened 2025-02-04
Update ClickHouseErrors.java

qoega:
Add new exception text to skip list.
Example of failure:
https://s3.amazonaws.com/clickhouse-test-reports/0/c1c6733f172fdb8ec655f4367bf62118102ef3df/sqlancer__debug_/TLPWhere.err


mrigger:
It seems the formatting test fails. Could you execute `mvn formatted:format` and ping me, so we can merge this change as soon as possible?

qoega:
Strange that I just added a single line and it complains. It was added via GitHub UI. I'll check what it is when I will get back to laptop.

qoega:
@mrigger now it is ok?

mrigger:
Thanks, I've merged it.

Issue #912, opened 2024-02-24
Add support for Postgres v13


robins:
Add support for Postgres v13.
These are the broad changes in v13, most base SQLs are not yet supported so NA for this task.

1. ALTER NO DEPENDS ON - not yet supported
2. ALTER STATISTICS SET STATISTICS - not yet supported
3. ALTER TABLE ALTER COLUMN DROP EXPRESSION - generated identity not yet supported
4. ALTER TABLE SET STORAGE - no change required
5. ALTER TYPE - not yet supported
6. ALTER VIEW RENAME COLUMN - not supported
7. CREATE DATABASE WITH LOCALE - not supported
8. CREATE INDEX - deduplication - no change required
9. CREATE TABLE / ALTER TABLE - new attributes - PENDING
10. CREATE TABLESPACE - not yet supported
11. DROP DATABASE FORCE - not yet supported
12. EXPLAIN ANALYSE - not yet supported
13. INSERT - Remove "Cannot insert intocolumn"?? PENDING
14. ROW - not yet supported
15. SELECT WITH TIES - not yet supported
16. VACUUM PARALLEL - PENDING

robins:
Hi,

Earlier this year, SQLancer officially supported only Postgres v12. Given that Postgres v12 [became EOL](https://www.postgresql.org/about/news/postgresql-172-166-1510-1415-1318-and-1222-released-2965/) a few days back, I'd like to work on getting Postgres to work in latest of Postgres versions. For this, I understand that the path is to go +1 version at a time.

Here, I am trying to find what would be a good test to to confidently say "SQLancer supports Postgres v13". For this, I ran SQLancer for half an hour (7+million tests) on Postgres v13.18 and I didn't see any error / crashes. Revisiting the list above (in this Issue), I see there are commands that changed in Postgres v13, but SQLancer still does not support. However I also think that that situation is similar to v12 (i.e. there were commands that changed in Postgres v12 but were missing support for in SQLancer).

Basically, I am looking for a well-defined way to to help us decide that running N minutes OR P hours OR X million tests on Postgres version Y - without error should give confidence to say "SQLancer supports Postgres version Y" to allow us / me to move on to the next version of Postgres and rinse-and-repeat.

Thanks

```
[...]

PR #1072, opened 2025-02-05
Mark support for v13, update test / comment URLs accordingly.

robins:
As per discussionsin Issue #912 , SQLancer works fine with Postgres v13.

This commit updates the test harness to run against Postgres v13 (instead of v12 currently).

Also, the documentation links in README & code-comment URLs point to unreleased / outdated postgres documentation versions - For e.g. links were pointing to different postgres version documentations like devel (unreleased) / current / v12 or even v9.5 / v11 (both already EOL).

This commit ensures that all www.postgresql.org documentation links in the comments, point to the supported version of postgres (which as of this commit is v13). As and when we advance support for a more recent version, we should update the doc links accordingly too.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -467,7 +467,7 @@ jobs:
       - name: Set up PostgreSQL
         uses: harmon758/postgresql-action@v1.0.0
         with:
-          postgresql version: '12'
+          postgresql version: '13'
           postgresql user: 'sqlancer'
           postgresql password: 'sqlancer'
           postgresql db: 'test'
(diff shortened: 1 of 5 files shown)


mrigger:
Thanks for the PR! It seems that an `ALTER TABLE` currently causes an unexpected error. I guess it can be merged as an expected error, after which we can merge the change.

```
 ALTER TABLE  ONLY t0 ALTER c0 DROP NOT NULL, ENABLE ROW LEVEL SECURITY, REPLICA IDENTITY NOTHING, ALTER c0  SET(n_distinct=0.1), CLUSTER ON i0, SET WITHOUT CLUSTER, ALTER c0 SET STORAGE MAIN; -- ERROR: column "c0" is in index used as replica identity
```

robins:
I had let this PR be, so as to be a part of the on-going GSoC, but it seems to create more confusion than I was hoping for.

I'll push a change and submit for another review.

mrigger (approved):
Thanks! Sorry, just saw this now.

PR #1080, opened 2025-03-02
Add support for FULLTEXT indexes in MySQLIndexGenerator

abdala-elgendy:
Currently, the code randomly chooses to add "UNIQUE" before the index, but many MySQL databases support FULLTEXT indexes. 
I checked if the table contains any TEXT or VARCHAR columns suitable for FULLTEXT indexing and 
I added a branch to randomly choose between UNIQUE and FULLTEXT indexes based on column types. 
if there is another better way to do that tell me


abdala-elgendy:
how can I reach to failed tests?

PR #1081, opened 2025-03-03
Add Support for more String Data Types MySQL 

ahmed-kamal2004:
According to MySQL documentation, additional data types need to be included, such as `INTEGER`, which falls under the numeric category, and `NATIONAL CHAR`, `NATIONAL VARCHAR`, `BLOB`, etc., which belong to the string category.
Any suggested improvements are much welcomed :rocket: :rocket:

### DIFF
--- a/src/sqlancer/mysql/MySQLSchema.java
+++ b/src/sqlancer/mysql/MySQLSchema.java
@@ -139,12 +139,25 @@ public class MySQLSchema extends AbstractSchema<MySQLGlobalState, MySQLTable> {
         case "mediumint":
         case "int":
         case "bigint":
+        case "integer":
             return MySQLDataType.INT;
         case "varchar":
         case "tinytext":
         case "mediumtext":
         case "text":
         case "longtext":
+        case "char":
+        case "character":
+        case "binary":
+        case "varbinary":
+        case "tinyblob":
+        case "blob":
+        case "mediumblob":
+        case "longblob":
+        case "enum":
+        case "set":
+        case "national char":
+        case "national varchar":
             return MySQLDataType.VARCHAR;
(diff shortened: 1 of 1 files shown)


src/sqlancer/mysql/MySQLSchema.java
```diff
@@ -139,12 +139,25 @@ private static MySQLDataType getColumnType(String typeString) {
...
         case "int":
         case "bigint":
+        case "integer":
             return MySQLDataType.INT;
         case "varchar":
         case "tinytext":
         case "mediumtext":
         case "text":
         case "longtext":
+        case "char":
```
mrigger, line 149: It seems none of these new data types is actually exercised, so this seems to be dead code.

src/sqlancer/Randomly.java
```diff
@@ -532,13 +533,7 @@ public String getAlphabeticChar() {
...
     // SQLancer previously used ThreadLocalRandom.current().nextLong(lower, upper)
     private static long getNextLong(long lower, long upper) {
-        if (lower > upper) {
-            throw new IllegalArgumentException(lower + " " + upper);
-        }
-        if (lower == upper) {
-            return lower;
-        }
-        return getThreadRandom().get().longs(lower, upper).findFirst().getAsLong();
+        return ThreadLocalRandom.current().nextLong(lower, upper);
```
mrigger, line 536: Can you explain why you think this change is necessary or helpful?

src/sqlancer/mysql/MySQLSchema.java
mrigger: I think adding support for a currently unsupported data type would be generally valuable.

PR #1082, opened 2025-03-03
Resolve issue #1090

abdala-elgendy:
Moved index column generation logic from create() into a dedicated helper method.
Helper Method (generateRandomComment()):
This method creates a random alphanumeric string. This string is appended after the prefix "idx_" to form a more diverse comment.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -314,7 +314,7 @@
     <dependency>
       <groupId>org.duckdb</groupId>
       <artifactId>duckdb_jdbc</artifactId>
-      <version>1.1.3</version>
+      <version>1.2.0</version>
     </dependency>
     <dependency>
       <groupId>com.facebook.presto</groupId>
--- a/src/sqlancer/duckdb/DuckDBErrors.java
+++ b/src/sqlancer/duckdb/DuckDBErrors.java
@@ -115,7 +115,7 @@ public final class DuckDBErrors {
         errors.addAll(getFunctionErrors());
 
         errors.add("NOT NULL constraint failed");
-        errors.add("PRIMARY KEY or UNIQUE constraint violated");
+        errors.add("PRIMARY KEY or UNIQUE constraint violation");
         errors.add("Duplicate key");
         errors.add("can't be cast because the value is out of range for the destination type");
         errors.add("Could not convert string");
--- a/src/sqlancer/duckdb/gen/DuckDBIndexGenerator.java
+++ b/src/sqlancer/duckdb/gen/DuckDBIndexGenerator.java
@@ -42,7 +42,7 @@ public final class DuckDBIndexGenerator {
         sb.append(")");
         errors.add("already exists!");
         if (globalState.getDbmsSpecificOptions().testRowid) {
-            errors.add("Cannot create an index on the rowid!");
+            errors.add("cannot create an index on the rowid");
         }
         return new SQLQueryAdapter(sb.toString(), errors, true);
     }


mrigger:
Thanks for the PR! It seems the PR only extracts code into a helper method? Or did I overlook anything?

abdala-elgendy:
the first time I added a method to generate random comments to diversify generated queries but the tests did not pass I thought because of this modification, when I deleted and reran the tests it didn't also pass " I'm just extracting a helper method to clarify the code in the second time"

abdala-elgendy:
I add MySQLBooleanConstant for boolean support.
the project ( in MySQL part ) likely did not have a dedicated way to represent TRUE and FALSE as separate boolean constants. 
MySQL treats BOOLEAN as a synonym for TINYINT(1), meaning TRUE was likely represented as 1 and FALSE as 0 in SQLancer.

image

generated queries might have like "WHERE is_active = 1" instead of "WHERE is_active = TRUE" 

I don't yet edit the generated queries part.

src/sqlancer/mysql/ast/MySQLBooleanConstant.java
```diff
@@ -0,0 +1,77 @@
...
+    public String getTextRepresentation() {
+        return value ? "TRUE" : "FALSE";
+    }
+
+    @Override
+    public MySQLConstant isEquals(MySQLConstant rightVal) {
+        if (rightVal.isBoolean()) {
+            return MySQLConstant.createBoolean(this.value == rightVal.asBooleanNotNull());
+        }
+        throw new UnsupportedOperationException("Cannot compare boolean with non-boolean values.");
```
mrigger, line 64: I think we should throw an `IgnoreMeException` here. This logic is primarily used by Pivoted Query Synthesis (PQS). In general, given that PQS is not widely used, I think we can simplify the implementation here and directly throw an `IgnoreMeException` rather than providing the `isBoolean` method and implementing `isEquals`. The current logic also does not seem complete or correct. For example, `SELECT 1 = TRUE` can be evaluated and evaluates to true, which the method would not cover.

src/sqlancer/mysql/ast/MySQLCastOperation.java
```diff
@@ -6,7 +6,7 @@ public class MySQLCastOperation implements MySQLExpression {
     private final CastType type;
 
     public enum CastType {
-        SIGNED, UNSIGNED;
+        SIGNED, UNSIGNED , INT, BOOLEAN ,TEXT;
```
mrigger, line 9: Please keep the PR focused on a single change. We can review and merge changes more quickly this way.

src/sqlancer/mysql/gen/datadef/MySQLIndexGenerator.java
```diff
@@ -160,4 +124,46 @@ private void indexType() {
     public void setNewSchema(MySQLSchema schema) {
         this.schema = schema;
     }
+    private String generateIndexColumns(MySQLTable table, MySQLExpressionGenerator gen) {
```
mrigger, line 127: For such refactorings, it would be great to add a separate commit or PR. This makes reviewing easier, as it is clearer whether something is a new feature or should preserve the existing functionality.

mrigger (commented):
Thanks for the PR! I added some comments/feedback to help improve it.

abdala-elgendy:
I forget to delete 
image
so tests faild

PR #1087, opened 2025-03-04
Update PAPERS.md

ligurio:
Remove a statement about many unfixed bugs in MariaDB and MySQL, because all reported bugs were fixed, see [1].

1. https://github.com/sqlancer/bugs/blob/master/bugs.json


mrigger:
It's still quite easy to find bugs in MySQL and MariaDB. We report a couple of new ones when we develop new testing approaches, as reviewers often expect bugs to be found in popular systems, but they hardly ever get fixed. Here are some random examples:
* https://bugs.mysql.com/bug.php?id=113180
* https://bugs.mysql.com/bug.php?id=113298
* https://bugs.mysql.com/bug.php?id=117242
* https://bugs.mysql.com/bug.php?id=117439
* https://bugs.mysql.com/bug.php?id=117261

ligurio:
@mrigger Manuel, thanks! Got you.

PR #1091, opened 2025-03-05
Resolves #1089 update force_parallel_mode to debug_parallel_query

### LINKED ISSUE #1089: Postgres AssertionError "SET LOCAL force_parallel_mode=off"
`force_parallel_mode` is no longer supported for PostgreSQL v16 and more 
Image

I would like to work on this issue

ahmed-kamal2004:
this PR resolves #1089 
`force_parallel_mode ` is no longer supported in PostgreSQL v 16 and above, and it is already deprecated in PostgreSQL v15
in contrast, `debug_parallel_query` is supported in PostgreSQL latest version

Ref:
[force_parallel_mode](https://postgresqlco.nf/doc/en/param/force_parallel_mode/)
[degub_parallel_query](https://www.postgresql.org/docs/17/runtime-config-developer.html#GUC-DEBUG-PARALLEL-QUERY)

### DIFF
--- a/src/sqlancer/postgres/gen/PostgresSetGenerator.java
+++ b/src/sqlancer/postgres/gen/PostgresSetGenerator.java
@@ -112,7 +112,7 @@ public final class PostgresSetGenerator {
         JIT("jit", (r) -> Randomly.fromOptions(1, 0)),
         JOIN_COLLAPSE_LIMIT("join_collapse_limit", (r) -> r.getInteger(1, Integer.MAX_VALUE)),
         PARALLEL_LEADER_PARTICIPATION("parallel_leader_participation", (r) -> Randomly.fromOptions(1, 0)),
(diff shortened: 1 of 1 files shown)


mrigger:
It seems that the Postgres tests are failing:
> ERROR: unrecognized configuration parameter "debug_parallel_query"

The current version we are supporting does not support the new configuration parameter. Thus, before this PR, we should first upgrade to a newer Postgres version so that we can add the changed parameter.

mrigger:
I just found that we have a PR for this already by @robins. See https://github.com/sqlancer/sqlancer/pull/1046.

robins:
FYI - this parameter is effectively a rename, and importantly SQLancer currently doesn't support code for supporting multiple versions of postgres.

With the above in mind, the reason why SQLancer can't support this parameter rename today is because IMHO SQLancer first needs to 'reach' v16 support. Before v16 is supported, merging this PR would mean that postgres is unaware of the new parameter... in effect - PR #1046 is basically then waiting for SQLancer to begin supporting v16.

PR #1092, opened 2025-03-07
 add unit tests validates  generating queries in ClickHouse

abdala-elgendy:
- Add tests for NULL comparisons to ensure correct handling of NULL values.
- Add a test for Int64 and Float64 floating-point numbers and precision loss to verify correct behavior when comparing large integers.


mrigger:
Thanks for the PR! It seems the PR consists various unrelated changes. Could you please include only what is relevant for it?

mrigger:
In terms of tests, is the logic part of Pivoted Query Synthesis (PQS)? If so, @qoega, are you using PQS?

abdala-elgendy:
> It is not testing ClickHouse at all? Just tooling inside SQLancer?
> I only have NoREC and multiple TLP.

It validates SQLancer's internal logic (like ClickHouseConstant behavior) and Ensures generating valid constants (Int64, Float64, NULL), I don't handle NULL!=NULL yet 
.

mrigger:
Given @qoega's response, I think PQS is not used, and we could consider removing the methods associated with it in ClickHouse. Consequently, it also does not really help it if we test this logic. Sorry about this.

abdala-elgendy:
> Given @qoega's response, I think PQS is not used, and we could consider removing the methods associated with it in ClickHouse. Consequently, it also does not really help it if we test this logic. Sorry about this.

yes I understand Dr. Manuel

Issue #1100, opened 2025-03-09
Refactor: Remove redundant boundary checks in Randomly.java


AmrRiyad:
## Description

There's redundant boundary checking in the `getInteger` and `getLong` methods in the Randomly.java file.

The redundancy occurs because `getInteger()` checks if `left == right`, and then calls `getLong()`, which performs the same check again and also calls `getNextLong()` which performs the same check again. This creates an unnecessary duplicate check.

`getInteger()` -> `getLong()` -> `getNextLong()`

## Proposed Changes
- Remove the boundary checking in both functions `getInteger` and `getLong`.
- Ensure functionality remains unchanged.

## Impact

- Improves code clarity and reduces redundant operations.

mrigger:
Thanks for the issue! I'm not sure if we should change this. I agree that the check is redundant, but I think it is a defensive programming mechanism that each method sanitizes its input. Performance-wise, I doubt that the performance difference is measurable.

PR #1101, opened 2025-03-09
refactor: Redundant boundary check in getInteger and getLong methods #1100

bandish1304:
This code should remove redundancy.


mrigger:
Thanks for the PR! Please see me explanation in https://github.com/sqlancer/sqlancer/issues/1100

PR #1105, opened 2025-03-11
Remove redundant text.length() checks

### LINKED ISSUE #1104: Remove redundant text.length() > 0 checks in MySQLBinaryOperation.java
### Description  
In `src/sqlancer/mysql/ast/MySQLBinaryOperation.java`, there are multiple conditions that check `text.length() > 0` before calling `text.startsWith()`. However, if `text.startsWith(" ") || text.startsWith("\t")` is true, `text.length() > 0` is already guaranteed.

#### Example of redundant code:  
```java
while ((text.startsWith(" ") || text.startsWith("\t")) && text.length() > 0) {

SharmaNishchay:
Closes #1104 

Follow-up on Redundant Checks Removal

@mrigger I’ve updated the code by removing unnecessary text.length() > 0 and isEmpty() checks, since startsWith() already handles empty strings correctly.


mrigger (approved):
Thanks for the PR!

PR #1114, opened 2025-03-13
Bump duckdb version to 1.2.0

vic-ma:
The only assertion errors I found were due to rewordings of error messages.

I ran these two tests and they passed:
```
mvn clean -Dtest=TestDuckDBTLP test
mvn clean -Dtest=TestDuckDBNoREC test
```

The [release notes](https://duckdb.org/2025/02/05/announcing-duckdb-120#breaking-changes) mention breaking changes to `map` and `list_reduce`, but I don't think sqlancer uses either.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -314,7 +314,7 @@
     <dependency>
       <groupId>org.duckdb</groupId>
       <artifactId>duckdb_jdbc</artifactId>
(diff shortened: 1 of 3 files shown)


mrigger (approved):
LGTM! Thanks.

PR #1119, opened 2025-03-13
feat: Added support for orderBy

### LINKED ISSUE #1118: Add Support for ORDER BY in MySQLDeleteGenerator
# Support `ORDER BY` Clause in `MySQLDeleteGenerator`

## Description

The `MySQLDeleteGenerator` class currently generates `DELETE` statements with optional `ORDER BY` clauses. However, there is a need to ensure that the `ORDER BY` clause is properly supported.

## Proposed Changes

1. **Add Support for `ORDER BY` Clause:**
   - Ensure that the `ORDER BY` clause is correctly generated in the `DELETE` statements.
   - The `ORDER BY` clause should be optional and should randomly include either `ASC` or `DESC` order.

## Example

Here is an example of a `DELETE` statement with the `ORDER BY` clause:

```sql
DELETE FROM table_name WHERE condition ORDER BY column_name ASC, column_name2 DESC;

### PR DESCRIPTION
**Pull request by SharmaNishchay**
Closes #1118

### DIFF
--- a/src/sqlancer/mysql/gen/MySQLDeleteGenerator.java
+++ b/src/sqlancer/mysql/gen/MySQLDeleteGenerator.java
@@ -1,7 +1,6 @@
 package sqlancer.mysql.gen;
 
 import java.util.Arrays;
-import java.util.Random;
 
 import sqlancer.Randomly;
 import sqlancer.common.query.ExpectedErrors;
--- a/src/sqlancer/mysql/gen/MySQLDeleteGenerator.java
+++ b/src/sqlancer/mysql/gen/MySQLDeleteGenerator.java
@@ -49,9 +49,13 @@ public class MySQLDeleteGenerator {
             MySQLErrors.addExpressionErrors(errors);
         }
         if (Randomly.getBoolean()) {
-            sb.append(" ORDER BY ");
             List<MySQLExpression> orderBy = gen.generateOrderBys();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/sqlancer/mysql/gen/MySQLDeleteGenerator.java [hunk 1]
```diff
@@ -45,12 +45,17 @@ private SQLQueryAdapter generate() {
             sb.append(MySQLVisitor.asString(gen.generateExpression()));
             MySQLErrors.addExpressionErrors(errors);
         }
+        if (Randomly.getBoolean()) {
+            sb.append(" ORDER BY ");
+            sb.append(randomTable.getRandomColumn().getName());
```
mrigger, line 50: I think this code here can be more general, see https://github.com/sqlancer/sqlancer/blob/c2164e9f4831ae438c58c3a026adbe922b527525/src/sqlancer/mysql/oracle/MySQLPivotedQuerySynthesisOracle.java#L69.

SharmaNishchay: Thanks for the feedback! I will make the necessary changes.

mrigger (approved):
Thanks for the PR!

mrigger:
That was not the issue. I suspect you might have not run the code on MySQL. If you do, I think it becomes clear why it results in syntax errors.

SharmaNishchay:
Sir can you please provide the complete error for MySQL query?

SharmaNishchay:
Can we take this syntax fix in the next PR?

SharmaNishchay:
I’ll ensure all changes are properly tested before submitting next time. Appreciate the feedback!

PR #1123, opened 2025-03-13
[SQLite3] Optimize 'excluded' usage

### LINKED ISSUE #1124: [SQLite3] Optimize 'excluded' usage
# Ensure Consistent Use of Corresponding Excluded Columns in `ON CONFLICT UPDATE` Clause

## Description
The current implementation in the `ON CONFLICT UPDATE` clause randomly selects a column from the table when generating the update set expression. This randomness can lead to applying an excluded value from a column that may not share the same constraints (e.g., `nullable` vs. `non-nullable`) as the target column, potentially triggering constraint violations.

## Proposed Changes
- Modify the `UPDATE SET` clause to always reference the corresponding column from the excluded row (i.e., use `excluded.<column>` for the target column being updated) instead of randomly selecting a different column.
- Remove the randomness that currently allows a different column to be used, ensuring that the update expression consistently uses the correct matching column.

## Impact
- **Correctness:** Ensures that updates adhere to the column's constraints, reducing the risk of runtime errors due to constraint violations (e.g., non-nullable constraints or datatype mismatches).
- **Reduced Unsuccessful Queries:** Decreases the number of Unsuccessful queries generated by ensuring that the correct columns are used, minimizing errors due to constraint mismatches.

AmrRiyad:
# Ensure Consistent Use of Corresponding Excluded Columns in `ON CONFLICT UPDATE` Clause

## Description
The current implementation in the `ON CONFLICT UPDATE` clause randomly selects a column from the table when generating the update set expression. This randomness can lead to applying an excluded value from a column that may not share the same constraints (e.g., `nullable` vs. `non-nullable`) as the target column, potentially triggering constraint violations.

## Proposed Changes
- Modify the `UPDATE SET` clause to always reference the corresponding column from the excluded row (i.e., use `excluded.<column>` for the target column being updated) instead of randomly selecting a different column.
- Remove the randomness that currently allows a different column to be used, ensuring that the update expression consistently uses the correct matching column.

## Impact
- **Correctness:** Ensures that updates adhere to the column's constraints, reducing the risk of runtime errors due to constraint violations (e.g., non-nullable constraints or datatype mismatches).
- **Reduced Unsuccessful Queries:** Decreases the number of Unsuccessful queries generated by ensuring that the correct columns are used, minimizing errors due to constraint mismatches.

Feedback and suggestions for further improvements are welcome!
Resolves #1124

### DIFF
--- a/src/sqlancer/sqlite3/gen/dml/SQLite3InsertGenerator.java
+++ b/src/sqlancer/sqlite3/gen/dml/SQLite3InsertGenerator.java
@@ -103,10 +103,8 @@ public class SQLite3InsertGenerator {
                         sb.append(
                                 SQLite3Visitor.asString(SQLite3ExpressionGenerator.getRandomLiteralValue(globalState)));
                     } else {
-                        if (Randomly.getBoolean()) {
-                            sb.append("excluded.");
-                        }
-                        sb.append(table.getRandomColumn().getName());
+                        sb.append("excluded.");
+                        sb.append(columns.get(i).getName());
                     }
 
                 }


AmrRiyad:
## Example of an Unsuccessful Query Scenario

Consider the following scenario where an incorrect update might occur:

### Table Definition

```sql
CREATE TABLE test (
    id INTEGER PRIMARY KEY,
    col_1 INTEGER NOT NULL,
    col_2 INTEGER
);
```

## Original Query Flow
```sql
INSERT INTO test (id, col_1, col_2)
VALUES (1, 10, NULL);

INSERT INTO test (id, col_1)
VALUES (1, 30)
ON CONFLICT(id) DO UPDATE SET
    col_1 = excluded.col_2,
    col_2 = excluded.col_2;

SELECT * FROM test;
```
In this example, the random selection returned a column whose value is `NULL` and incorrectly updated the non-nullable column `col_1`.

## Updated Query Flow

```sql
INSERT INTO test (id, col_1)
VALUES (1, 30)
ON CONFLICT(id) DO UPDATE SET
    col_1 = excluded.col_1,
    col_2 = excluded.col_2;

SELECT * FROM test;
```
This corrected query ensures that each column is updated with its corresponding value from excluded, preventing unintended constraint violations.

mrigger:
Sorry for closing this PR. See the explanation in the issue you opened.

PR #1133, opened 2025-03-15
Added IllegalArgumentException to handle the possible exception.

codebreaker32:
- Added `IllegalArgumentException` to handle the possible exception and ensured consistent error handling across the ExpectedErrors.java


mrigger (commented):
Thanks for the PR. It is reasonable to apply more defensive programming practices, but please do not open further such PRs as there is much more room for higher-impact changes (such as addition of features).

codebreaker32:
Okay Sir.

PR #1137, opened 2025-03-16
update: update mysql to version 8.0.41 and added support for temporary tables

codebreaker32:
- Updated mysql image to 8.0.41
- Added support for temporary table
- Removed the usage of `DELAYED` keyword as it is officially deprecated and scheduled for removal in future releases.

Please go to [INSERT DELAYED DEPRECATED](https://dev.mysql.com/doc/refman/8.0/en/insert-delayed.html)


codebreaker32:
Please review it @robins @mrigger

PR #1141, opened 2025-03-16
[Postgres] feat: Add JSON support

ahmed-kamal2004:
Hi all :wave:,
This PR adds support for the basic functionality of JSON data types in Postgres
Supported Operations
- JSON in Standard Json representation
Screenshot from 2025-03-16 19-08-56

- JSON with Operators (support for `->>` and `->` operators)
Screenshot from 2025-03-16 19-05-40

Under this Code I have achieved (actually I stopped it, but it can continue)
Screenshot from 2025-03-16 19-07-31

Actually this PR is almost complete for basic functionality of JSON in Postgres but it isn't ready for merging with Sqlancer
as I need to modify the JSON generation strategy and handle using JSON with other blocks of the SQL query, those modifications are required to eliminate random errors

But I would like to hear your opinion on the Approach, and what can be added during the process, Thanks in advance :rocket:

### DIFF
--- a/src/sqlancer/postgres/PostgresExpectedValueVisitor.java
+++ b/src/sqlancer/postgres/PostgresExpectedValueVisitor.java
@@ -1,8 +1,28 @@
 package sqlancer.postgres;
 
-import sqlancer.postgres.ast.*;
+import sqlancer.postgres.ast.PostgresAggregate;
+import sqlancer.postgres.ast.PostgresBetweenOperation;
+import sqlancer.postgres.ast.PostgresBinaryLogicalOperation;
+import sqlancer.postgres.ast.PostgresCastOperation;
+import sqlancer.postgres.ast.PostgresCollate;
+import sqlancer.postgres.ast.PostgresColumnReference;
+import sqlancer.postgres.ast.PostgresColumnValue;
+import sqlancer.postgres.ast.PostgresConstant;
+import sqlancer.postgres.ast.PostgresExpression;
+import sqlancer.postgres.ast.PostgresExtractorJsonOperation;
+import sqlancer.postgres.ast.PostgresFunction;
+import sqlancer.postgres.ast.PostgresInOperation;
+import sqlancer.postgres.ast.PostgresLikeOperation;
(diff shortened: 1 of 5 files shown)


robins:
Thanks for working on this! @ahmed-kamal2004 

Before diving deeper, I see that the image in section "JSON in DDL" actually shows a DML. Is that intentional?

src/sqlancer/postgres/gen/PostgresInsertGenerator.java
```diff
@@ -113,12 +113,13 @@ private static void insertRow(PostgresGlobalState globalState, StringBuilder sb,
             }
             if (!Randomly.getBooleanWithSmallProbability() || !canBeDefault) {
                 PostgresExpression generateConstant;
-                if (Randomly.getBoolean()) {
-                    generateConstant = PostgresExpressionGenerator.generateConstant(globalState.getRandomly(),
+                if (Randomly.getBoolean() || columns.get(i).getType() == PostgresSchema.PostgresDataType.JSON) {
```
mrigger, line 116: I think we can leave the code here unchanged?

src/sqlancer/postgres/gen/PostgresUpdateGenerator.java
```diff
@@ -61,8 +61,8 @@ private SQLQueryAdapter generate() {
 
     @Override
     protected void updateValue(PostgresColumn column) {
-        if (!Randomly.getBoolean()) {
-            PostgresExpression constant = PostgresExpressionGenerator.generateConstant(globalState.getRandomly(),
+        if (!Randomly.getBoolean() || column.getType() == PostgresDataType.JSON) {
+            PostgresExpression constant = PostgresExpressionGenerator.generateConstantDDL(globalState.getRandomly(),
```
mrigger, line 65: I think we can leave the code here unchanged?

mrigger (commented):
Thanks for the PR! It seems the tests are currently failing.

I think it would make it easier to review the PR by splitting it into two parts:
1. Adding the data type and inserting values
2. Adding the extractor operators.

Issue #1139, opened 2025-03-16
Feat: Generate and check for NaN


SharmaNishchay:
## Description  
The `testDouble()` method in `TestRandomly` was missing `NaN` as a possible generated value.

## Expected Behavior  
- `getDouble()` should return all relevant edge cases, including `NaN`.  
- `testDouble()` should validate `NaN` along with other cases.  
- The test should complete successfully without infinite loops.

## Fix  
- Update `getDouble()` to generate `NaN`.  
- Modify `testDouble()` to check for `NaN`.

mrigger:
One challenge here is that we need to determine how we can map Java NaN values to NaN values of the different database systems.

shristi76:
can i solve this issue

PR #1142, opened 2025-03-16
Feat: Generate and check for NaN

SharmaNishchay:
Resolves: #1139 
Tested Run for `mvn package -DskipTests` and `TestRandomly.java`

P.S. I think some fixes are needed in caching of doubles , Working on that.


SharmaNishchay:
@mrigger The checks are failing , Do I need to modify every database dir for NaN handling and are there any specific files that I need to modify?

PR #1143, opened 2025-03-16
Update: Add OrderBy in MySQLDeleteGenerator

### LINKED ISSUE #1118: Add Support for ORDER BY in MySQLDeleteGenerator

codebreaker32:
- Add OrderBy in `MySQLDeleteGenerator`
- Remove usage of `DELAYED`

Resolves #1118 

I tried adding temporary table But after adding it I noticed that each
*   `CREATE TEMPORARY TABLE` got followed by `CREATE TABLE`  and query was generated on the basis of permanent table But temporary table shadows the permanent table. As both table have different attributes, It resulted in Error.

I didn't understand the cause of it.
Any guidance will be highly appreciated @mrigger @robins

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -406,7 +406,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       mysql:
-        image: mysql:8.0.36
+        image: mysql:8.0.41
         env:
           MYSQL_ROOT_PASSWORD: root
         ports:
--- a/pom.xml
+++ b/pom.xml
@@ -252,7 +252,7 @@
       <plugin>
         <groupId>com.github.spotbugs</groupId>
         <artifactId>spotbugs-maven-plugin</artifactId>
-        <version>4.7.3.0</version>
+        <version>4.9.2.0</version>
         <executions>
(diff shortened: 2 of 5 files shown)


codebreaker32:
please review it @mrigger

mrigger:
Please see my feedback in your previous PR. Also, please only include one logical change per PR.

codebreaker32:
Okay SIr.

Issue #1147, opened 2025-03-17
Missing Exception Handling in `catch` Block


PradeepFSTdhane123:
#### **Description**
In the `bugStillTriggers` method, the `catch` block silently ignores all exceptions (`Throwable`). This is problematic because:
1. **Silent Failures**: Exceptions are not logged or handled, making it difficult to debug issues.
2. **Loss of Debugging Information**: Critical error details (e.g., stack traces) are lost.

#### ** Code**
The following code snippet shows the issue:

```java
try {
    if (reproducer.bugStillTriggers(newGlobalState)) {
        newGlobalState.getLogger().logReduced(newGlobalState.getState());
        return true;
    }
} catch (Throwable ignoredException) {
    // Exceptions handeling ignored
}

mrigger:
This seems like an overly generic explanation and is not an issue in this case. After double-checking my doubts, it seems that it was AI-generated. Based on this, I will lower my involvement in reviewing your PRs.

![Image](https://github.com/user-attachments/assets/45eadfd1-aa62-49da-b6f8-b1a993864dee)

PradeepFSTdhane123:
@mrigger I apologize for the overly generic response,this wasn't  my attension to provide an AI - generated text. after solving this catch block many issue occer so for solving this issue I used AI   for perticularly where  issue is arrived in catch block. can you suggest me  in the situation where we need chages/solve issue but also issue is not solving properly then where we can ask for help

PR #1148, opened 2025-03-17
Missing Exception Handling in catch Block

PradeepFSTdhane123:
### **Description**
[#1147](https://github.com/sqlancer/sqlancer/issues/1147) 

fixes the issue of  ** exception handling** in the `catch` block of the `bugStillTriggers` method. Previously, exceptions were silently ignored, making it difficult to debug issues. This PR adds proper exception handling by logging the exception and providing meaningful context.


mrigger:
This is intended functionality.

Issue #1153, opened 2025-03-18
improving the insertValue method in the YCQLInsertGenerator class


PradeepFSTdhane123:
## Description
The `insertValue` method in `YCQLInsertGenerator` currently generates random constants for column values  into YCQL tables,as indicated by the following TODO comment in YCQLInsertGenerator.java, which may not always be meaningful or appropriate for the column's data type. This can lead to invalid or unrealistic data being inserted into the database.

Example:
```java
 @Override
    protected void insertValue(YCQLColumn columnYCQL) {
        // TODO: select a more meaningful value
        sb.append(YCQLToStringVisitor.asString(new YCQLExpressionGenerator(globalState).generateConstant()));
    }



### The current implementation does not take into account the column's data type when generating values, which can lead to unrealistic or invalid data being inserted.


#### **Need Improment**
- Values generated for each column should match the column's data type.
- The generated data should be realistic and valid for the database schema.

PR #1155, opened 2025-03-18
improving the insertValue method in the YCQLInsertGenerator class

PradeepFSTdhane123:
improving the insertValue method in the YCQLInsertGenerator class- [#1153 ](https://github.com/sqlancer/sqlancer/issues/1153)

### Changes
- Updated the `insertValue` method to generate values that are more meaningful and appropriate for the column's data type.
- Added support for generating valid values for `INT`, `TEXT`, `BOOLEAN`, `FLOAT`, `UUID`, `TIMESTAMP`, and `BLOB` columns.

this solution -- Improves the quality of generated data, making testing more effective.

### DIFF
--- a/src/sqlancer/yugabyte/ycql/gen/YCQLInsertGenerator.java
+++ b/src/sqlancer/yugabyte/ycql/gen/YCQLInsertGenerator.java
@@ -57,9 +57,44 @@ public class YCQLInsertGenerator extends AbstractInsertGenerator<YCQLColumn> {
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


mrigger:
It seems you essentially inlined a method call. See https://github.com/sqlancer/sqlancer/blob/7fbfe3d091a6254644dc0f4457e296ada94fe03e/src/sqlancer/yugabyte/ycql/gen/YCQLExpressionGenerator.java#L83.

PR #1166, opened 2025-03-19
Resolves #1163 Add function that supports query index dynamically

### LINKED ISSUE #1163: Query index information for DuckDB dynamically
DuckDB's index generator currently hardcodes the index names:

https://github.com/sqlancer/sqlancer/blob/7fbfe3d091a6254644dc0f4457e296ada94fe03e/src/sqlancer/duckdb/gen/DuckDBIndexGenerator.java#L26

This information can now be dynamically queried:
```
SELECT * FROM duckdb_indexes();
```

Thus, it would be helpful to maintain information about indexes in the schema classes similar to other database system implementations in SQLancer and query this information dynamically:

https://github.com/sqlancer/sqlancer/blob/7fbfe3d091a6254644dc0f4457e296ada94fe03e/src/sqlancer/mysql/gen/datadef/MySQLIndexGenerator.java#L50

yaojiejia:
If I understand correctly, I just need to remove hard-coded index and use
```globalState.getSchema().getFreeIndexName()``` instead
Tests ran:
- [x] ```mvn package -DskipTests```
- [x] ```mvn formatter:format```
- [x] ```mvn verify -DskipTests=true```
- [x] ```mvn -Dtest=TestDuckDBTLP test```
- [x] ```mvn -Dtest=TestDuckDBNoREC test```

I don't think duckdb supports b-tree and hash index for the index type therefore I didn't include them in here

### DIFF
--- a/src/sqlancer/duckdb/gen/DuckDBIndexGenerator.java
+++ b/src/sqlancer/duckdb/gen/DuckDBIndexGenerator.java
@@ -23,7 +23,7 @@ public final class DuckDBIndexGenerator {
             sb.append("UNIQUE ");
         }
         sb.append("INDEX ");
-        sb.append(Randomly.fromOptions("i0", "i1", "i2", "i3", "i4")); // cannot query this information
+        sb.append(globalState.getSchema().getFreeIndexName());
         sb.append(" ON ");
         DuckDBTable table = globalState.getSchema().getRandomTable(t -> !t.isView());
         sb.append(table.getName());


yaojiejia:
Hmm strange! I ran all the tests locally and they seem to be fine, will investigate further

mrigger:
It seems like this triggered a bug in DuckDB. You could try to find a reduced version of the bug-inducing test case, see whether it still reproduces on the latest DuckDB version, and then report it to the developers.

mrigger:
We will also need to add the indexes to the internal schema representation. See https://github.com/sqlancer/sqlancer/blob/7fbfe3d091a6254644dc0f4457e296ada94fe03e/src/sqlancer/duckdb/DuckDBSchema.java#L222

mrigger:
There is another PR that already addresses this: https://github.com/sqlancer/sqlancer/pull/1171

Did you already continue on this PR? Otherwise, I'd be inclined to focus on reviewing and merging the other PR, as it is more comprehensive despite you being first. What's your thought on this?

yaojiejia:
> There is another PR that already addresses this: #1171
> 
> Did you already continue on this PR? Otherwise, I'd be inclined to focus on reviewing and merging the other PR, as it is more comprehensive despite you being first. What's your thought on this?

Yeah I was trying to find out the root cause and trying to add the indexes to internal schema, but in the mean time I think you should look at the other PR and I will try to figure out what's going on in my end. Thanks!!

PR #1167, opened 2025-03-20
 add error handling in TiDBOracleFactory to prevent crashes

abdala-elgendy:
this issue is related https://github.com/sqlancer/sqlancer/pull/1127 after crashes when adding unit tests to Randomly class 
and this change makes the rowCountParser more robust by:
Proper error handling
And makes the queryPlanParser more robust by:
Adding null/empty checks
Converting operation names to lowercase for consistency
Proper handling of TiDB's specific operation format


mrigger:
I do not understand what the issue is that this change aims to address. Could you perhaps explain this based on an example?

abdala-elgendy:
> I do not understand what the issue is that this change aims to address. Could you perhaps explain this based on an example?

image
there is a wrong in TiDBCERT after adding unit tests to randomly (but tests now pass so it isn't related to tests randomly and The problem is in TiDB's execution plan parsing, where different row count formats weren't properly handled  )

PR #1169, opened 2025-03-20
Add option to enable/disable rowid columns in DuckDB tables

Saahi30:
This PR adds a configuration option to control whether rowid columns are automatically added to DuckDB tables that don't have a primary key. Currently, rowid columns are added unconditionally, which may not be desirable in all use cases.

## Approach
My  implementation followed these steps:
1. First, I identified that the rowid column was being added unconditionally in `DuckDBSchema.getTableColumns()`
2. I noticed that there was already a `testRowid` option in `DuckDBOptions` that was being used in other places (like `DuckDBIndexGenerator`)
3. I needed to make the global state accessible in the schema to check this option
4. I modified the `SQLConnection` class to store the global state, following the pattern used in other parts of the codebase
5. Finally, I updated the schema to conditionally add the rowid column based on the option

### 1. Added global state to SQLConnection
```java
// src/sqlancer/SQLConnection.java
public class SQLConnection implements SQLancerDBConnection {
    private final Connection connection;
    private GlobalState<?, ?, ?> globalState;  // Added field

    // Added getter and setter
    public void setGlobalState(GlobalState<?, ?, ?> globalState) {
        this.globalState = globalState;
    }

    public GlobalState<?, ?, ?> getGlobalState() {
        return globalState;
    }
    // ... rest of the class
}
```

### 2. Updated DuckDBProvider to set global state
```java
// src/sqlancer/duckdb/DuckDBProvider.java
@Override
public SQLConnection createDatabase(DuckDBGlobalState globalState) throws SQLException {
    // ... existing connection setup code ...
    SQLConnection sqlConn = new SQLConnection(conn);
    sqlConn.setGlobalState(globalState);  // Added this line
    return sqlConn;
}
```

### 3. Modified DuckDBSchema to use the option
```java
// src/sqlancer/duckdb/DuckDBSchema.java
private static List<DuckDBColumn> getTableColumns(SQLConnection con, String tableName) throws SQLException {
    // ... existing column collection code ...
    if (columns.stream().noneMatch(c -> c.isPrimaryKey()) && !AbstractSchema.matchesViewName(tableName)) {
        // https://github.com/cwida/duckdb/issues/589
        // https://github.com/cwida/duckdb/issues/588
        DuckDBGlobalState globalState = (DuckDBGlobalState) con.getGlobalState();
        if (globalState.getDbmsSpecificOptions().testRowid) {  // Added condition
            columns.add(new DuckDBColumn("rowid", new DuckDBCompositeDataType(DuckDBDataType.INT, 4), false, false));
        }
    }
    return columns;
}
```

## Implementation Details
- The rowid column is only added when:
  - The table has no primary key
  - The table is not a view
  - The `testRowid` option is enabled (default: true)
- The rowid column is an INTEGER type with size 4
- The rowid column is not nullable and not a primary key

## Usage
The feature can be controlled using the existing `--test-rowid` command-line option:
```bash
# Enable rowid columns (default)
--test-rowid true

# Disable rowid columns
--test-rowid false
```

I believe this pr will accurately address the issue #1163


Vedant1parekh:
why you have updated this **eclipse files(org.eclipse.jdtt**)**, is it necessary to merge into the main?

src/sqlancer/duckdb/DuckDBProvider.java
```diff
@@ -153,7 +153,9 @@ public SQLConnection createDatabase(DuckDBGlobalState globalState) throws SQLExc
         Statement stmt = conn.createStatement();
         stmt.execute("PRAGMA checkpoint_threshold='1 byte';");
         stmt.close();
-        return new SQLConnection(conn);
+        SQLConnection sqlConn = new SQLConnection(conn);
+        sqlConn.setGlobalState(globalState);
+        return sqlConn;
```
Vedant1parekh, line 158: Add proper variable name

src/sqlancer/duckdb/DuckDBSchema.java
```diff
@@ -272,8 +273,10 @@ private static List<DuckDBColumn> getTableColumns(SQLConnection con, String tabl
         if (columns.stream().noneMatch(c -> c.isPrimaryKey()) && !AbstractSchema.matchesViewName(tableName)) {
             // https://github.com/cwida/duckdb/issues/589
             // https://github.com/cwida/duckdb/issues/588
-            // TODO: implement an option to enable/disable rowids
-            columns.add(new DuckDBColumn("rowid", new DuckDBCompositeDataType(DuckDBDataType.INT, 4), false, false));
+            DuckDBGlobalState globalState = (DuckDBGlobalState) con.getGlobalState();
```
Vedant1parekh, line 276: why initializing two time globalStat?

PR #1171, opened 2025-03-20
[DuckDB] feat: Query index information for DuckDB dynamically

ahmed-kamal2004:
This PR resolves issue #1163

Before change
-
https://github.com/sqlancer/sqlancer/blob/7fbfe3d091a6254644dc0f4457e296ada94fe03e/src/sqlancer/duckdb/gen/DuckDBIndexGenerator.java#L26
The indexes names are hard-coded, there is no way AFAIK to store or retrieve them

After change
-
+ store the indexes information in the table schema
+ extend  SQLancer ability to generate random indexes for DuckDB testing, this ensures unique indexes names up to 100 index_name, now indexes names generation strategy is limited by `‎AbstractSchema::getFreeIndexName` only.

Snippet of the generated creation statements for Indexes
Screenshot from 2025-03-20 15-19-36
Screenshot from 2025-03-20 15-18-47

Files Changed 
-
+ DuckDBIndexGenerator.java : [OLD-Remved] random selection of indexes names out of a defined set of names - [NEW-Added] following generation strategy implemented & used by other database engines implementations
+ DuckDBSchema.java : [NEW-Added] store indexes information inside `DuckDBTable` object for further using 

Testing
-
+ Running for almost 2.5 million queries on DuckDB 
Screenshot from 2025-03-20 15-49-13
command used `java -jar sqlancer-*.jar --num-threads 4 duckdb --oracle NOREC `

### DIFF
--- a/src/sqlancer/duckdb/DuckDBSchema.java
+++ b/src/sqlancer/duckdb/DuckDBSchema.java
@@ -215,23 +215,12 @@ public class DuckDBSchema extends AbstractSchema<DuckDBGlobalState, DuckDBTable>
         return new DuckDBCompositeDataType(primitiveType, size);
     }
 
-    public static class DuckDBTable extends AbstractRelationalTable<DuckDBColumn, DuckDBIndex, DuckDBGlobalState> {
+    public static class DuckDBTable extends AbstractRelationalTable<DuckDBColumn, TableIndex, DuckDBGlobalState> {
 
-        public DuckDBTable(String tableName, List<DuckDBColumn> columns, List<DuckDBIndex> indexes, boolean isView) {
+        public DuckDBTable(String tableName, List<DuckDBColumn> columns, List<TableIndex> indexes, boolean isView) {
             super(tableName, columns, indexes, isView);
         }
 
-    }
-
-    public static final class DuckDBIndex extends TableIndex {
-
-        private DuckDBIndex(String indexName) {
-            super(indexName);
-        }
-
-        public static DuckDBIndex create(String indexName) {
-            return new DuckDBIndex(indexName);
-        }
     }
 
     public static DuckDBSchema fromConnection(SQLConnection con, String databaseName) throws SQLException {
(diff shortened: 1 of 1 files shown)


src/sqlancer/duckdb/DuckDBSchema.java
```diff
@@ -216,14 +215,25 @@ private static DuckDBCompositeDataType getColumnType(String typeString) {
...
 
-        public DuckDBTable(String tableName, List<DuckDBColumn> columns, boolean isView) {
-            super(tableName, columns, Collections.emptyList(), isView);
+        public DuckDBTable(String tableName, List<DuckDBColumn> columns, List<DuckDBIndex> indexes, boolean isView) {
+            super(tableName, columns, indexes, isView);
         }
 
     }
 
+    public static final class DuckDBIndex extends TableIndex {
```
mrigger, line 226: I think we could actually just use the `TableIndex` class here instead.

src/sqlancer/duckdb/DuckGlobalState.java
```diff
@@ -0,0 +1,10 @@
+package sqlancer.duckdb;
+
+import sqlancer.SQLGlobalState;
+
+public class DuckGlobalState extends SQLGlobalState<DuckDBOptions, DuckDBSchema> {
```
mrigger, line 5: We already have a global state class for `DuckDB`.

mrigger (commented):
Thanks for the PR! I have added some improvement suggestions.

mrigger:
Thanks a lot for the PR! It looks good to me, but it seems the tests are currently failing due to a test oracle report. Could you perhaps try reducing the bug-inducing test cases? It might be a bug in DuckDB.

mrigger:
Based on the current output, it seems that we are repeatedly generating indexes with the same name, so it seems like there might be another issue.

mrigger:
This functionality is already present in SQLancer and also other DBMSs use it. In the `DuckDBIndexGenerator`, when instantiating the `SQLQueryAdapter`, you can see that an argument is set that indicates a schema change.

PR #1181, opened 2025-03-22
Support CODDTest for SQLite

DerZc:
Support CODDTest for SQLite. This is a sub-task of the larger implementation tracked in #1054.


.settings/org.eclipse.jdt.core.prefs
```diff
@@ -11,9 +11,9 @@ org.eclipse.jdt.core.compiler.annotation.nullable.secondary=
 org.eclipse.jdt.core.compiler.annotation.nullanalysis=disabled
```
mrigger, line 11: This is a temporary file, so we should avoid checking it in.

DerZc:
Thank you for your feedback! I’ll make these updates as soon as possible.

.settings/org.eclipse.jdt.core.prefs
DerZc: So sorry for this and I have addressed this now.

src/sqlancer/common/schema/AbstractTables.java
```diff
@@ -34,4 +34,28 @@ public String columnNamesAsString(Function<C, String> function) {
...
+        if (!this.tables.contains(table)) {
+            this.tables.add(table);
+            columns.addAll(table.getColumns());
+        }
+    }
+
+    public void removeTable(T table) {
+        if (this.tables.contains(table)) {
+            this.tables.remove(table);
+            for (C c : table.getColumns()) {
```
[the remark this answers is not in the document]
DerZc: I don't think it's necessary for CODDTest. I introduced the `removeTable` method because, when using a join clause in the Original query, we need the same clause in the Auxiliary query. Since the Auxiliary query is generated first, the join clauses are fixed when generating the Original query. To avoid redundant tables in the `FROM` clause, I remove them accordingly.
While it's not essential to remove columns in CODDTest—since they're used to generate other expressions—it might still be better to do so. If `removeTable` is reused by other components, it could leave behind columns in `WHERE` or other clauses that no longer belong to any table.

src/sqlancer/common/schema/AbstractTables.java
```diff
@@ -34,4 +34,28 @@ public String columnNamesAsString(Function<C, String> function) {
...
+                columns.remove(c);
+            }
+        }
+    }
+
+    public Boolean isContained(T table) {
```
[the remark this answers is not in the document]
DerZc: I have addressed this.

src/sqlancer/sqlite3/SQLite3Options.java
```diff
@@ -79,6 +79,9 @@ public class SQLite3Options implements DBMSSpecificOptions<SQLite3OracleFactory>
     @Parameter(names = { "--max-num-indexes" }, description = "The maximum number of indexes that can be created")
     public int maxNumIndexes = 20;
 
+    @Parameter(names = { "--coddtest-model" }, description = "Apply CODDTest on expression, subquery, or random")
+    public String coddTestModel = "random";
```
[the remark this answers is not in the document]
DerZc: Yes, this will be better and I updated this.

src/sqlancer/sqlite3/SQLite3ToStringVisitor.java
```diff
@@ -139,7 +156,7 @@ public void visit(SQLite3Select s, boolean inner) {
             visit(whereClause);
             sb.append(")");
         }
-        if (!s.getGroupByClause().isEmpty()) {
+        if (s.getGroupByClause().size() > 0) {
```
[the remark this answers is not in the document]
DerZc: Yes, I honestly can't remember why I made that change at the time, even though they look the same. I have addressed this now.

src/sqlancer/sqlite3/ast/SQLite3Select.java
```diff
@@ -161,4 +164,33 @@ public SQLite3Expression getHavingClause() {
...
+
+    public void updateWithClauseRight(SQLite3Expression withClauseRight) {
+        this.withClause.updateRight(withClauseRight);
+    }
+
+    public SQLite3Expression getWithClause() {
+        return this.withClause;
+    }
+
+    public void replaceFromTable(String tableName, SQLite3Expression newFromExpression) {
```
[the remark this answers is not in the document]
DerZc: I add a short comment fot this method.

src/sqlancer/sqlite3/schema/SQLite3DataType.java
```diff
@@ -1,6 +1,23 @@
 package sqlancer.sqlite3.schema;
 
+import sqlancer.IgnoreMeException;
+
 public enum SQLite3DataType {
     NULL, INT, TEXT, REAL, NONE, BINARY;
 
+    public static SQLite3DataType getTypeFromName(String name) {
```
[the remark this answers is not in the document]
DerZc: I tried to use `valueOf` but found some name are different from type, for example `integer` and `INT`

src/sqlancer/sqlite3/SQLite3ToStringVisitor.java
```diff
@@ -482,4 +509,127 @@ public void visit(SQLite3SetClause set) {
...
+                    sb.append(", ");
+                }
+                if (vs.get(name).get(i).getDataType() == SQLite3DataType.NULL) {
+                    visit(vs.get(name).get(i));
+                } else {
+                    sb.append("(CAST(");
+                    visit(vs.get(name).get(i));
+                    sb.append(" AS ");
+                    switch(vs.get(name).get(i).getDataType()) {
+                    case BINARY:
```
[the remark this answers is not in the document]
DerZc: Yes! I have addressed this.

src/sqlancer/sqlite3/SQLite3ToStringVisitor.java
```diff
@@ -482,4 +509,127 @@ public void visit(SQLite3SetClause set) {
...
+
+    @Override
+    public void visit(SQLite3Typeof expr) {
+        sb.append("typeof(");
+        visit(expr.getInnerExpr());
+        sb.append(")");
+    }
+
+    @Override
+    public void visit(SQLite3ResultMap tableSummary) {
```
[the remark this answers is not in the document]
DerZc: I removed the code handling cases where an expression returns no results under the current database state, since I now discard the expression when it has no results. To ensure safety in case this method is used by other components, I’ve updated it to throw an exception if the expression yields an empty result.

src/sqlancer/sqlite3/SQLite3Options.java
```diff
@@ -79,6 +79,21 @@ public class SQLite3Options implements DBMSSpecificOptions<SQLite3OracleFactory>
     @Parameter(names = { "--max-num-indexes" }, description = "The maximum number of indexes that can be created")
     public int maxNumIndexes = 20;
 
+    public enum coddtest_model {
```
mrigger, line 82: We should use Java naming conventions like "CODDTestModel" here and below: `RANDOM`, `EXPRESSION`,  and `SUBQUERY`.

src/sqlancer/sqlite3/SQLite3ToStringVisitor.java
```diff
@@ -482,4 +508,104 @@ public void visit(SQLite3SetClause set) {
...
+                if (vs.get(name).get(i).getDataType() == SQLite3DataType.NULL) {
+                    visit(vs.get(name).get(i));
+                } else {
+                    sb.append("(CAST(");
+                    visit(vs.get(name).get(i));
+                    sb.append(" AS ");
+                    sb.append(vs.get(name).get(i).getDataType().toString());
+                    sb.append("))");
+                }
+                isFirstColumn = true;
```
mrigger, line 555: Is this intended? I think this should be `false`?

src/sqlancer/common/schema/AbstractTables.java
```diff
@@ -34,4 +34,28 @@ public String columnNamesAsString(Function<C, String> function) {
         return getColumns().stream().map(function).collect(Collectors.joining(", "));
     }
 
+
+    public void addTable(T table) {
```
mrigger, line 38: Ideally, we would also add test cases later on for these new methods.

src/sqlancer/sqlite3/SQLite3Options.java
DerZc: Updated!

src/sqlancer/sqlite3/SQLite3ToStringVisitor.java
DerZc: So sorry I make an error here, and fix it now

src/sqlancer/common/schema/AbstractTables.java
DerZc: I add the unit tests for these new methods and an unit tests for SQLite3 CODDTest oracle.

src/sqlancer/sqlite3/oracle/SQLite3CODDTestOracle.java
```diff
@@ -0,0 +1,971 @@
...
+    public SQLite3CODDTestOracle(SQLite3GlobalState globalState) {
+        super(globalState);
+        this.s = globalState.getSchema();
+        SQLite3Errors.addExpectedExpressionErrors(errors);
+        SQLite3Errors.addMatchQueryErrors(errors);
+        SQLite3Errors.addQueryErrors(errors);
+        errors.add("misuse of aggregate");
+        errors.add("misuse of window function");
+        errors.add("second argument to nth_value must be a positive integer");
+        errors.add("no such table");
```
mrigger, line 87: It's surprising that such an error could happen. Do you know why?

src/sqlancer/sqlite3/oracle/SQLite3CODDTestOracle.java
```diff
@@ -0,0 +1,971 @@
...
+        errors.add("no query solution");
+        errors.add("unable to use function MATCH in the requested context");
+        errors.add("[SQLITE_ERROR] SQL error or missing database (unrecognized token:");
```
mrigger, line 90: Does this indicate a syntax error?

mrigger:
The CI tests currently fail (see `mvn -B verify -DskipTests=true`). After fixing them and checking that the CI tests run CODDTest, I think we can merge the PR. I find the test oracle implementation quite complex and think there would be room for simplifying it, but it would be better to first get the PR in.

mrigger:
> Error:  Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.20.0:validate (eclipseformat) on project sqlancer: File '/home/runner/work/sqlancer/sqlancer/src/sqlancer/hive/HiveToStringVisitor.java' has not been previously formatted. Please format file (for example by invoking `mvn net.revelc.code.formatter:formatter-maven-plugin:2.20.0:format`) and commit before running validation! -> [Help 1]

Strange, perhaps this is an issue that made it in with another PR. Could you please run `mvn formatter:format` nevertheless so that the general tests pass (again)?

mrigger:
For the tests to run, I think you'll need to add them here:

https://github.com/sqlancer/sqlancer/blob/3863169037a34b01ceddb5152e4237ace8ceadac/.github/workflows/main.yml#L569-L571

mrigger (approved):
Thanks, lgtm!

PR #1192, opened 2025-03-27
improve handling optimizer switches in MySQL

abdala-elgendy:
**I have improved Optimizer switch organization in MySQL:**
by grouping optimizer switches by category (index, join, subquery, other)
and adding systematic handling of known bugs that may occur cause of wrong configurations.

There is more handling to do in Optimizer Switch in MySQL


mrigger:
Thanks for the PR! I think the change is not desirable, as it reduces the current search space if we enforce that one optimization flag is present from each category.

abdala-elgendy:
I know that because I categorize each category

PR #1194, opened 2025-03-30
TODO: throw an exception instead on testInteger

SharmaNishchay:


SharmaNishchay:
@mrigger The checks are passing. Please review the changes. Thanks!

bhavaniprasad-tech (commented):
@Test
public void testIntegerValidRange() {
    Randomly r = new Randomly();
    for (int i = 0; i < 100; i++) {
        int val = r.getInteger(5, 10);
        assertTrue(val >= 5 && val <= 10);
    }
}

bhavaniprasad-tech:
Add input validation for getInteger() and update corresponding test

- Added a check in Randomly.getInteger(left, right) to throw IllegalArgumentException when left > right
- Updated testInteger() in TestRandomly.java to assert this exception
- Ensures safer and more predictable behavior for invalid input ranges

SharmaNishchay:
> Add input validation for getInteger() and update corresponding test
> 
> * Added a check in Randomly.getInteger(left, right) to throw IllegalArgumentException when left > right
> * Updated testInteger() in TestRandomly.java to assert this exception
> * Ensures safer and more predictable behavior for invalid input ranges

I think the getLong2() method already handles that scenario.

PR #1201, opened 2025-04-02
[Hive] Support TLP oracle

liyxiris:
This PR creates a basic framework for Hive SQL generation and uses TLP oracle to test Hive.
- Implements JDBC connection to Hive cluster.
- Implements creating Hive tables with basic column types, i.e. BOOLEAN, INT, DOUBLE, STRING.
- Implements some Hive table/column constraints, e.g. PRIMARY_KEY_DISABLE, CHECK, NOT_NULL, etc.
- Implements common unary, binary and other expressions.
- Implements row insertion.
- Implement SELECT WHERE clause for TLP. (joins and aggregations are on my TODO list)


mrigger:
Thanks! Will try to review as soon as possible.

mrigger:
It seems the CI cannot resolve the Maven dependency:
> Error:  Failed to execute goal on project sqlancer: Could not resolve dependencies for project com.sqlancer:sqlancer:jar:2.0.0
Error:  dependency: jdk.tools:jdk.tools:jar:1.6

src/sqlancer/hive/HiveErrors.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private HiveErrors() {
+    }
+
+    public static List<String> getExpressionErrors() {
+        ArrayList<String> errors = new ArrayList<>();
+
+        errors.add("cannot recognize input near");
+        errors.add("Argument type mismatch");
+        errors.add("Error while compiling statement");
```
mrigger, line 18: Is this an expected error or could it be an actual bug (e.g., like an internal error)?

src/sqlancer/hive/HiveSchema.java
```diff
@@ -0,0 +1,114 @@
...
+    public HiveSchema(List<HiveTable> databaseTables) {
+        super(databaseTables);
+    }
+
+    public static HiveSchema fromConnection(SQLConnection con, String databaseName) throws SQLException {
+        // TODO: discover built-in tables
+        List<HiveTable> databaseTables = new ArrayList<>();
+        List<String> tableNames = getTableNames(con);
+        for (String tableName : tableNames) {
+            if (DBMSCommon.matchesIndexName(tableName)) {
```
mrigger, line 63: I think this might be a copy of DuckDB, or did you encounter a table name that matches an index during testing?

mrigger: If not, we should remove the if block.

src/sqlancer/hive/HiveSchema.java
```diff
@@ -0,0 +1,114 @@
...
+                    HiveColumn c = new HiveColumn(columnName, null, getColumnType(dataType.toUpperCase()));
+                    columns.add(c);
+                }
+            }
+        }
+        return columns;
+    }
+
+    private static HiveDataType getColumnType(String typeString) {
+        if (typeString.startsWith("DECIMAL")) {
```
mrigger, line 105: I think we could omit the if block here?

src/sqlancer/hive/ast/HiveConstant.java
```diff
@@ -0,0 +1,215 @@
...
+            return value;
+        }
+
+        @Override
+        public String toString() {
+            return String.valueOf(value);
+        }
+    }
+
+    // public static class HiveFloatConstant extends HiveConstant {
```
mrigger, line 44: I think it would be better to remove commented-out code.

src/sqlancer/hive/gen/HiveExpressionGenerator.java
```diff
@@ -0,0 +1,334 @@
...
+
+    private final HiveGlobalState globalState;
+
+    private enum Expression {
+        // TODO: add or delete expressions.
+        UNARY_PREFIX, UNARY_POSTFIX, BINARY_COMPARISON, BINARY_LOGICAL, BINARY_ARITHMETIC,
+        CAST, FUNC, BETWEEN, IN, CASE;
+    }
+
+    private enum ConstantType {
```
mrigger, line 36: Can we just use the enum from `HiveSchema` here?

src/sqlancer/hive/gen/HiveTableGenerator.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public static SQLQueryAdapter generate(HiveGlobalState globalState, String tableName) {
+        HiveTableGenerator generator = new HiveTableGenerator(globalState, tableName);
+        return generator.create();
+    }
+
+    private SQLQueryAdapter create() {
+        ExpectedErrors errors = new ExpectedErrors();
+
+        sb.append("CREATE ");
```
mrigger, line 51: I guess we can combine the `CREATE` and `TABLE` into one string here.

src/sqlancer/hive/gen/HiveTableGenerator.java
```diff
@@ -0,0 +1,125 @@
...
+            case NOT_NULL:
+                sb.append(" NOT NULL");
+                break;
+            case DEFAULT:
+                // TODO: add errors
+                sb.append(" DEFAULT (");
+                sb.append(HiveToStringVisitor.asString(gen.generateConstant()));
+                sb.append(")");
+            case CHECK:
+                errors.add("Invalid Constraint syntax");
```
mrigger, line 114: I think we should always generate constraints with a valid syntax. Or is there a justified case here?

src/sqlancer/hive/oracle/HiveTLPBase.java
```diff
@@ -0,0 +1,68 @@
...
+import sqlancer.hive.HiveSchema.HiveColumn;
+import sqlancer.hive.HiveSchema.HiveTable;
+import sqlancer.hive.HiveSchema.HiveTables;
+import sqlancer.hive.ast.HiveColumnReference;
+import sqlancer.hive.ast.HiveExpression;
+import sqlancer.hive.ast.HiveSelect;
+import sqlancer.hive.ast.HiveTableReference;
+import sqlancer.hive.gen.HiveExpressionGenerator;
+
+public class HiveTLPBase extends TernaryLogicPartitioningOracleBase<HiveExpression, HiveGlobalState>
```
mrigger, line 23: We refactored the TLP classes recently to avoid code duplication. Could you please look into the new way in how we implement this?

mrigger (commented):
Thanks, this looks great! Sorry for the delay in reviewing, I was traveling. I added some comments to improve the PR, but I believe we can merge the change soon.

src/sqlancer/hive/gen/HiveExpressionGenerator.java
liyxiris: `ConstantType` has an additional NULL type compared to `HiveDataType`. But you are right, no need to use two enums for similar purposes.

src/sqlancer/hive/oracle/HiveTLPBase.java
liyxiris: Thanks for mentioning! I've refactored these code using the new generic oracle.

src/sqlancer/hive/HiveErrors.java
mrigger: Just a quick follow up question on this: for an untyped expression generator, I would expected semantically invalid statements, but not syntactically invalid ones. Would it be possible to show an example where this causes such an error?

mrigger (commented):
Thanks for the follow-up changes and sorry for the delayed response. I have another follow-up questions on the check constraints. Besides this, the PR looks ready to be merged.

mrigger:
Just to double-check: this is a contribution independent from the GSoC program, right? Just want to be sure that we do not overlook your proposal in case you submitted one for GSoC.

liyxiris:
Right, this PR is not part of the GSoC program. It’s just pure interest ​​:)

src/sqlancer/hive/HiveErrors.java
liyxiris: type mismatch
wrong argument
Those are error details. They are showed as Hive SemanticException but actually are all caused by mismatched types. So I believe maybe they are expected errors when using untyped expression generator?

mrigger: Thanks for responding. Yes, Agreed! Does the "Invalid Constraint syntax" below also relate to an actual semantic error?

liyxiris: Oh thanks for mentioning, "Invalid Constraint syntax" also relates to mismatched types, but I believe it's not needed anymore since it will be covered by the above HiveErrors. It's removed in the latest commit.

liyxiris:
> It seems the CI cannot resolve the Maven dependency:
> 
> > Error:  Failed to execute goal on project sqlancer: Could not resolve dependencies for project com.sqlancer:sqlancer:jar:2.0.0
> > Error:  dependency: jdk.tools:jdk.tools:jar:1.6

This problem is a tricky one. It seems related with JDK version but I haven't encountered it on my PC using the same JDK 11. I'm trying to solve it by upgrading Hive dependency versions (3.1.3 -> 4.0.1) because this seems the only change are related with JDK tools.

mrigger (approved):
Thanks a lot again for your contribution. The code looks good to me. The CI tests also run successfully.

SQLancer seems to be finding bugs in Apache Hive. Did you already report any of these? If so, would it be possible to disable the bug-inducing patterns for now (see https://github.com/sqlancer/sqlancer/blob/main/CONTRIBUTING.md#unfixed-bugs)? This would prevent the CI test from exiting with a non-zero code.

Here is one example of a bug-inducing test reported by SQLancer:

```sql
CREATE TABLE database4.t0 (c0 BOOLEAN NOT NULL);
INSERT INTO t0 VALUES (true);
SELECT * FROM t0; -- cardinality: 1
SELECT * FROM t0 WHERE (t0.c0 IN ((((('q?[Tw5')=(t0.c0)))<=(((t0.c0)=(t0.c0)))))) UNION ALL SELECT * FROM t0 WHERE (NOT (t0.c0 IN ((((('q?[Tw5')=(t0.c0)))<=(((t0.c0)=(t0.c0))))))) UNION ALL SELECT * FROM t0 WHERE (((t0.c0 IN ((((('q?[Tw5')=(t0.c0)))<=(((t0.c0)=(t0.c0))))))) IS NULL); -- cardinality: 0;
```

mrigger:
Merging this first as I think having support for Hive as well as the CI tests is already very valuable. Thanks a lot for contributing the code!

As a next step, we should make sure to make the CI tests green so that we will not accidentally break the implementation in the future.

liyxiris:
@mrigger Sorry for the late reply. And really thank you for merging this! I plan to continue supporting more oracles for Hive. 

About the bugs it found, I haven't reported them yet but will soon do so, and I'm also going to disable these patterns in the code to avoid generating similar bugs.

PR #1202, opened 2025-04-02
[PostgreSQL]Update: Complete Implementation of RESET generation in Postgres

codebreaker32:
I have completed the `RESET` usage in `PostgresProvider.java` by 
- Creating a dedicated class for RESET generation i.e `PostgresResetGenerator.java`
- I implemented this class by mimicking `PostgresSetGenerator.java` 

Previously, It only supported `RESET ALL` for all queries.

@mrigger

### DIFF
--- a/src/sqlancer/postgres/gen/PostgresResetGenerator.java
+++ b/src/sqlancer/postgres/gen/PostgresResetGenerator.java
@@ -1,0 +1,73 @@
+package sqlancer.postgres.gen;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import sqlancer.Randomly;
+import sqlancer.common.query.SQLQueryAdapter;
+import sqlancer.postgres.PostgresGlobalState;
+
+public final class PostgresResetGenerator {
+    private PostgresResetGenerator() {
+    }
+
+    public enum DynamicConfigOption {
+        SYNCHRONOUS_COMMIT("synchronous_commit"), WAL_COMPRESSION("wal_compression"), COMMIT_DELAY("commit_delay"),
+        COMMIT_SIBLINGS("commit_siblings"),
+
+        TRACK_ACTIVITIES("track_activities"), TRACK_COUNTS("track_counts"), TRACK_IO_TIMING("track_io_timing"),
+        TRACK_FUNCTIONS("track_functions"),
+
+        VACUUM_FREEZE_TABLE_AGE("vacuum_freeze_table_age"), VACUUM_FREEZE_MIN_AGE("vacuum_freeze_min_age"),
+        VACUUM_MULTIXACT_FREEZE_TABLE_AGE("vacuum_multixact_freeze_table_age"),
+        VACUUM_MULTIXACT_FREEZE_MIN_AGE("vacuum_multixact_freeze_min_age"),
+
+        ENABLE_BITMAPSCAN("enable_bitmapscan"), ENABLE_GATHERMERGE("enable_gathermerge"),
+        ENABLE_HASHJOIN("enable_hashjoin"), ENABLE_INDEXSCAN("enable_indexscan"),
+        ENABLE_INDEXONLYSCAN("enable_indexonlyscan"), ENABLE_MATERIAL("enable_material"),
+        ENABLE_MERGEJOIN("enable_mergejoin"), ENABLE_NESTLOOP("enable_nestloop"),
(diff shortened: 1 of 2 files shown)


codebreaker32:
Please review @mrigger

codebreaker32:
Please review @mrigger

src/sqlancer/postgres/gen/PostgresResetGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public static SQLQueryAdapter create(PostgresGlobalState globalState) {
+        StringBuilder sb = new StringBuilder();
+        ArrayList<PostgresResetGenerator.DynamicConfigOption> options = new ArrayList<>(
+                Arrays.asList(PostgresResetGenerator.DynamicConfigOption.values()));
+        PostgresResetGenerator.DynamicConfigOption option = Randomly.fromList(options);
+        sb.append("RESET ");
+        if (Randomly.getBoolean()) {
+            sb.append(option.configName);
```
mrigger, line 66: Here, we could actually choose a random subset rather than a single value.

src/sqlancer/postgres/gen/PostgresResetGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+import sqlancer.Randomly;
+import sqlancer.common.query.SQLQueryAdapter;
+import sqlancer.postgres.PostgresGlobalState;
+
+public final class PostgresResetGenerator {
+    private PostgresResetGenerator() {
+    }
+
+    public enum DynamicConfigOption {
+        SYNCHRONOUS_COMMIT("synchronous_commit"), WAL_COMPRESSION("wal_compression"), COMMIT_DELAY("commit_delay"),
```
mrigger, line 15: It seems that this is essentially the same list as in `PostgresSetGenerator`. Can we re-use that implementation?

codebreaker32: But then I have to make that class public

src/sqlancer/postgres/gen/PostgresResetGenerator.java
codebreaker32: Why? I mean Postgres doesn't support multiple parameters with `RESET` or you are asking for something else?

src/sqlancer/postgres/gen/PostgresResetGenerator.java
codebreaker32: In that case, should I make it package-private or keep both generator functions within the same class? Personally, I feel making it package-private would be a better approach. What do you think?
@mrigger

codebreaker32: In that case, should I make it package-private or keep both generator functions within the same class? Personally, I feel making it package-private would be a better approach. What do you think?
@mrigger

src/sqlancer/postgres/gen/PostgresResetGenerator.java
codebreaker32: Why? I mean Postgres doesn't support multiple parameters with `RESET` 
@mrigger

src/sqlancer/postgres/gen/PostgresResetGenerator.java
```diff
@@ -0,0 +1,28 @@
...
+import sqlancer.common.query.SQLQueryAdapter;
+import sqlancer.postgres.PostgresGlobalState;
+
+public final class PostgresResetGenerator {
+    private PostgresResetGenerator() {
+    }
+
+    public static SQLQueryAdapter create(PostgresGlobalState globalState) {
+        StringBuilder sb = new StringBuilder();
+        ArrayList<PostgresSetGenerator.ConfigurationOption> options = new ArrayList<>(
```
mrigger, line 16: I think we can simplify this logic using https://github.com/sqlancer/sqlancer/blob/30e10419d1ea85c6e03ab1517496cf3faa1e9468/src/sqlancer/Randomly.java#L116

mrigger (commented):
Thanks, this looks almost ready to be merged.

codebreaker32:
Fixed the merge conflicts and made the changes you suggested
@mrigger

codebreaker32:
I have made suggested changes
@mrigger

codebreaker32:
Please review @mrigger

PR #1206, opened 2025-04-03
Fix CHECK Constraint Generation to Prevent Integer Overflow

Eklavya-Sahani:
fixes an issue in PostgresTableGenerator.java where the generated CHECK constraints could cause integer overflows or always evaluate to FALSE.

Issue:
The previous implementation generated random boolean expressions that sometimes resulted in "integer out of range" errors.

This prevented valid table creation in SQLancer tests.

Fix:
The CHECK constraint now ensures values stay within a safe integer range (-1000000000 to 1000000000).

Instead of random expressions, the constraint is directly applied to the column.

This prevents always-false constraints and improves test reliability.

### DIFF
--- a/src/sqlancer/Main.java
+++ b/src/sqlancer/Main.java
@@ -792,7 +792,6 @@ public final class Main {
                 lastNrQueries = currentNrQueries;
                 lastNrDbs = currentNrDbs;
             }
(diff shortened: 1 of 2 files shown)


mrigger:
Thanks for opening the PR. This change is not desirable as it restricts the search space.

PR #1207, opened 2025-04-05
[MySQL] Support generating CASE operators

EmilyOng:
This PR adds support for generating `CASE` operators, as described in the [documentation](https://dev.mysql.com/doc/refman/8.4/en/flow-control-functions.html#operator_case) for MySQL.

```sql
[CASE value WHEN compare_value THEN result [WHEN compare_value THEN result ...] [ELSE result] END]
[CASE WHEN condition THEN result [WHEN condition THEN result ...] [ELSE result] END]
```

## Tests

I tested the PR by running SQLancer on [MySQL version 8.0.36](https://hub.docker.com/layers/library/mysql/8.0.36/images/sha256-65ce0889751900d2dd5fc5aa5a5fb59073d401fc43df2eae6bd6afb18e7626dd), which is the [latest version of MySQL supported on SQLancer](https://github.com/sqlancer/sqlancer/blob/be3d3b5cd1de40acfef7adae19767dc1a1aab6e0/.github/workflows/main.yml#L409).

```bash
java -jar target/sqlancer-*.jar --username root --password "" --port 33012 --num-threads 5 mysql --oracle PQS
```

Screenshot 2025-04-13 at 11 28 49 AM


src/sqlancer/mysql/ast/MySQLCaseOperator.java
```diff
@@ -0,0 +1,48 @@
...
+public class MySQLCaseOperator extends NewCaseOperatorNode<MySQLExpression> implements MySQLExpression {
+
+    public MySQLCaseOperator(MySQLExpression switchCondition, List<MySQLExpression> whenExprs,
+            List<MySQLExpression> thenExprs, MySQLExpression elseExpr) {
+        super(switchCondition, whenExprs, thenExprs, elseExpr);
+    }
+
+    @Override
+    public MySQLConstant getExpectedValue() {
+        int nrConditions = getConditions().size();
```
mrigger, line 16: It's great that you made the effort to implement the PQS component as well, but I think it would also be fine, in principle, to just bail out (e.g., throw an `IgnoreMeException` or so).

mrigger: But if this works, it would be great to also merge this part.

mrigger (commented):
I see the PR is still marked as a draft, but it looks ready to be merged to me.

src/sqlancer/mysql/ast/MySQLCaseOperator.java
EmilyOng: Sure, thanks! I have added some tests to validate the `getExpectedValue` method and added a screenshot of a SQLancer run in the PR description.

mrigger (approved):
LGTM, thanks!

PR #1040, opened 2024-12-08
[Postgres][v13] Add PARALLEL option for VACUUM

robins:
This patch adds `PARALLEL` support to `VACUUM` command in Postgres (launched in v13)

This is working towards the Issue #912 .

- Sample Postgres ERROR logfile capture given below
```
lancer@lenovo:~/proj/lancerpg$ grep VACUUM logfile | grep PARALLEL
2024-12-08 17:39:51.042 ACDT [653315] LOG:  execute <unnamed>: VACUUM (PARALLEL 0)
2024-12-08 17:39:51.315 ACDT [653285] LOG:  execute <unnamed>: VACUUM (VERBOSE, PARALLEL 0) t5
2024-12-08 17:39:51.375 ACDT [653283] LOG:  execute <unnamed>: VACUUM (PARALLEL 0, VERBOSE 1)
2024-12-08 17:45:03.234 ACDT [673926] LOG:  execute <unnamed>: VACUUM (SKIP_LOCKED 0, PARALLEL 2)
2024-12-08 17:47:22.036 ACDT [702212] LOG:  execute <unnamed>: VACUUM (FULL 1, PARALLEL 0)
2024-12-08 17:48:07.203 ACDT [712462] LOG:  execute <unnamed>: VACUUM (PARALLEL 6) t5
```

- `mvn package -DskipTests` works
- `mvn package` fails with #799 .
```
[INFO] Running sqlancer.TestUsageNamingConvention
No DBMS implementations (i.e., instantiations of the DatabaseProvider class) were found. You likely ran into an issue described in https://github.com/sqlancer/sqlancer/issues/799. As a workaround, I now statically load all supported providers as of June 7, 2023.
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.057 s <<< FAILURE! - in sqlancer.TestUsageNamingConvention
[ERROR] testNonEmptyDescription  Time elapsed: 0.043 s  <<< ERROR!
java.lang.reflect.InaccessibleObjectException: Unable to make private static java.lang.annotation.Annotation java.lang.reflect.AnnotatedElement.lambda$getDeclaredAnnotationsByType$0(java.lang.annotation.Annotation,java.lang.annotation.Annotation) accessible: module java.base does not "opens java.lang.reflect" to unnamed module @1376c05c
        at sqlancer.TestUsageNamingConvention.testNonEmptyDescription(TestUsageNamingConvention.java:25)
```

PR #1208, opened 2025-04-05
Added PostgreSQL v13 VACUUM PARALLEL option support

guptapratykshh:
Here's a PR draft for the PostgreSQL v13 VACUUM PARALLEL implementation:

# PostgreSQL v13 VACUUM PARALLEL option support

## Description
This PR adds support for the PARALLEL option of VACUUM commands in PostgreSQL 13+. The PARALLEL option was introduced in PostgreSQL 13 and allows specifying number of parallel worker processes to use when running a parallel vacuum operation.

## Changes made
1. Added PostgreSQL version detection in `PostgresGlobalState.java` -
 Added a `majorVersion` field to store the server's major version, implemented `getMajorVersion()` method that extracts the major version from `server_version_num`and updated `setConnection()` to initialize the major version when connecting

2. Enhanced `PostgresVacuumGenerator.java` to support the PARALLEL option -
 Added PARALLEL option to the list of vacuum options when PostgreSQL version is 13 or higher , ensured PARALLEL option always has valid integer value (between 0-4 for testing) and added error handling for the "VACUUM option PARALLEL cannot be used with FULL" error

## Testing
Tested with PostgreSQL 13.20 to verify:
- Version detection works correctly
- VACUUM commands with PARALLEL option are properly generated
- The PARALLEL option always receives a valid integer value
- Error handling is appropriate for VACUUM PARALLEL constraints

## Related Issues
Resolves one of the pending items from #912
Related to #1040 "[Postgres][v13] Add PARALLEL option for VACUUM"

Screenshot 2025-04-06 at 3 54 19 AM

### DIFF
--- a/src/sqlancer/postgres/PostgresGlobalState.java
+++ b/src/sqlancer/postgres/PostgresGlobalState.java
@@ -27,6 +27,7 @@ public class PostgresGlobalState extends SQLGlobalState<PostgresOptions, Postgre
     // store and allow filtering by function volatility classifications
     private final Map<String, Character> functionsAndTypes = new HashMap<>();
     private List<Character> allowedFunctionTypes = Arrays.asList(IMMUTABLE, STABLE, VOLATILE);
+    private int majorVersion;
 
     @Override
     public void setConnection(SQLConnection con) {
@@ -36,6 +37,7 @@ public class PostgresGlobalState extends SQLGlobalState<PostgresOptions, Postgre
             this.operators = getOperators(getConnection());
             this.collates = getCollnames(getConnection());
             this.tableAccessMethods = getTableAccessMethods(getConnection());
+            this.majorVersion = getMajorVersion(getConnection());
         } catch (SQLException e) {
             throw new AssertionError(e);
         }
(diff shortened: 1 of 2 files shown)


guptapratykshh:
@robins @mrigger please look into this support that i added for PostgreSQL v13. Please review it.

guptapratykshh:
Can you please look into this PR @robins @mrigger

robins:
@guptapratykshh - A few things.

1. We don't want to go the route of `server_version_num` here. The goal is that ideally SQLancer doesn't need to check the version and should support all commands in the existing / supported version - See this discussion - (as discussed [here](https://github.com/sqlancer/sqlancer/issues/912#issuecomment-2566737713)) .

2. The PR #1053 seems to be on similar lines, and is waiting for v13 support to go live.

3. I think the intention of this PR, to restrict `PARALLEL` count is understandable, but sometimes restricting search space goes against what the tool is trying to do - in effect that number is intentionally left unrestricted so that if it (negative or very large int) triggers the engine to go wild, then so be it.

PR #1210, opened 2025-04-06
Fix UNARY_PLUS operator type promotion to DOUBLE PRECISION

### LINKED ISSUE #1209: Fix : UNARY_PLUS operator should convert value to DOUBLE in expected value generation
Currently, the getExpectedValue method for the UNARY_PLUS operator doesn't properly handle type promotion. In SQL (including Materialize/PostgreSQL), applying the unary + operator on a numeric value promotes it to DOUBLE PRECISION. However, the current implementation simply returns the input value as-is, without performing any conversion. This can lead to mismatches in expected vs. actual results during testing and cause invalid test oracle behavior.

Image

Proposed change
Implement the TODO comment by:

1) Adding support for type promotion to double precision in the UNARY_PLUS operator's expected value logic
2) Promoting the operand to a double using expectedValue.asDouble()
3) Creating and returning a MaterializeConstant using the promoted double value
4) Ensuring the UNARY_PLUS operator’s declared return type reflects this conversion

guptapratykshh:
# Description
This PR addresses an issue where the UNARY_PLUS operator (+) wasn't properly handling type promotion. In SQL (including Materialize/PostgreSQL), applying the unary + operator on a numeric value should promote it to DOUBLE PRECISION. The previous implementation simply returned the input value without performing any conversion, which could lead to mismatches in expected vs. actual results during testing.

# Changes
Changed UNARY_PLUS operator's return type from INT to FLOAT (double precision). Implemented proper conversion of values to double precision. Added handling for different input types (int, boolean, string, float). Added missing asDouble() method to FloatConstant class. Added comprehensive test cases to verify the proper behavior

# Testing
Added a tests file `TestMaterializeUnaryPlus.java` to verify the changes made.

Closes #1209

### DIFF
--- a/src/sqlancer/materialize/ast/MaterializeConstant.java
+++ b/src/sqlancer/materialize/ast/MaterializeConstant.java
@@ -316,6 +316,12 @@ public abstract class MaterializeConstant implements MaterializeExpression {
             return getTextRepresentation();
         }
 
+        @Override
+        public double asDouble() {
+            return (double) val;
+        }
+
+
     }
 
     public static MaterializeConstant createNullConstant() {
@@ -363,7 +369,15 @@ public abstract class MaterializeConstant implements MaterializeExpression {
         throw new UnsupportedOperationException(this.toString());
     }
 
+    public double asDouble() {
+        throw new UnsupportedOperationException(this.toString());
+    }
+
     public boolean isBoolean() {
+        return false;
+    }
+
+    public boolean isFloat() {
         return false;
     }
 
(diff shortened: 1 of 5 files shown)


robins:
@guptapratykshh - The idea possibly has merit (I'll let other reviewers comment on that aspect but IMHO) This is an avoidably complicated PR.

To assist / save time for the reviewers, please ensure that 1 PR has changes related to only 1 issue. (Here it seems the changes for v13 / VACUUM were added then reverted subsequently).

PR #1224, opened 2025-04-30
Support CODDTest for DuckDB

DerZc:
Support CODDTest for DuckDB. This is a sub-task of the larger implementation tracked in https://github.com/sqlancer/sqlancer/pull/1054.


src/sqlancer/duckdb/DuckDBErrors.java
```diff
@@ -57,6 +57,14 @@ public static List<String> getExpressionErrors() {
         errors.add("Cannot subtract infinite timestamps");
         errors.add("Timestamp difference is out of bounds");
 
+        // added by CODDTest
+        // errors.add("must appear in the GROUP BY clause or be used in an aggregate function");
```
mrigger, line 61: It seems we can remove these commented-out messages?

src/sqlancer/duckdb/DuckDBOptions.java
```diff
@@ -77,6 +77,25 @@ public class DuckDBOptions implements DBMSSpecificOptions<DuckDBOracleFactory> {
     @Parameter(names = "--max-num-updates", description = "The maximum number of UPDATE statements that are issued for a database", arity = 1)
     public int maxNumUpdates = 5;
 
+    public enum CODDTestModel {
```
mrigger, line 80: If this is the same for all CODDTest implementations, perhaps we could extract this to a common enum with the next PR?

src/sqlancer/duckdb/DuckDBSchema.java
```diff
@@ -41,9 +41,19 @@ public static class DuckDBCompositeDataType {
 
         private final int size;
 
+        // This is used to handle the type that out of the scope of our code
```
mrigger, line 44: Hmm, I don't understand the comment and use of this. Could we perhaps explain or give an example?

src/sqlancer/duckdb/DuckDBToStringVisitor.java
```diff
@@ -82,6 +104,72 @@ private void visit(DuckDBSelect select) {
...
+            }
+            sb.append(" THEN ");
+            visit(result.get(i));
+            sb.append(" ");
+        }
+        sb.append("END ");
+    }
+
+    public void visit(DuckDBTypeCast expr) {
+        if (expr.getExpression() instanceof DuckDBNullConstant) {
```
mrigger, line 149: It seems that we are re-interpreting the AST here (e.g., if it's a `NULL`, we ignore the cast). In the other parts, I think we treat the AST as a ground truth (i.e., we just directly print its structure). Would it be possible to construct the AST earlier on correctly, by omitting casts if we have `null` values or if the type is `unknown`?

src/sqlancer/duckdb/test/DuckDBCODDTestOracle.java
```diff
@@ -0,0 +1,1065 @@
...
+            }
+            idxTypeMap.put(i, cType);
+        }
+
+        // revert innerquery
+        if (innerQuery != null) {
+            innerQuery.setFetchColumns(Arrays.asList(innerQueryFetchColumn));
+        }
+
+        // for (Map.Entry<Integer, DuckDBCompositeDataType> entry: idxTypeMap.entrySet()) {
```
mrigger, line 947: Can we remove such commented-out code?

src/sqlancer/duckdb/test/DuckDBCODDTestOracle.java
```diff
@@ -0,0 +1,1065 @@
...
+    public boolean useSubquery() {
+        if (this.state.getDbmsSpecificOptions().coddTestModel.isRandom()) {
+            return Randomly.getBoolean();
+        } else if (this.state.getDbmsSpecificOptions().coddTestModel.isExpression()) {
+            return false;
+        } else if (this.state.getDbmsSpecificOptions().coddTestModel.isSubquery()) {
+            return true;
+        } else {
+            System.out.printf("Wrong option of --coddtest-model, should be one of: RANDOM, EXPRESSION, SUBQUERY");
+            System.exit(1);
```
mrigger, line 1047: I think using a switch-case on the enum would be clearer here, throwing an `AssertionError` in the default case.

src/sqlancer/duckdb/test/DuckDBCODDTestOracle.java
```diff
@@ -0,0 +1,1065 @@
...
+            List<String> v2Value = new ArrayList<>(
+                    v2.stream().map(c -> ((DuckDBConstant) c).toString()).collect(Collectors.toList()));
+            Collections.sort(v1Value);
+            Collections.sort(v2Value);
+            // if (!v1Value.equals(v2Value)) {
+            // return false;
+            // }
+            for (int i = 0; i < v1Value.size(); ++i) {
+                if (!v1Value.get(i).equals(v2Value.get(i))) {
+                    String regx = "[+-]*\\d+\\.?\\d*[Ee]*[+-]*\\d+";
```
mrigger, line 1023: What does this regex do? We should document this.

src/sqlancer/duckdb/test/DuckDBCODDTestOracle.java
```diff
@@ -0,0 +1,1065 @@
...
+                    Pattern pattern = Pattern.compile(regx);
+                    if (pattern.matcher(v1Value.get(i)).matches() && pattern.matcher(v2Value.get(i)).matches()) {
+                        if (!v1Value.get(i).substring(0, 6).equals(v2Value.get(i).substring(0, 6))) {
```
mrigger, line 1026: Where does the 6 come from? It would be good to document this.

src/sqlancer/duckdb/test/DuckDBCODDTestOracle.java
```diff
@@ -0,0 +1,1065 @@
...
+                        .add(DuckDBJoin.createRightOuterJoin(leftTable, rightTable, joinGen.generateExpression()));
+                break;
+            default:
+                throw new AssertionError();
+            }
+        }
+        return joinExpressions;
+    }
+
+    private List<DuckDBExpression> genOrderBys(DuckDBExpression specificCondition) {
```
mrigger, line 667: Do we have similar logic in the expression generator of DuckDB? If so, we can reuse it?

src/sqlancer/duckdb/test/DuckDBCODDTestOracle.java
```diff
@@ -0,0 +1,1065 @@
...
+        super(globalState);
+        this.s = globalState.getSchema();
+        DuckDBErrors.addExpressionErrors(errors);
+        DuckDBErrors.addInsertErrors(errors);
+        DuckDBErrors.addGroupByErrors(errors);
+    }
+
+    @Override
+    public void check() throws Exception {
+        reproducer = null;
```
mrigger, line 91: I feel like the logic in this class and method is difficult to understand. It would be ideal to refactor it, for example, by extracting methods and reducing code duplication between CODDTest implementations. However, I also understand that this might be risky, as we would no longer be certain that the implementation still works after refactoring.

mrigger (commented):
Sorry for the late review. I looked through the files and added some comments and suggestions. Let me know once you've had a look at them. In particular, the `DuckdbCODDTestOracle` is difficult to understand, and I think it will be tough to maintain and extend it in the future. However, I also understand that refactoring this now might be challenging. I'm fine with merging the current state, perhaps after addressing some of the easy issues, but it would also be good to think about what we could improve.

PR #1227, opened 2025-05-24
Add support for Alter Statistics and Set Statistics

yaojiejia:
Add support for altering and setting statistics.
Force_parallel_mode is deprecated and has been renamed to debug_parallel_query, therefore, I commented it out. 
Read more: https://www.postgresql.org/docs/16/release-16.html
I think postgres has a limitation where "exclude" constraints cannot be used together with partition keys that include expressions, therefore I added some errors handling in the ```PostgresCommon``` 

Apologize for putting all these in a single issue. Will make sure other issues are separated in the future.

Thanks!

### DIFF
--- a/src/sqlancer/postgres/PostgresProvider.java
+++ b/src/sqlancer/postgres/PostgresProvider.java
@@ -99,6 +99,7 @@ public class PostgresProvider extends SQLProviderAdapter<PostgresGlobalState, Po
         }), //
         CREATE_STATISTICS(PostgresStatisticsGenerator::insert), //
         DROP_STATISTICS(PostgresStatisticsGenerator::remove), //
+        ALTER_STATISTICS(PostgresStatisticsGenerator::alter), //
         DELETE(PostgresDeleteGenerator::create), //
         DISCARD(PostgresDiscardGenerator::create), //
         DROP_INDEX(PostgresDropIndexGenerator::create), //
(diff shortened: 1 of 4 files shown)


robins:
Thanks for the PR.

- Errors for EXCLUDE constraints seem okay to add. That is a technical limitation in postgres for this version, which can be removed / reviewed when a future version's release notes mentions otherwise.

- > "putting all these in a single issue".
  - +1. Regarding both points above, please avoid unrelated changes in the same PR - for simplicity / comprehension.

Lastly, as of now, SQLancer is adding support for features in v13 - #1072 . Also, there is already a PR for `force_parallel_mode` which is [waiting for v16 support](https://github.com/sqlancer/sqlancer/pull/1046#issuecomment-2564577725).

mrigger (approved):
LGTM! @robins, do you think this is ready to be merged as well?

robins:
> LGTM! @robins, do you think this is ready to be merged as well?

The changes are good and the 2 errors seem required too.
The `force_parallel_mode` is an unnecessary addition here (like the other PR), but otherwise LGTM too.

yaojiejia:
> > LGTM! @robins, do you think this is ready to be merged as well?
> 
> The changes are good and the 2 errors seem required too. The `force_parallel_mode` is an unnecessary addition here (like the other PR), but otherwise LGTM too.
> 
> @yaojiejia - Any chances you're using a non-v13 postgres build to test these?

Yep I think that is the case, I have been using v17 to test these and every time I get the force parallel mode issues. Just revert my local back to 13, so it should be good now.

mrigger:
> Yep I think that is the case, I have been using v17 to test these and every time I get the force parallel mode issues. Just revert my local back to 13, so it should be good now.

I think this is still part of the PR? Anyway, I'm merging this for now to not block progress.

PR #1228, opened 2025-05-24
Add support for Alter View Rename Column in Postgres v13

yaojiejia:
Closes 6. ALTER VIEW RENAME COLUMN in #912 
For some reason, FORCE_PARALLEL_MODE will not work locally without commenting out and I suspect it is due to a deprecated issue.

### DIFF
--- a/src/sqlancer/postgres/gen/PostgresAlterTableGenerator.java
+++ b/src/sqlancer/postgres/gen/PostgresAlterTableGenerator.java
@@ -48,7 +48,7 @@ public class PostgresAlterTableGenerator {
         SET_LOGGED_UNLOGGED, //
         NOT_OF, //
         OWNER_TO, //
-        REPLICA_IDENTITY
+        REPLICA_IDENTITY, ALTER_VIEW_RENAME_COLUMN // RENAME COLUMN old_name TO new_name (for views)
     }
 
     public PostgresAlterTableGenerator(PostgresTable randomTable, PostgresGlobalState globalState,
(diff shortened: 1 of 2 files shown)


src/sqlancer/postgres/gen/PostgresAlterTableGenerator.java
```diff
@@ -98,6 +98,20 @@ public List<Action> getActions(ExpectedErrors errors) {
             // make it more likely that the ALTER TABLE succeeds
             action = Randomly.subset(Randomly.smallNumber(), Action.values());
         }
+
+        // If this is a view, only allow view-compatible operations
+        if (randomTable.isView()) {
+            // Remove all table-specific operations for views
+            action.removeIf(a -> a != Action.ALTER_VIEW_RENAME_COLUMN);
```
robins, line 105: Do we want to create a list of actions (that are only possible on a view) and use that list in this check? That way, if / when we add more view-only actions, we wouldn't miss them here (and line 112 below).

src/sqlancer/postgres/gen/PostgresSetGenerator.java
```diff
@@ -112,7 +112,7 @@ private enum ConfigurationOption {
         JIT("jit", (r) -> Randomly.fromOptions(1, 0)),
         JOIN_COLLAPSE_LIMIT("join_collapse_limit", (r) -> r.getInteger(1, Integer.MAX_VALUE)),
         PARALLEL_LEADER_PARTICIPATION("parallel_leader_participation", (r) -> Randomly.fromOptions(1, 0)),
-        FORCE_PARALLEL_MODE("force_parallel_mode", (r) -> Randomly.fromOptions("off", "on", "regress")),
+        // FORCE_PARALLEL_MODE("force_parallel_mode", (r) -> Randomly.fromOptions("off", "on", "regress")),
```
mrigger, line 115: I guess this is an accidental commit? I see this change also in https://github.com/sqlancer/sqlancer/pull/1227.

mrigger (approved):
LGTM!

PR #1229, opened 2025-05-24
Add support for Drop Database Force in Postgres v13 

yaojiejia:
Closes 11. DROP DATABASE FORCE from #912

### DIFF
--- a/src/sqlancer/postgres/PostgresProvider.java
+++ b/src/sqlancer/postgres/PostgresProvider.java
@@ -277,7 +277,8 @@ public class PostgresProvider extends SQLProviderAdapter<PostgresGlobalState, Po
         globalState.getState().logStatement(String.format("\\c %s;", entryDatabaseName));
         
         String dropCommand = "DROP DATABASE";
-        if (Randomly.getBoolean()) {
+        boolean forceDrop = Randomly.getBoolean();
+        if (forceDrop) {
             dropCommand += " FORCE";
         }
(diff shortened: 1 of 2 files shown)


src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -275,12 +275,32 @@ public SQLConnection createDatabase(PostgresGlobalState globalState) throws SQLE
...
+        }
+        dropCommand += " IF EXISTS " + databaseName;
+        
+        globalState.getState().logStatement(dropCommand + ";");
         try (Statement s = con.createStatement()) {
-            s.execute("DROP DATABASE IF EXISTS " + databaseName);
+            s.execute(dropCommand);
+        } catch (SQLException e) {
+            // If force fails, fall back to regular drop
+            if (dropCommand.contains("FORCE")) {
```
mrigger, line 290: Could we have a boolean flag instead (e.g., `forceDrop`)? In this way, in case we later on remove setting `FORCE` above, we will also know that we need to remove this part here.

mrigger:
Looks good, I've just added another small nitpick. I think that we can merge the PR after that.

src/sqlancer/postgres/PostgresProvider.java
yaojiejia: Sorry for the late reply but of course! I will add a boolean instead.

PR #1230, opened 2025-05-24
Add support for Explain Analyze for Postgres v13

yaojiejia:
Closes 12. EXPLAIN ANALYSE from #912

### DIFF
--- a/src/sqlancer/postgres/gen/PostgresSetGenerator.java
+++ b/src/sqlancer/postgres/gen/PostgresSetGenerator.java
@@ -112,7 +112,7 @@ public final class PostgresSetGenerator {
         JIT("jit", (r) -> Randomly.fromOptions(1, 0)),
         JOIN_COLLAPSE_LIMIT("join_collapse_limit", (r) -> r.getInteger(1, Integer.MAX_VALUE)),
         PARALLEL_LEADER_PARTICIPATION("parallel_leader_participation", (r) -> Randomly.fromOptions(1, 0)),
-        // FORCE_PARALLEL_MODE("force_parallel_mode", (r) -> Randomly.fromOptions("off", "on", "regress")),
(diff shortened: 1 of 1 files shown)


yaojiejia:
> I think the code is currently used only by the Query Plan Guidance (QPG) component. If we execute `ANALYZE` for it, performance would drop. I guess we could have separate public methods, one for QPG and one that we use in the provider class.

I see, looking over the ```PostgresProvider``` class, I can only see that the ```PostgresExplainGenerator``` is being used in ```getQueryPlan``` method which it will definitely drop the performance by adding the analyze part, but you also mentioned I should create another method for the provider class, could you maybe elaborate on that part?

mrigger:
We could add an enum (see https://github.com/sqlancer/sqlancer/blob/49c735829617aef9f4dc351b5844a32cbeb6b441/src/sqlancer/postgres/PostgresProvider.java#L129). In this enum, we can specify a newly-created method in the explain generator that can generate the full range of `EXPLAIN`, while we can keep using the restricted method for QPG.

mrigger:
@yaojiejia, I see you have been working on multiple other PRs. Do you think it would be possible to focus on the open ones, so we can merge those first?

yaojiejia:
> @yaojiejia, I see you have been working on multiple other PRs. Do you think it would be possible to focus on the open ones, so we can merge those first?

Of course, will prioritize these for now, thanks

yaojiejia:
> We could add an enum (see
> 
> https://github.com/sqlancer/sqlancer/blob/49c735829617aef9f4dc351b5844a32cbeb6b441/src/sqlancer/postgres/PostgresProvider.java#L129
> ). In this enum, we can specify a newly-created method in the explain generator that can generate the full range of `EXPLAIN`, while we can keep using the restricted method for QPG.

Just included a method for EXPLAIN. Lmk if this is something we are looking for

PR #1231, opened 2025-05-24
Add support for Row Generator in Postgres v13

yaojiejia:
Closes 14. ROW from #912


mrigger (commented):
Are there any places where we could already generate `ROW` values (e.g., for certain operators)? With an actual use case or test case, I would feel more comfortable with merging this PR.

PR #1234, opened 2025-06-01
Add support for Create DB with Locale in Postgres v13

yaojiejia:
Closes 7. CREATE DATABASE WITH LOCALE from #912


mrigger (approved):
LGTM, thanks!

PR #1235, opened 2025-06-01
Remove cannot insert into column in Postgres v13

yaojiejia:
Closes 13. INSERT - Remove "Cannot insert intocolumn" from #912

### DIFF
--- a/src/sqlancer/postgres/gen/PostgresInsertGenerator.java
+++ b/src/sqlancer/postgres/gen/PostgresInsertGenerator.java
@@ -25,7 +25,7 @@ public final class PostgresInsertGenerator {
 
     public static SQLQueryAdapter insertRows(PostgresGlobalState globalState, PostgresSchema.PostgresTable table) {
         ExpectedErrors errors = new ExpectedErrors();
-        errors.add("cannot insert into column");
+        // errors.add("cannot insert into column");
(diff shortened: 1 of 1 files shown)


mrigger:
I think we can remove this rather than commenting it out? Can you perhaps check `git blame` or so to see when we introduced this (e.g., it might be that a PostgreSQL-compatible database system like Citus was the reason for this change)?

yaojiejia:
> I think we can remove this rather than commenting it out? Can you perhaps check `git blame` or so to see when we introduced this (e.g., it might be that a PostgreSQL-compatible database system like Citus was the reason for this change)?

Of course, after running ```git log -p --all -S "cannot insert into column"``` I can see you made a commit ```e58e4188d5ca22558738271910964a53b4ff2671``` that introduced ```cannot insert into column```. After this commit, there were two other commits ```46b7eecfdfabe71ab10431d981e0fcdc4654929d``` and ```46b7eecfdfabe71ab10431d981e0fcdc4654929d``` that were closely related to yugabytedb and materialize, maybe was from here?

mrigger:
It seems that the PostgreSQL tests fail if we disable this error. Thus, I am closing this for now.

By the way, it would be great to see whether we could fix the CI for the implementations that build on PostgreSQL like Citus and YugabyteDB. Like this, we see whether any additions or refactoring breaks them.

Issue #1226, opened 2025-05-12
The default configuration disable the generation of generated columns in SQLite


DerZc:
I find there is a potential incorrect code in the SQLite column builder.

In this line
https://github.com/sqlancer/sqlancer/blob/3d9a5cc63693731ebb6eee133cbd376137c14ec9/src/sqlancer/sqlite3/gen/SQLite3ColumnBuilder.java#L54
the incorrect constraint makes the code can not generate the generated columns with the default configuration.

mrigger:
Thanks, yes, that seems incorrect. In case you could create a PR to fix this, it would be great.

PR #1238, opened 2025-06-05
fix issue #1226

DerZc:
fix issue #1226 that incorrect constraint make the code can not generate the generated columns with the default configuration for SQLite


mrigger (approved):
LGTM, thanks!

PR #1250, opened 2025-06-29
Fix sqlancer#1249: Postgres REINDEX now generate only one index name per statement

albertZhangTJ:
In Postgres, each REINDEX statement can only work on one index, as can be seen [here](https://www.postgresql.org/docs/devel/sql-reindex.html). 

The previous edition will instead generate all index names, concatenated without delimiter.

### DIFF
--- a/src/sqlancer/postgres/gen/PostgresReindexGenerator.java
+++ b/src/sqlancer/postgres/gen/PostgresReindexGenerator.java
@@ -1,7 +1,6 @@
 package sqlancer.postgres.gen;
 
 import java.util.List;
-import java.util.stream.Collectors;
 
 import sqlancer.IgnoreMeException;
 import sqlancer.Randomly;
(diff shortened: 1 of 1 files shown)


src/sqlancer/postgres/gen/PostgresReindexGenerator.java
```diff
@@ -39,7 +38,7 @@ public static SQLQueryAdapter create(PostgresGlobalState globalState) {
             if (indexes.isEmpty()) {
                 throw new IgnoreMeException();
             }
-            sb.append(indexes.stream().map(i -> i.getIndexName()).collect(Collectors.joining()));
+            sb.append(indexes.stream().skip(Randomly.getNotCachedInteger(0, indexes.size())).findFirst().map(i -> i.getIndexName()).orElse(""));
```
mrigger, line 41: Thanks! That looks quite complex. Could we use the `Randomly` class to get a random element of the `indexes` list and obtain the name of the index?

Also, I think we should investigate why this code currently does not trigger any bugs. There might be a more fundamental issue. For example, do we only generate one index at most?

PR #1251, opened 2025-06-29
Feat: Integrating DQE Testing Approaches into SQLancer first in MySQL

luliqwerty:
This PR is to Integrating DQE (Differential Query Execution) Testing Approaches into SQLancer. [More information see DQE paper](https://ieeexplore.ieee.org/document/10172736).

Take MySQL as the first step. 
The code is divided into three main parts:

1. `DQEBase.class`: defines the core logic of DQE, including the addition and deletion of auxiliary columns, and the comparison of query results.

2. `SQLQueryError.java`: provide standardized error representation and comparison functionality for SQL query operations in DQE.

3. `MySQLDQEOracle.class`: Inherits from `DQEBase` and implements the specific logic of executing DQE in MySQL, including generating queries, executing queries, comparing results, etc.

### DIFF
--- a/src/sqlancer/common/oracle/DQEBase.java
+++ b/src/sqlancer/common/oracle/DQEBase.java
@@ -1,7 +1,6 @@
 package sqlancer.common.oracle;
 
 import java.sql.SQLException;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -117,16 +116,10 @@ public abstract class DQEBase<S extends SQLGlobalState<?, ?>> {
             return queryErrors.isEmpty();
         }
 
-        public boolean hasErrors() {
-            return !hasEmptyErrors();
-        }
-
         public boolean hasSameErrors(SQLQueryResult that) {
             if (queryErrors.size() != that.getQueryErrors().size()) {
                 return false;
             } else {
-                Collections.sort(queryErrors);
-                Collections.sort(that.getQueryErrors());
                 for (int i = 0; i < queryErrors.size(); i++) {
                     if (!queryErrors.get(i).equals(that.getQueryErrors().get(i))) {
                         return false;
--- a/src/sqlancer/mysql/oracle/MySQLDQEOracle.java
+++ b/src/sqlancer/mysql/oracle/MySQLDQEOracle.java
@@ -4,6 +4,7 @@ import static sqlancer.ComparatorHelper.getResultSetFirstColumnAsString;
 
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -49,9 +50,7 @@ public class MySQLDQEOracle extends DQEBase<MySQLGlobalState> implements TestOra
         MySQLErrors.addExpressionErrors(selectExpectedErrors);
 
         MySQLErrors.addExpressionErrors(updateExpectedErrors);
-        updateExpectedErrors.add("cannot be null");
-        updateExpectedErrors.add("Duplicate entry");
-        updateExpectedErrors.add("The value specified for generated column");
+        MySQLErrors.addInsertUpdateErrors(updateExpectedErrors);
 
         MySQLErrors.addExpressionErrors(deleteExpectedErrors);
         deleteExpectedErrors.add("a foreign key constraint fails");
@@ -173,8 +172,6 @@ public class MySQLDQEOracle extends DQEBase<MySQLGlobalState> implements TestOra
 
         if (!errorMessage.isEmpty()) {
             throw new AssertionError(errorMessage);
-        } else {
-            state.getState().getLocalState().log("PASS");
         }
 
         for (MySQLTable table : tables.getTables()) {
@@ -184,7 +181,7 @@ public class MySQLDQEOracle extends DQEBase<MySQLGlobalState> implements TestOra
 
     public String compareSelectAndUpdate(SQLQueryResult selectResult, SQLQueryResult updateResult) {
         if (updateResult.hasEmptyErrors()) {
-            if (selectResult.hasErrors()) {
+            if (!selectResult.hasEmptyErrors()) {
                 return "SELECT has errors, but UPDATE does not.";
             }
             if (!selectResult.hasSameAccessedRows(updateResult)) {
@@ -246,7 +243,7 @@ public class MySQLDQEOracle extends DQEBase<MySQLGlobalState> implements TestOra
 
     public String compareSelectAndDelete(SQLQueryResult selectResult, SQLQueryResult deleteResult) {
         if (deleteResult.hasEmptyErrors()) {
-            if (selectResult.hasErrors()) {
+            if (!selectResult.hasEmptyErrors()) {
                 return "SELECT has errors, but DELETE does not.";
             }
             if (!selectResult.hasSameAccessedRows(deleteResult)) {
@@ -459,7 +456,7 @@ public class MySQLDQEOracle extends DQEBase<MySQLGlobalState> implements TestOra
             }
             resultSet.close();
         }
-
+        Collections.sort(queryErrors);
         return queryErrors;
     }
 
--- a/src/sqlancer/mysql/MySQLErrors.java
+++ b/src/sqlancer/mysql/MySQLErrors.java
@@ -53,6 +53,7 @@ public final class MySQLErrors {
         errors.add("Data truncated for functional index");
         errors.add("cannot be null");
         errors.add("Incorrect decimal value");
+        errors.add("The value specified for generated column");
 
         return errors;
     }


JensonSung:
I think current implementation is good but specialized for DQE method.

JensonSung:
> Thanks, that's great! Some high-level comments first:
> 
> * for the `SQLQueryError`, this looks quite similar to `ExpectedErrors`. Do we really need this class or can we combine these classes?
> * for statements like `UPDATE`, `DELETE`, and `ALTER TABLE`, do you think we can use and extend the generator classes rather than implementing the logic in the DQE oracle?

First, revising ExepctedErrors like SQLQueryError to provide a structured information may help ease to identify errors compared to a string-formed information. While DQE does not distinguish which error is expected but compare whether errors are consistent. A clear name could be required.
Second, it seems that every kind of statements like `update` and `delete` deserve a proper generator, preparing for new testing methods.
These two improvements can be put on the agenda.

src/sqlancer/common/query/SQLQueryAdapter.java
```diff
@@ -103,8 +113,12 @@ public String getUnterminatedQueryString() {
             return true;
         } catch (Exception e) {
             Main.nrUnsuccessfulActions.addAndGet(1);
-            checkException(e);
-            return false;
+            // checkException(e);
+            if (reportException) {
```
mrigger, line 117: How does commenting this out change things? Is this compatible with other oracles and expected errors? It would be great if we could extract changes to the core parts into separate PRs so we can review and merge them more quickly.

src/sqlancer/common/query/SQLQueryError.java
```diff
@@ -0,0 +1,116 @@
+package sqlancer.common.query;
+
+public class SQLQueryError implements Comparable<SQLQueryError> {
```
mrigger, line 3: We could try to merge this quickly by creating a separate PR for this and adding also some tests for these methods.

src/sqlancer/mysql/oracle/MySQLDQEOracle.java
```diff
@@ -0,0 +1,502 @@
...
+        for (MySQLTable table : tables.getTables()) {
+            selectColumns.add(table.getName() + "." + COLUMN_ROWID);
+        }
+        boolean operateOnSingleTable = tables.getTables().size() == 1;
+        boolean generateOrderBy = false;
+        List<String> orderColumns = new ArrayList<>();
+        boolean generateLimit = false;
+        int limit = 0;
+        if (operateOnSingleTable) {
+            if (Randomly.getBooleanWithSmallProbability()) {
```
mrigger, line 76: I think the expression generator already has methods for generating `ORDER BY`s, so we should try to re-use those. Could we perhaps also re-use or factor out components from the random query generator?

src/sqlancer/mysql/oracle/MySQLDQEOracle.java
```diff
@@ -0,0 +1,502 @@
...
+        } finally {
+            queryErrors = getErrors();
+
+            for (MySQLTable table : tables.getTables()) {
+                String tableName = table.getName();
+                String rowId = tableName + "." + COLUMN_ROWID;
+                String selectRowId = String.format("SELECT %s FROM %s", rowId, tableName);
+                SQLancerResultSet resultSet = new SQLQueryAdapter(selectRowId).executeAndGet(state);
+                HashSet<String> rows = new HashSet<>();
+                if (resultSet != null) {
```
mrigger, line 452: I think we have some auxiliary methods for fetching rows (used by other test oracles). Can we perhaps re-use any of these here?

src/sqlancer/common/oracle/DQEBase.java
```diff
@@ -0,0 +1,107 @@
...
+import sqlancer.SQLConnection;
+import sqlancer.SQLGlobalState;
+import sqlancer.common.query.ExpectedErrors;
+import sqlancer.common.query.SQLQueryAdapter;
+import sqlancer.common.query.SQLQueryError;
+import sqlancer.common.schema.AbstractRelationalTable;
+
+public abstract class DQEBase<S extends SQLGlobalState<?, ?>> {
+
+    public static String COLUMN_ROWID = "rowId";
```
mrigger, line 20: for those not familiar with DQE, it would be great to explain important bits in the common base classes

src/sqlancer/mysql/oracle/MySQLDQEOracle.java
```diff
@@ -0,0 +1,502 @@
...
+        MySQLErrors.addExpressionErrors(updateExpectedErrors);
+        updateExpectedErrors.add("cannot be null");
+        updateExpectedErrors.add("Duplicate entry");
+        updateExpectedErrors.add("The value specified for generated column");
+
+        MySQLErrors.addExpressionErrors(deleteExpectedErrors);
+        deleteExpectedErrors.add("a foreign key constraint fails");
+    }
+
+    public void check() throws SQLException {
```
mrigger, line 51: This method is quite long. Is it possible to break it apart into multiple smaller methods, or would that require passing around many parameters? In addition, we could try to make it easier for other dialects to implement DQE by specifying some methods in the superclass that should be implemented.

mrigger:
> Given these differences, merging them into a single class could lead to confusion, increased coupling, and reduced clarity in terms of responsibility separation. I believe keeping them separate helps maintain a clean architecture and supports better extensibility in the future.

Okay, then let's leave them separate for now. I guess there is a conceptual commonality (both are about errors returned by the database system), but if that does not (yet) make sense at an implementation level, we could try to address this later.

src/sqlancer/common/query/SQLQueryAdapter.java
JensonSung: DQE does not check exception separately, while other testing methods may need. I wonder if we can use `reportException` to control this behavior. If a testing method requires checking exception, we first check whether such exception is expected. If it is not expected, the exception should be handled later by the testing method.

src/sqlancer/mysql/oracle/MySQLDQEOracle.java
JensonSung: @luliqwerty Try to find if we can re-use some auxiliary methods for fetching rows in other test oracles.

luliqwerty: Hello @mrigger @JensonSung 
I can't find a suitable method to get rows you mentioned. The confusing method is CERTOracle.getRow(), which is used to get the count of rows. Can you give me some advice or maybe we can just use the way I used to do.

mrigger: I think there could be one in https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/ComparatorHelper.java. If not, we could stick to the current implementation.

src/sqlancer/mysql/oracle/MySQLDQEOracle.java
```diff
@@ -0,0 +1,518 @@
...
+    private SQLQueryResult executeDelete(String deleteStmt, MySQLTables tables) throws SQLException {
+        Map<AbstractRelationalTable<?, ?, ?>, Set<String>> accessedRows = new HashMap<>();
+        List<SQLQueryError> queryErrors;
+        try {
+            for (MySQLTable table : tables.getTables()) {
+                String tableName = table.getName();
+                String rowId = tableName + "." + COLUMN_ROWID;
+                String selectRowId = String.format("SELECT %s FROM %s", rowId, tableName);
+                SQLancerResultSet resultSet = new SQLQueryAdapter(selectRowId).executeAndGet(state, false);
+                HashSet<String> rows = new HashSet<>();
```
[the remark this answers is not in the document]
sayJason: Yes, some methods could be extracted for fetching these helper columns, like fetching rowID. @luliqwerty

src/sqlancer/mysql/oracle/MySQLDQEOracle.java
JensonSung: Current implementation can reuse the method `getResultSetFirstColumnAsString`.

src/sqlancer/mysql/oracle/MySQLDQEOracle.java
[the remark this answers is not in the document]
luliqwerty: @mrigger you are right, thanks a lot, I've reused the existing method.

src/sqlancer/common/oracle/DQEBase.java
```diff
@@ -0,0 +1,156 @@
...
+
+        public boolean hasErrors() {
+            return !hasEmptyErrors();
+        }
+
+        public boolean hasSameErrors(SQLQueryResult that) {
+            if (queryErrors.size() != that.getQueryErrors().size()) {
+                return false;
+            } else {
+                Collections.sort(queryErrors);
```
mrigger, line 128: I think it would be better to sort the errors at the time they are added to an object, not when comparing them.

src/sqlancer/common/oracle/DQEBase.java
```diff
@@ -0,0 +1,156 @@
...
+
+        public Map<AbstractRelationalTable<?, ?, ?>, Set<String>> getAccessedRows() {
+            return accessedRows;
+        }
+
+        public List<SQLQueryError> getQueryErrors() {
+            return queryErrors;
+        }
+
+        public boolean hasEmptyErrors() {
```
mrigger, line 116: I think we should avoid introducing redundant methods. This and the one below seem to have the same semantics.

src/sqlancer/common/oracle/DQEBase.java
```diff
@@ -0,0 +1,156 @@
...
+            return queryErrors.isEmpty();
+        }
+
+        public boolean hasErrors() {
+            return !hasEmptyErrors();
+        }
+
+        public boolean hasSameErrors(SQLQueryResult that) {
```
mrigger, line 124: Since this method invokes `equals` on all the elements, I think this method should probably override `equals` as well?

src/sqlancer/mysql/oracle/MySQLDQEOracle.java
```diff
@@ -0,0 +1,498 @@
...
+    private int limit;
+
+    public MySQLDQEOracle(MySQLGlobalState state) {
+        super(state);
+        schema = state.getSchema();
+
+        MySQLErrors.addExpressionErrors(selectExpectedErrors);
+
+        MySQLErrors.addExpressionErrors(updateExpectedErrors);
+        updateExpectedErrors.add("cannot be null");
```
mrigger, line 52: Instead of re-adding the errors here, can we just use existing methods like https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/mysql/MySQLErrors.java#L60?

src/sqlancer/mysql/oracle/MySQLDQEOracle.java
```diff
@@ -0,0 +1,498 @@
...
+        String compareUpdateAndDelete = compareUpdateAndDelete(updateExecutionResult, deleteExecutionResult);
+
+        String errorMessage = compareSelectAndUpdate == null ? "" : compareSelectAndUpdate + "\n";
+        errorMessage += compareSelectAndDelete == null ? "" : compareSelectAndDelete + "\n";
+        errorMessage += compareUpdateAndDelete == null ? "" : compareUpdateAndDelete + "\n";
+
+        if (!errorMessage.isEmpty()) {
+            throw new AssertionError(errorMessage);
+        } else {
+            state.getState().getLocalState().log("PASS");
```
mrigger, line 177: I think it would be better to remove this to not clutter the log files with non-essential information.

mrigger (commented):
Just did another quick and noticed some final small requests. Could you please have a last look at them before we merge the PR? Other than that, @JensonSung, do the changes look good to you?

Excited to finally have DQE supported in SQLancer! Thanks a lot, @luliqwerty!

mrigger:
Agreed, lgtm, thanks a lot!

mrigger:
I just noticed that when executing the tests, there seems to be an issue:
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running sqlancer.dbms.TestMySQLDQE
java.sql.SQLNonTransientConnectionException: No operations allowed after connection closed.
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:110)
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97)
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:89)
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:63)
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:73)
```

Could you check whether you can reproduce this locally? This subsequently also results in a crash.

luliqwerty:
> I just noticed that when executing the tests, there seems to be an issue:
> 
> ```
> [INFO] -------------------------------------------------------
> [INFO]  T E S T S
> [INFO] -------------------------------------------------------
> [INFO] Running sqlancer.dbms.TestMySQLDQE
> java.sql.SQLNonTransientConnectionException: No operations allowed after connection closed.
> 	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:110)
> 	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97)
> 	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:89)
> 	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:63)
> 	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:73)
> ```
> 
> Could you check whether you can reproduce this locally? This subsequently also results in a crash.

This is an unexpected error, as the CI for the merge commit passed. The error indicates that the code is attempting to execute an SQL operation on a database connection that has already been closed. However, it seems that DQE did not perform this action. @mrigger @JensonSung

mrigger:
Could it be that the local version of MySQL and the one in the CI are different? If they are the same, perhaps it could be helpful to run for DQE a bit longer. It could be that it triggered a crash bug, as pointed out by @JensonSung.

Issue #1241, opened 2025-06-06
Integrate C-Reduce for Automatic Test Case Reduction


jinhuix:
### Problem Statement

Currently, SQLancer users must manually capture database states and write interestingness scripts when using C-Reduce for test case minimization. This process is error-prone and limits automation capabilities.

### Implementation Plan

1. **Add CLI option (`MainOptions.java`):**

```java
@Parameter(names = "--use-creduce", description = "EXPERIMENTAL Attempt to reduce queries using c-reduce")
private boolean useCReduce = false; // NOPMD
```

2. **Modify Exception Handling (`Main.java`):**

Inside `executeMain(...)`: extend the existing `catch (Throwable reduce)` block.

```java
try {
    ...
} catch (Throwable reduce) {
    reduce.printStackTrace();
    executor.getStateToReproduce().exception = reduce.getMessage();
    executor.getLogger().logFileWriter = null;
    executor.getLogger().logException(reduce, executor.getStateToReproduce());
    
    if (options.useCreduce()) {
        Reducer<G> reducer = new CReducer<>(provider);
        creducer.reduce(/* state, reduce, ... */);
    }
    return false;
}
```

3. **CReducer.reduce(...) should:**
   - Generate `databasex.sql`: Dump all statements from `state.getStatements()`, and append the failing statement (from `reduce.detailMessage`)
   - Generate `databasex.sh` (interestingness script): Extract the error substring from `reduce.cause.detailMessage` (e.g., from `[SQLITE_CORRUPT]  The database disk image is malformed (database disk image is malformed)` grab `database disk image is malformed`), then can generate a script like: `sqlite3 databasex.sql 2>&1 | grep -zq "database disk image is malformed"`.
   - Invoke C-reduce by `ProcessBuilder`, like: `Process p = new ProcessBuilder("creduce", "databasex.sh", "databasex.sql").start();`.

### Questions

- Is this design sensible? Any obvious pitfalls?
- How to handle missing or mismatched SQLite versions? The interestingness script may fail.
- This approach only works for SQLite. How can we generalize it to support other DBMSs?

@mrigger @KabilanMA Could you review this design before implementation? Thanks!

mrigger:
Here is some feedback:
* As a first step, a very simple self-implemented reducer might be better, I think (e.g., implemented in a Java file that is part of the repository). After this, we can think about how we best want to handle dependency management.
[...]

PR #1252, opened 2025-06-30
Add a standalone reducer and serialization support

jinhuix:
**Resolves [#1241](https://github.com/sqlancer/sqlancer/issues/1241)**

This PR introduces a standalone SQL reducer mechanism to SQLancer. It enables automated simplification of bug-inducing SQL queries in an offline manner via serialized context. The PR is split into 5 commits for clarity:

1. **Serializable context**: Add `SerializableReducerContext` class to store reproducible context (SQL queries, expected errors, and oracle type, etc).
2. **SimpleReducer**: Implement a lightweight `SimpleReducer` engine that uses delta-debugging-like reduction to minimize failing SQL inputs. It supports both exception-based and oracle-based reduction (`NoRECOracle`, `TLPWhereOracle`).
3. **Integration Hooks**: Add minimal changes to the SQLancer core to support optional serialization of reducer context data (via CLI flags).
4. **Testing**: Add a `TestSimpleReducer.java` class for validation, covering `Exception`, `NoREC`, and `TLPWhere` bugs.
5. **Documentation**: Add usage guide with examples and more details in `README.md`.

@mrigger @KabilanMA Happy to hear any thoughts or suggestions!


mrigger:
Thanks, this looks like a great start, and it's great to see that this also works!

The information you save in the reducer context is quite low level (i.e., strings). Do you think we can save more high-level information? This would make the approach more robust, as the current parsing seems quite complex. We could also do this for the expected errors. Rather than checking for every statement whether an error matches any expected error of the system, we could store a detailed statement-to-expected-error mapping. When reading back the reduced statements, this would require doing some kind of best-effort alignment. For now, with the line granularity, this could be a 1:1 match per statement.

jinhuix:
> If we serialize the reproducer, I assume we will not need this information?

My original concern was that `Reproducer` is an interface and may have different implementations for each oracle, which could make direct serialization difficult. So instead, we can create oracle-specific `ReproducerData` classes (e.g., `NoRECReproducerData`) that store only the necessary fields for reproduction.

> I think this should probably be a Map<SQLQuery, ExpectedErrors> mapping.

Agreed — I'll change the structure accordingly.

reducer/TestSimpleReducer.java
```diff
@@ -0,0 +1,141 @@
...
+        System.out.println("\n--- " + testName + " ---");
+        test.run();
+    }
+
+    private static void testException() {
+        try {
+            ReducerContext context = createExceptionContext();
+            executeTest(context);
+        } catch (Exception e) {
+            System.err.println("Exception test failed: " + e.getMessage());
```
mrigger, line 28: It's unusual to print information about success and failure in unit tests. It would be better to let the JUnit framework handle this.

reducer/TestSimpleReducer.java
```diff
@@ -0,0 +1,141 @@
...
+        } catch (Exception e) {
+            System.err.println("TLP WHERE Oracle test failed: " + e.getMessage());
+        }
+    }
+
+    private static void executeTest(ReducerContext context) throws Exception {
+        SimpleReducer reducer = new SimpleReducer(context);
+        List<String> result = reducer.reduce();
+
+        System.out.println("Final reduced statements:");
```
mrigger, line 54: We should check the expected result of each test case.

src/sqlancer/ReducerContext.java
```diff
@@ -0,0 +1,108 @@
+package sqlancer;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Set;
+import java.util.Map;
+import java.util.HashMap;
+
+public class ReducerContext implements Serializable {
```
mrigger, line 9: I'm not very sure why we need this class. Wouldn't it be sufficient and simpler to serialize the `Reproducer` object or so?

jinhuix: > I'm not very sure why we need this class. Wouldn't it be sufficient and simpler to serialize the `Reproducer` object or so?

You're right! we can serialize the existing [`StateToReproduce`](https://github.com/sqlancer/sqlancer/blob/main/src/sqlancer/StateToReproduce.java), since `Reproducer` is an interface and doesn’t hold enough info for reduction.

mrigger:
I think either works for me, but I think having some smaller self-contained changes would help us in more quickly reviewing and merging things.

Issue #1244, opened 2025-06-10
PostgreSQL Version 14 Upgrade - Supporting Common Table Expression


yaojiejia:
## Problem
Currently, Common Table Expression (CTE) is introduced after version 12 and not yet supported in our Postgres system: https://www.postgresql.org/docs/13/queries-with.html
We need to implement the CTE as well as modify the select statement on selecting the CTE

## Implementation Plan
1. Create new CTE class in the AST directory
2. Add CTE support in ```PostgresSelect``` class
3. Create CTE generation in ```PostgresExpressionGenerator``` class
4. Modify the ```generateSelect()``` method in ```PostgresExpressionGenerator``` class

PR #1253, opened 2025-07-02
Add CTE Function to Postgres System

yaojiejia:
#1244
Still developing the recursive CTE function


mrigger:
Sounds good. Do you think we could also add a test that runs a certain number of randomly-generated CTEs to check whether there are no unexpected syntax errors or so? Otherwise, they have a low chance of being generated (as we don't generate many views, and not many CTEs in them according to your change).

PR #1255, opened 2025-07-05
Add Support for Merge Generator in Postgres v15

### LINKED ISSUE #1254: PostgreSQL Version 15 Upgrade - Supporting MERGE Function
MERGE function will give users the option to manually synchronize two tables, namely the target table and the source table. It supports:
- Inserting
- Updating
- Deleting 
Basic syntax:
```
[ WITH with_query [, ...] ]
MERGE INTO [ ONLY ] target_table_name [ * ] [ [ AS ] target_alias ]
USING data_source ON join_condition
when_clause [...]

where data_source is:

{ [ ONLY ] source_table_name [ * ] | ( source_query ) } [ [ AS ] source_alias ]

and when_clause is:

{ WHEN MATCHED [ AND condition ] THEN { merge_update | merge_delete | DO NOTHING } |
  WHEN NOT MATCHED [ AND condition ] THEN { merge_insert | DO NOTHING } }

and merge_insert is:

INSERT [( column_name [, ...] )]
[ OVERRIDING { SYSTEM | USER } VALUE ]
{ VALUES ( { expression | DEFAULT } [, ...] ) | DEFAULT VALUES }

and merge_update is:

UPDATE SET { column_name = { expression | DEFAULT } |
             ( column_name [, ...] ) = [ ROW ] ( { expression | DEFAULT } [, ...] ) |
             ( column_name [, ...] ) = ( sub-SELECT )
           } [, ...]

and merge_delete is:

DELETE

```

Read more here: https://www.postgresql.org/docs/15/sql-merge.html

yaojiejia:
Closes #1254

### DIFF
--- a/src/sqlancer/postgres/PostgresProvider.java
+++ b/src/sqlancer/postgres/PostgresProvider.java
@@ -40,6 +40,7 @@ import sqlancer.postgres.gen.PostgresDropIndexGenerator;
 import sqlancer.postgres.gen.PostgresExplainGenerator;
 import sqlancer.postgres.gen.PostgresIndexGenerator;
 import sqlancer.postgres.gen.PostgresInsertGenerator;
+import sqlancer.postgres.gen.PostgresMergeGenerator;
 import sqlancer.postgres.gen.PostgresNotifyGenerator;
 import sqlancer.postgres.gen.PostgresReindexGenerator;
 import sqlancer.postgres.gen.PostgresSequenceGenerator;
@@ -104,6 +105,7 @@ public class PostgresProvider extends SQLProviderAdapter<PostgresGlobalState, Po
         DISCARD(PostgresDiscardGenerator::create), //
         DROP_INDEX(PostgresDropIndexGenerator::create), //
         INSERT(PostgresInsertGenerator::insert), //
+        MERGE(PostgresMergeGenerator::merge), //
         UPDATE(PostgresUpdateGenerator::create), //
         TRUNCATE(PostgresTruncateGenerator::create), //
         VACUUM(PostgresVacuumGenerator::create), //
@@ -193,6 +195,9 @@ public class PostgresProvider extends SQLProviderAdapter<PostgresGlobalState, Po
         case UPDATE:
             nrPerformed = r.getInteger(0, 10);
             break;
+        case MERGE:
+            nrPerformed = r.getInteger(0, 5);
+            break;
         case INSERT:
             nrPerformed = r.getInteger(0, globalState.getOptions().getMaxNumberInserts());
             break;
@@ -279,14 +284,14 @@ public class PostgresProvider extends SQLProviderAdapter<PostgresGlobalState, Po
         }
         Connection con = DriverManager.getConnection("jdbc:" + entryURL, username, password);
         globalState.getState().logStatement(String.format("\\c %s;", entryDatabaseName));
-        
+
         String dropCommand = "DROP DATABASE";
         boolean forceDrop = Randomly.getBoolean();
         if (forceDrop) {
             dropCommand += " FORCE";
         }
         dropCommand += " IF EXISTS " + databaseName;
-        
+
         globalState.getState().logStatement(dropCommand + ";");
         try (Statement s = con.createStatement()) {
             s.execute(dropCommand);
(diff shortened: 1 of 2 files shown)


src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -279,14 +284,14 @@ public SQLConnection createDatabase(PostgresGlobalState globalState) throws SQLE
         }
         Connection con = DriverManager.getConnection("jdbc:" + entryURL, username, password);
         globalState.getState().logStatement(String.format("\\c %s;", entryDatabaseName));
-        
+
```
mrigger, line 287: It would be better to avoid whitespace changes.

src/sqlancer/postgres/gen/PostgresMergeGenerator.java
```diff
@@ -0,0 +1,426 @@
...
+        errors.add("but expression is of type");
+        errors.add("duplicate key value violates unique constraint");
+        errors.add("identity column defined as GENERATED ALWAYS");
+        errors.add("out of range");
+        errors.add("violates check constraint");
+        errors.add("no partition of relation");
+        errors.add("invalid input syntax");
+        errors.add("division by zero");
+        errors.add("data type unknown");
+        errors.add("MERGE is not supported in this version of PostgreSQL");
```
mrigger, line 49: I think it would be better to focus on a single version of PostgreSQL, so I think we should not add this as an expected error.

src/sqlancer/postgres/gen/PostgresMergeGenerator.java
```diff
@@ -0,0 +1,426 @@
...
+        errors.add("syntax error at or near \"MERGE\"");
```
mrigger, line 50: I think we should not have any syntax errors. What causes this?

mrigger: Seems there is also another syntax error specification below.

src/sqlancer/postgres/gen/PostgresMergeGenerator.java
```diff
@@ -0,0 +1,426 @@
...
+                PostgresExpressionGenerator gen = new PostgresExpressionGenerator(globalState).setColumns(columns);
+                PostgresExpression expression = gen.generateExpression(0, column.getType());
+                sb.append(PostgresVisitor.asString(expression));
+            }
+        }
+
+        return sb.toString();
+    }
+
+    private static String generateWhereCondition(PostgresTable targetTable) {
```
mrigger, line 322: I guess we can use the existing expression generator here? We can also add additional pseudo columns (like those prefixed by "target".

src/sqlancer/postgres/gen/PostgresMergeGenerator.java
```diff
@@ -0,0 +1,426 @@
...
+            if (i > 0) {
+                sb.append(", ");
+            }
+
+            PostgresColumn column = columns.get(i);
+
+            boolean sourceHasColumn = sourceTable != null
+                    && sourceTable.getColumns().stream().anyMatch(c -> c.getName().equals(column.getName()));
+
+            if (Randomly.getBoolean() && sourceHasColumn) {
```
mrigger, line 414: I see common patterns that also appear, for example, for the update case. Can we factor the common code out?

src/sqlancer/postgres/gen/PostgresMergeGenerator.java
yaojiejia: Oops I forgot to remove it as it was only for debugging purposes

src/sqlancer/postgres/gen/PostgresMergeGenerator.java
```diff
@@ -0,0 +1,378 @@
...
+                hasMatchedClause = true;
+                if (Randomly.getBoolean()) {
+                    sb.append(" UPDATE SET ");
+                    try {
+                        if (useSubquery) {
+                            sb.append(generateUpdateSet(targetTable, sourceColumns));
+                        } else {
+                            sb.append(generateUpdateSet(targetTable, sourceTable));
+                        }
+                    } catch (Exception e) {
```
mrigger, line 117: This looks a bit surprising. Why does throwing an exception affect the logic of what we generate here? Can we achieve the intended outcome without relying on exceptions?

PR #1260, opened 2025-07-26
fix wrong value in the tidb_enable_stmt_summary

hawkingrei:
https://docs.pingcap.com/tidb/stable/system-variables/#tidb_scatter_region

The default value of `tidb_scatter_region` is not a number. I modified the value so that it won't cause any errors.


mrigger:
Thanks! Merging this PR. Thanks also for fixing the other issues, which were likely caused due to changes in TiDB that are not yet reflected in SQLancer.

PR #1261, opened 2025-07-27
Serialize StateToReproduce to JSON

jinhuix:
### Motivation

To support a standalone reducer in SQLancer, we need a way to serialize the `StateToReproduce` object, which encapsulates all the information needed to reproduce a bug.

### Solution

- This PR introduces a mechanism to serialize `StateToReproduce` into a JSON file when a bug is found.
- The serialization captures all essential fields: `statements`, `errorType` (EXCEPTION, NOREC, TLP_WHERE), `databaseName`, `databaseProvider`, `reproducerData`, `exception`
- Serialization is controlled via a new command-line option: `--serialize-reproduce-state`, and serialized `.json` files are written to `logs/<dbms>/reproduce/`.
- A private static inner class `StateToReproduceSerializor` is introduced to control which fields are included in the output.

### Example Output

```json
{
  "statements": [
    "CREATE VIRTUAL TABLE rt0 USING rtree(c0, c1, c2, c3, c4);",
    ...
  ],
  "databaseName": "database6",
  "databaseProvider": "sqlancer.sqlite3.SQLite3Provider",
  "errorType": "NOREC",
  "expectedErrorsMap": {
    "INSERT OR FAIL INTO rt0 VALUES (NULL, 0.9105867290640974, \u0027\u0027, x\u00270a74\u0027, NULL), (x\u0027\u0027, NULL, 0x3cef0261, NULL, NULL), (\u0027280862872\u0027, \u00271022296673\u0027, \u0027n\u0027, x\u0027c9c3f71d\u0027, NULL);": [
      "[SQLITE_CONSTRAINT]  Abort due to constraint violation (rtree constraint failed: rt0.(c1\u003c\u003dc2))"
    ],
    ...
  },
  "reproducerData": {
    "optimizedQueryString": "SELECT COUNT(*) FROM rt0 WHERE (((SQLITE_SOURCE_ID())\u003c\u003d(rt0.c4)))",
    "unoptimizedQueryString": "SELECT SUM(count) FROM (SELECT ((((SQLITE_SOURCE_ID())\u003c\u003d(rt0.c4))) IS TRUE)  as count FROM rt0)",
    "shouldUseAggregate": "true"
  }
}
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -114,7 +114,6 @@
           <target>11</target>
           <compilerArguments>
             <properties>${project.basedir}/.settings/org.eclipse.jdt.core.prefs</properties>
-            <arg>-Xlint:-serial</arg>
           </compilerArguments>
           <compilerId>eclipse</compilerId>
           <showWarnings>true</showWarnings>
--- a/pom.xml
+++ b/pom.xml
@@ -114,6 +114,7 @@
           <target>11</target>
           <compilerArguments>
             <properties>${project.basedir}/.settings/org.eclipse.jdt.core.prefs</properties>
+            <arg>-Xlint:-serial</arg>
           </compilerArguments>
           <compilerId>eclipse</compilerId>
           <showWarnings>true</showWarnings>
--- a/src/sqlancer/StateToReproduce.java
+++ b/src/sqlancer/StateToReproduce.java
@@ -16,7 +16,7 @@ import sqlancer.common.query.Query;
 public class StateToReproduce implements Serializable {
     private static final long serialVersionUID = 1L;
 
-    private transient List<Query<?>> statements = new ArrayList<>();
+    private List<Query<?>> statements = new ArrayList<>();
 
     private final String databaseName;
 
@@ -157,19 +157,11 @@ public class StateToReproduce implements Serializable {
     private void writeObject(ObjectOutputStream out) throws IOException {
         out.defaultWriteObject();
 
-        List<String> statementStrings = new ArrayList<>();
-        for (Query<?> q : this.statements) {
-            statementStrings.add(q.getLogString());
-        }
-        out.writeObject(statementStrings);
         out.writeObject(this.databaseProvider != null ? this.databaseProvider.getDBMSName() : null);
     }
 
-    @SuppressWarnings("unchecked")
     private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
         in.defaultReadObject();
-        this.statements = new ArrayList<>();
-        List<String> statementStrings = (List<String>) in.readObject();
         String dbmsName = (String) in.readObject();
 
         DatabaseProvider<?, ?, ?> provider = null;
@@ -182,14 +174,7 @@ public class StateToReproduce implements Serializable {
                 }
             }
         }
-
-        if (provider == null || statementStrings == null) {
-            throw new AssertionError("Database provider or statement is null");
-        }
         this.databaseProvider = provider;
-        for (String s : statementStrings) {
-            this.statements.add(provider.getLoggableFactory().getQueryForStateToReproduce(s));
-        }
     }
 
     public void setStatements(List<Query<?>> statements) {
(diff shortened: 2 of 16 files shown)


jinhuix:
> Why do we convert it to JSON? Can we not just use the Java serialization API?

Yes. I used JSON for readability, but Java serialization is indeed simpler. I’m fine switching to that.

KabilanMA (approved):
LGTM

mrigger:
Sorry for the delay and thanks for pinging me, @KabilanMA.

I think this could be simplified a lot. At a high level, I see what we are adding an additional class `StateToReproduceSerializor` as well as various fields and methods that we use to store the data and information about the test oracles. Would it not be possible to just serialize the existing objects, without creating any extra classes or fields?

KabilanMA:
We can serialize the `StateToReproduce` class directly to store the necessary information, unless it's used in other contexts with different requirements. If that's the case, it's cleaner to keep the serialization in a separate class and include only the required fields.

Either approach is valid; it ultimately comes down to maintainability. Keeping a separate serializable class provides better clarity and separation of concerns, especially if `StateToReproduce` serves multiple purposes.

src/sqlancer/StateToReproduce.java
```diff
@@ -1,37 +1,40 @@
...
 import java.util.List;
 
 import sqlancer.common.query.Query;
 
-public class StateToReproduce {
+public class StateToReproduce implements Serializable {
+    private static final long serialVersionUID = 1L;
 
-    private List<Query<?>> statements = new ArrayList<>();
+    private transient List<Query<?>> statements = new ArrayList<>();
```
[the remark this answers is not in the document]
jinhuix: The PR passes locally using the `javac` compiler (`mvn -B verify -DskipTests=true`), but CI uses ECJ which does not support `<arg>-Xlint:-serial</arg>`, causing the build to fail.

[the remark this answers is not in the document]
KabilanMA: That build failure can be a bit problematic compared to the problem of adding a `serialVersionUID` to every subclass, I guess.

pom.xml
```diff
@@ -114,6 +114,7 @@
           <target>11</target>
           <compilerArguments>
             <properties>${project.basedir}/.settings/org.eclipse.jdt.core.prefs</properties>
+            <arg>-Xlint:-serial</arg>
```
mrigger, line 117: What does this change do? If it's unrelated, it would be better to open another PR.

src/sqlancer/StateToReproduce.java
```diff
@@ -1,37 +1,40 @@
...
+    private static final long serialVersionUID = 1L;
 
     private List<Query<?>> statements = new ArrayList<>();
 
     private final String databaseName;
 
-    private final DatabaseProvider<?, ?, ?> databaseProvider;
+    private transient DatabaseProvider<?, ?, ?> databaseProvider;
+
+    public String exception;
```
mrigger, line 25: I guess we can also remove this field?

src/sqlancer/postgres/PostgresProvider.java
```diff
@@ -279,14 +279,14 @@ public SQLConnection createDatabase(PostgresGlobalState globalState) throws SQLE
         }
         Connection con = DriverManager.getConnection("jdbc:" + entryURL, username, password);
         globalState.getState().logStatement(String.format("\\c %s;", entryDatabaseName));
-        
+
```
mrigger, line 282: It would be better to avoid committing such whitespace changes.

src/sqlancer/StateToReproduce.java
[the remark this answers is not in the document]
mrigger: I agree with Kabilan that adding the `serialVersionUID` would be a better change.

pom.xml
jinhuix: That was to suppress warnings about missing `serialVersionUID`. I’ll remove it.

mrigger:
Btw, adding a test in a follow-up PR would be quite useful.

PR #1270, opened 2025-08-24
Add a standalone reducer

jinhuix:
### Summary

This PR introduces **a standalone reducer**, independent of SQLancer, that reduces serialized `StateToReproduce` files (`.ser`) using partition-based delta debugging. The reducer progressively minimizes the query set while preserving the failure.

### Usage

Run the reducer with:

```bash
java -cp target/sqlancer-2.0.0.jar sqlancer.StandaloneReducer <path-to-ser-file>
```

### Example Output

```bash
Reduction completed successfully!
Final size: 1 statement (98.7% reduction)
Final queries:
INSERT INTO vt0(vt0) VALUES('integrity-check');
```

### DIFF
--- a/src/sqlancer/StandaloneReducer.java
+++ b/src/sqlancer/StandaloneReducer.java
@@ -1,0 +1,130 @@
+package sqlancer;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+import sqlancer.common.query.Query;
+
+/**
+ * A standalone tool to reduce bug-triggering SQL statements using the delta debugging algorithm.
+ */
+public class StandaloneReducer {
+    private int partitionNum = 2;
+    private final StateToReproduce originalState;
+    private final DatabaseProvider<?, ?, ?> databaseProvider;
+
+    public StandaloneReducer(Path serFilePath) throws Exception {
+        this.originalState = StateToReproduce.deserialize(serFilePath);
+        this.databaseProvider = originalState.getDatabaseProvider();
+        if (this.databaseProvider == null) {
+            throw new IllegalStateException("Failed to get database provider from .ser file");
+        }
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


src/sqlancer/StandaloneReducer.java
```diff
@@ -0,0 +1,130 @@
...
+                partitionNum = 2;
+            } else {
+                if (partitionNum >= queries.size()) {
+                    break;
+                }
+                partitionNum = Math.min(partitionNum * 2, queries.size());
+            }
+        }
+
+        System.out.println("Reduction completed successfully!");
```
mrigger, line 57: Printing both statistics and the output on stdout might make it difficult to extract the queries. I'd propose either one of the following:
1. Deleting the statistics output
2. Writing the final queries to a file rather than printing them to stdout

src/sqlancer/StandaloneReducer.java
```diff
@@ -0,0 +1,130 @@
...
+            G globalState = typedProvider.getGlobalStateClass().getDeclaredConstructor().newInstance();
+
+            try (C connection = typedProvider.createDatabase(globalState)) {
+                globalState.setConnection(connection);
+                for (Query<?> query : queries) {
+                    try {
+                        Query<C> typedQuery = (Query<C>) query;
+                        typedQuery.execute(globalState);
+                    } catch (Throwable e) {
+                        // Any exception is considered a success
```
mrigger, line 103: I was a bit confused after reading this comment, but I think the code is doing the right thing. Perhaps the comment could be "Any exception not declared as an expected error by the query indicates that an (unexpected) exception still exists"

src/sqlancer/StandaloneReducer.java
jinhuix: Thanks for the feedback! I updated the code:

1. Only printed statistics to stdout, while writing final queries to a .sql file.
2. Replaced the confusing comment.

PR #1274, opened 2025-09-18
tidb: skip Data too long errors

hawkingrei:
I found some errors, which seem to be caused by data generation issues rather than bugs in TiDB, so I think we can skip them.


hawkingrei:
@mrigger The output of TestTiDBTLP seems a bit strange. Is there a problem here?

image

PR #1275, opened 2025-09-18
Add DeleteGenerator for MariaDB

Mohab-Sobhy:
### Summary

This PR adds support for generating `DELETE` statements in MariaDB.

### Changes

* Added `MariaDBDeleteGenerator` for creating random `DELETE` queries.
* Integrated the new `DELETE` action into `MariaDBProvider`.
* Simplified imports in `MariaDBProvider` (`import sqlancer.mariadb.gen.*`).

### DIFF
--- a/src/sqlancer/mariadb/MariaDBProvider.java
+++ b/src/sqlancer/mariadb/MariaDBProvider.java
@@ -19,13 +19,7 @@ import sqlancer.SQLProviderAdapter;
 import sqlancer.common.DBMSCommon;
 import sqlancer.common.query.SQLQueryAdapter;
 import sqlancer.mariadb.MariaDBProvider.MariaDBGlobalState;
-import sqlancer.mariadb.gen.MariaDBIndexGenerator;
-import sqlancer.mariadb.gen.MariaDBInsertGenerator;
-import sqlancer.mariadb.gen.MariaDBSetGenerator;
-import sqlancer.mariadb.gen.MariaDBTableAdminCommandGenerator;
-import sqlancer.mariadb.gen.MariaDBTableGenerator;
-import sqlancer.mariadb.gen.MariaDBTruncateGenerator;
-import sqlancer.mariadb.gen.MariaDBUpdateGenerator;
+import sqlancer.mariadb.gen.*;
 
 @AutoService(DatabaseProvider.class)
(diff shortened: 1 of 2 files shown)


hawkingrei:
```
 Error:  Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.20.0:validate (eclipseformat) on project sqlancer: File '/home/runner/work/sqlancer/sqlancer/src/sqlancer/mariadb/gen/MariaDBDeleteGenerator.java' has not been previously formatted. Please format file (for example by invoking `mvn net.revelc.code.formatter:formatter-maven-plugin:2.20.0:format`) and commit before running validation! -> [Help 1]
```

Please run ``` mvn formatter:format``` to fix the linter.

mrigger:
Thanks for the PR! Feel free to ping us for reviews (at least I don't get automatic notifications).

Based on the CI/CD checks, the change causes some tests to fail. Could you please take a look?

Mohab-Sobhy:
@mrigger Thanks for the feedback!

I’ve addressed the Checkstyle and MariaDB issues, and the corresponding `General tests` and `DBMS Tests (MariaDB)` should now be fixed.

Some of the required checks (e.g., `Materialize`, `TiDB`, `YugabyteDB`) don’t appear to be related to my changes. For instance, the Materialize job fails during setup with the following error:

```
ERROR: syntax error at or near "SETTING"
LINE 1: SET CLUSTER SETTING sql.stats.forecasts.enabled = false
```

This seems to be an environment or version issue in the CI setup rather than something caused by the changes in the MariaDB generator.  
Could you please confirm if that’s expected or if I should investigate further?

src/sqlancer/mariadb/MariaDBProvider.java
```diff
@@ -77,6 +79,9 @@ public void generateDatabase(MariaDBGlobalState globalState) throws Exception {
             case CREATE_INDEX:
                 nrPerformed = globalState.getRandomly().getInteger(0, 2);
                 break;
+            case DELETE:
```
mrigger: Perhaps we could also generate them with a low probability? This logic will include 10 `DELETE` statements into every test case, rather than probabilistically generate some for some test cases, so tables are likely to be empty for many test cases.

PR #1277, opened 2025-10-26
Add initial QuestDB support with simple SELECT generator (Fixes #457)

### LINKED ISSUE #457: Add support for QuestDB
See https://github.com/questdb/questdb/issues/595.

Swastik092:
This PR adds initial support for QuestDB in SQLancer.

Changes included:
- Created `QuestDBQueryGenerator.java` to generate simple SELECT queries.
- Updated `QuestDBProvider.java` to use the new generator.
- Ensured queries are executed against a running QuestDB instance.
- Limited multi-threading (`--num-threads 1`) for stability during testing.

Currently, only basic SELECT queries are supported. INSERT, DELETE, and more complex queries will be added in future iterations.

This addresses Issue #457 and lays the foundation for full QuestDB testing support.


src/sqlancer/questdb/QuestDBTest.class
mrigger: Seems like an accidental commit.

mrigger (commented):
Thanks, this is great! We currently don't have a CI/CD test for QuestDB. Do you think it would be possible to add one first before adding new features?

.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,114 +0,0 @@
-eclipse.preferences.version=1
```
mrigger: Can you revert this change?

src/sqlancer/questdb/QuestDBErrors.java
```diff
@@ -42,15 +56,30 @@ public static void addGroupByErrors(ExpectedErrors errors) {
     public static List<String> getInsertErrors() {
         ArrayList<String> errors = new ArrayList<>();
 
-        // TODO (anxing)
+        // QuestDB-specific insert errors
         errors.add("Invalid column");
         errors.add("inconvertible types:");
         errors.add("inconvertible value:");
+        errors.add("column count mismatch");
```
mrigger: The "column count mismatch" seems like a bug in the generator.

src/sqlancer/questdb/QuestDBErrors.java
```diff
@@ -42,15 +56,30 @@ public static void addGroupByErrors(ExpectedErrors errors) {
...
+        errors.add("duplicate key value");
+        errors.add("constraint violation");
+        errors.add("invalid data type");
+        errors.add("value too large");
+        errors.add("missing required column");
+        errors.add("table does not exist");
+        errors.add("permission denied");
+        errors.add("syntax error in INSERT statement");
```
mrigger: Syntax errors are always unexpected, so we should fix this in the generator.

src/sqlancer/questdb/QuestDBProvider.java
```diff
@@ -59,6 +61,10 @@ private static int mapActions(QuestDBGlobalState globalState, Action a) {
             return r.getInteger(0, 3);
         case TRUNCATE:
             return r.getInteger(0, 5);
+        case UPDATE:
```
mrigger: I think we should only add new features once we have a CI/CD that works for a given QuestDB version.

PR #1318, opened 2026-04-08
Materialize: Update expected query plan for QPG test

mrigger:
The latest Materialize version changed the EXPLAIN OPTIMIZED PLAN output format: the With block now appears before Return, and column name annotations shifted positions. Update the expected string to match.


def- (approved):
Indeed, thanks!

PR #1336, opened 2026-04-22
CI: cache pinned Docker images for ClickHouse, OceanBase, and TiDB

mrigger:
Saves the pulled image as a tarball via actions/cache, then docker-load on subsequent runs. Only applied to jobs with version-pinned tags; jobs using :latest or service-declared containers are skipped since caching there would either serve stale images or run before any step.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -127,9 +127,21 @@ jobs:
           cache: 'maven'
       - name: Build SQLancer
         run: mvn -B package -DskipTests=true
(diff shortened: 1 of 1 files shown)


mrigger:
Closing — per-job timing comparison shows caching 250-600MB docker images is neutral-to-slower than direct `docker pull` from Docker Hub. See PR discussion. Pivoting to caching the Rust build and raw tarball downloads, where the benefit is clearer.

PR #1345, opened 2026-06-19
Fix YugabyteDB CI by adding ysql_yb_enable_listen_notify flag

tlmorgan24:
## Summary

The YugabyteDB CI job was failing because SQLancer generates `LISTEN`, `NOTIFY`, and `UNLISTEN` statements, but YugabyteDB has this feature disabled by default. This caused an `AssertionError` each time one of these statements was executed, terminating the test run.

The fix enables the `ysql_yb_enable_listen_notify` runtime flag on both the tserver and master when starting YugabyteDB in CI, allowing these statements to execute as intended.

## Changes

- `.github/workflows/main.yml`: Pass `--tserver_flags` and `--master_flags` to `yugabyted start` to enable `ysql_yb_enable_listen_notify`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -648,6 +648,7 @@ jobs:
           docker pull yugabytedb/yugabyte:latest
           docker run -d --name yugabyte -p7000:7000 -p9000:9000 -p5433:5433 -p9042:9042 yugabytedb/yugabyte:latest bin/yugabyted start --daemon=false --tserver_flags="ysql_yb_enable_listen_notify=true" --master_flags="ysql_yb_enable_listen_notify=true"
           until pg_isready -h localhost -p 5433; do sleep 1; done
+          until nc -z localhost 9042; do sleep 1; done
       - name: Run Tests
         run: |
           YUGABYTE_AVAILABLE=true mvn -Djacoco.skip=true -Dtest=TestYSQLNoREC test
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -646,8 +646,9 @@ jobs:
       - name: Set up Yugabyte
         run: |
           docker pull yugabytedb/yugabyte:latest
-          docker run -d --name yugabyte -p7000:7000 -p9000:9000 -p5433:5433 -p9042:9042 yugabytedb/yugabyte:latest bin/yugabyted start --daemon=false
+          docker run -d --name yugabyte -p7000:7000 -p9000:9000 -p5433:5433 -p9042:9042 yugabytedb/yugabyte:latest bin/yugabyted start --daemon=false --tserver_flags="ysql_yb_enable_listen_notify=true" --master_flags="ysql_yb_enable_listen_notify=true"
           until pg_isready -h localhost -p 5433; do sleep 1; done
+          until nc -z localhost 9042; do sleep 1; done
```
mrigger, line 651: Is this waiting loop also needed?

tlmorgan24: It is not related to the ysql_yb_enable_listen_notify fix, but it is another robustness improvement to the YugabyteDB CI.

The CI previously had only `until pg_isready -h localhost -p 5433; do sleep 1; done`, which confirms YSQL on port 5433 is up, but does not necessarily imply YCQL on port 9042 is up (which gets tested in TestYCQL). So without the additional wait, it is possible for TestYCQL to fail due to connection refused.

mrigger: Okay, sounds good. Thanks for the explanation!

PR #1346, opened 2026-07-11
Remove dead oracle base classes: `NoRECBase` and `DocumentRemovalOracleBase`

tlmorgan24:
This PR deletes two abstract base classes in `src/sqlancer/common/oracle/`:

- `NoRECBase.java`
- `DocumentRemovalOracleBase.java`

Both classes had zero subclasses and zero references outside their own files, confirmed by a full grep across the source tree. These may be remnants of incomplete or speculative work. The standard software engineering approach is to remove dead code because version control already archives it. If either class is needed in the future, it can be recovered from git history. If the main source stores dead code, it can risk misleading contributors and introducing unnecessary cognitive load. 

This said, there may be good reasons for keeping these particular classes that I am not aware of, so please let me know your thoughts.


mrigger (approved):
LGTM, thanks!

PR #1347, opened 2026-07-15
Fix bugs in reproducer and test-case reduction

tlmorgan24:
## Bug 1

`TLPWhereOracle.check()` assigned the `TLPWhereReproducer` after calling
`ComparatorHelper.assumeResultSetsAreEqual`:

```java
ComparatorHelper.assumeResultSetsAreEqual(...); // throws AssertionError on mismatch
reproducer = new TLPWhereReproducer(...);  // only reached if error not thrown
```

`check()` also resets `reproducer = null` at its very start, so when `ProviderAdapter`
caught the `AssertionError` and called `oracle.getLastReproducer()`, it always received
`null` for a bug-finding invocation.

So, `--use-reducer` did nothing for `TLPWhereOracle`.

### Fix

Moved the `TLPWhereReproducer` construction to before `assumeResultSetsAreEqual`. All the
data it captures is already computed at that point, so the reproducer is populated regardless
of whether the comparison throws:

```java
reproducer = new TLPWhereReproducer(...);  // set before comparison
ComparatorHelper.assumeResultSetsAreEqual(...);
```

This matches the pattern already used correctly by `NoRECOracle`, which constructs its
`NoRECReproducer` inside the failure branch before throwing.

## Bug 2

Even after fixing bug 1, logs always ended up empty.

Two `StateLogger` instances point at the same reduce file: the executor's `logger`, and a
second logger created for `newGlobalState`, which is the one the reducers actually write
through. `StateLogger.getReduceFileWriter()` opens the file lazily and in truncate mode
(`new FileWriter(reduceFile, false)`).

After reduction finished, the cleanup code in `Main.run()` called:

```java
try {
    logger.getReduceFileWriter().close();
    logger.reduceFileWriter = null;
} catch (IOException e) {
    throw new AssertionError(e);
}
```

`logger` had never opened its reduce writer, so `getReduceFileWriter()`
opened the file, truncating everything the reducers had just written through
`newGlobalState`'s logger.

### Fix

Check the field instead of calling the lazy getter, so a writer is never opened purely to
be closed:

```java
if (logger.reduceFileWriter != null) {
    try {
        logger.reduceFileWriter.close();
        logger.reduceFileWriter = null;
    } catch (IOException e) {
        throw new AssertionError(e);
    }
}
```

Note, as part of Bug 5's fix, `logger` in this block was ultimately changed to `reduceLogger`.

## Bug 3

Reductions collapsed to a single statement (e.g. a lone `INSERT` statement) that did not
reproduce the original bug.

This was because the reproducers did not distinguish which failure a reduced candidate triggers. Any
failure counted as "bug still triggers":

- `TLPWhereReproducer.bugStillTriggers` wrapped both query execution and the result-set
  comparison in one `catch (AssertionError) { return true; }`. But
  `ComparatorHelper.getResultSetFirstColumnAsString` also wrapped every unexpected SQL
  error in an `AssertionError`. So when the reducer deleted the `CREATE TABLE`, the oracle
  queries failed with "no such table", which was caught and reported as the bug still
  triggering. The reducer therefore kept deleting statements until nothing meaningful was
  left.
- `NoRECReproducer` had a milder variant: `countRows`/`extractCounts` returned `-1` when a
  query fails or returns no result set, and the reproducer compared raw counts, so
  `-1 != <real count>` was accepted as a count mismatch.

### Fix

Reproducers are now category-aware: they record which failure `check()` originally
observed and accept a reduced candidate only if it triggers the same category.

- If the original bug was a result-set mismatch, the queries must execute
  successfully on the reduced database and then disagree. Any execution error
  (`SQLException`, `IgnoreMeException`, or the `AssertionError` wrapping an unexpected
  DBMS error) means the candidate is rejected. A `-1` count in NoREC is likewise rejected,
  mirroring `check()`, which throws `IgnoreMeException` in that case.
- If the original bug was an unexpected DBMS error, the reproducer stores the error's
  message (extracted from the wrapping `AssertionError`'s cause chain by the new
  `TestOracleUtils.getUnexpectedErrorMessage`) and accepts a candidate only if the queries
  fail with the identical message.

This also extends reducer support: previously, a DBMS-error bug produced no reproducer
at all in `NoRECOracle` (its reproducer was only constructed in the count-mismatch branch)
and in `TLPWhereOracle` when the error occurred on the original (no-WHERE) query. Both
paths now construct a reproducer tagged with the expected error message, so those bugs are
reducible too.

## Bug 4

`TLPWhereReproducer` compared against a stale baseline. It stored the result set of the
original query as captured on the original database, and compared it with the combined
TLP result set computed on the reduced database. Removing a legitimate statement (e.g.
an `INSERT`) changes both result sets on the reduced database, so the stale-vs-fresh
comparison reported a "mismatch" even on a bug-free DBMS.

### Fix

`bugStillTriggers` now re-executes the original query on the reduced database and compares fresh against fresh.

## Problem 5 - lack of clarity in main log

With `--use-reducer`, the main log (outside the `reduce/` folder) was not reproducible: it
contained only the reduced generation statements; it omitted the database setup
(`DROP DATABASE IF EXISTS ...; CREATE DATABASE ...; USE ...;`) and queries.

### Fix

`Main.run()` now partitions the statement list before reduction and reassembles it after:

- The setup-statement count is snapshotted right after `createDatabase`, and the oracle query comments are identified via the oracle run's local state (new `OracleRunReproductionState.getStatements()` accessor).
- Only the generation statements are handed to the reducers, so the setup and queries remain as-is.
- Afterwards the list is reassembled as: setup + reduced generation statements + the original commented oracle queries + the reproducer's bug description.

The main log now has similar information to one produced without `--use-reducer`, the only difference being that the generation statements are a final reduced batch. It therefore contains enough information to fully reproduce the bug, while the reduce-folder log shows the reduction steps in detail.

## Problem 6 - lack of clarity in reduce-folder log

The reduce-folder log was an unseparated stream of attempts, which made it hard to tell attempts apart.

### Fix

Each successful reduction attempt is now logged as a delimited block. E.g.:

```
-- Reduction attempt 3: the bug was still triggered with the following statements:
DROP DATABASE IF EXISTS database0;
CREATE DATABASE database0;
USE database0;
CREATE TABLE t0 (...);
...
-- On the database set up by the statements above, the following queries trigger an unexpected error with message: <message>
-- SELECT ...;

-- Reduction attempt 4: ...
```

This better reflects what actually happens on every attempt.

# Impact

With these fixes, `--use-reducer` produces usable results for `TLPWhereOracle` and
`NoRECOracle` (and the oracles that delegate to them, e.g. Presto's and YugabyteDB's TLP
WHERE oracles):

- a reduced test case is guaranteed to trigger the same category of failure as the original bug
- DBMS-error bugs, as opposed to only oracle logic bugs, are now reducible (previously they either had no reproducer or were reduced against the wrong criterion)
- when using --use-reducer, the main log now contains sufficient information to reproduce the bug, just like it does when --use-reducer is not activated.
- the detailed log in the reduce subfolder is now more readable and clearly shows the reduction process.

### DIFF
--- a/src/sqlancer/ASTBasedReducer.java
+++ b/src/sqlancer/ASTBasedReducer.java
@@ -103,8 +103,7 @@ public class ASTBasedReducer<G extends GlobalState<O, ?, C>, O extends DBMSSpeci
                     });
 
                     if (!initFlag) {
-                        newGlobalState.getLogger()
-                                .logReducer("warning: failed parsing the statement at transformer : " + t);
+                        System.out.println("Error when parsing the statement at transformer :" + t);
                         continue;
                     }
                     t.apply();
@@ -114,7 +113,8 @@ public class ASTBasedReducer<G extends GlobalState<O, ?, C>, O extends DBMSSpeci
         } while (observeChange);
 
         newGlobalState.getState().setStatements(new ArrayList<>(reducedStatements));
-        newGlobalState.getLogger().logReduced(newGlobalState.getState());
+        newGlobalState.getLogger().logReduced(newGlobalState.getState(),
+                "AST-based reduction finished; the following statements remain");
     }
 
     public boolean bugStillTriggers() throws Exception {
--- a/src/sqlancer/Main.java
+++ b/src/sqlancer/Main.java
@@ -80,6 +80,9 @@ public final class Main {
         private FileWriter queryPlanFileWriter;
         private FileWriter reduceFileWriter;
         private Path reproduceFilePath;
+        private List<Query<?>> reduceSetupStatements;
+        private String reduceBugInformation;
+        private int nrReductionAttempts;
 
         private static final List<String> INITIALIZED_PROVIDER_NAMES = new ArrayList<>();
         private final boolean logEachSelect;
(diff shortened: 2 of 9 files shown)


mrigger (approved):
Thanks a lot, this is great! I had a look at the code, and it looks good to me. I don't understand all the detailed changes, though. One alternative approach that would have made it easier to review the changes and also ensure that things won't be broken in the future would be the following:
1. Create a commit with a test case exercising the behavior that fails
2. Create another commit that addresses the issue to show that the test case now passes

If you open one PR for each issue (or perhaps bundling only a few related ones), reviewing each would be straightforward. I can see that this might be difficult for the reducer, as it operates on logging files, which might be more of a hassle to test. And, as you realized, the logic was quite broken anyway, so I can see that it needed multiple fixes to achieve a meaningful output.

PR #1348, opened 2026-07-27
Fix MySQL index hint syntax in the DQP hint generator

splf-jojo:
## Problem

`MySQLHintGenerator.indexesHint()` puts a comma between the table name and the first index name:

```java
sb.append(table.getName());
sb.append(", ");
```

The [documented grammar](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html) is `hint_name([@query_block_name] tbl_name [index_name [, index_name] ...])`, so a space separates the table from the first index, while commas separate index names from one another. The manual's own example shows both at once: `INDEX(t1 f3, PRIMARY)`.

The generator therefore emits ``NO_INDEX(t0, `PRIMARY`)``, which MySQL rejects with warning 1064 and silently ignores. SQLancer does not inspect warnings, so this stays invisible during a run.

Only the first separator is wrong. The `Collectors.joining(", ")` between index names is correct and is left unchanged.

## Reproduction

Verified on MySQL 8.1.0 and on 9.7.0, the version used by the CI job. `FORMAT=TRADITIONAL` is specified because the default `EXPLAIN` format differs between the two.

```sql
CREATE DATABASE hinttest; USE hinttest;
CREATE TABLE h0(id INT PRIMARY KEY AUTO_INCREMENT, c0 INT, c1 INT, INDEX idx_c0(c0));
SET SESSION cte_max_recursion_depth = 10000;
INSERT INTO h0(c0, c1)
  WITH RECURSIVE seq(n) AS (SELECT 1 UNION ALL SELECT n+1 FROM seq WHERE n < 5000)
  SELECT n, n % 100 FROM seq;
ANALYZE TABLE h0;

EXPLAIN FORMAT=TRADITIONAL SELECT c1 FROM h0 WHERE c0 = 123;
EXPLAIN FORMAT=TRADITIONAL SELECT /*+ NO_INDEX(h0, `idx_c0`) */ c1 FROM h0 WHERE c0 = 123; SHOW WARNINGS;
EXPLAIN FORMAT=TRADITIONAL SELECT /*+ NO_INDEX(h0 `idx_c0`) */ c1 FROM h0 WHERE c0 = 123; SHOW WARNINGS;
```

| Hint | `type` | `key` | Warning |
| --- | --- | --- | --- |
| none | `ref` | `idx_c0` | none |
| `NO_INDEX(h0, idx_c0)` (current) | `ref` | `idx_c0` | **1064** |
| `NO_INDEX(h0 idx_c0)` (with this change) | `ALL` | `NULL` | none |

The current form yields exactly the plan produced without any hint. `SHOW WARNINGS` after the `EXPLAIN` also returns Note 1003 with the statement the server will actually execute. With the current syntax the hint is absent from it entirely, while with the fix it is present and normalized by the server as ``/*+ NO_INDEX(`h0`@`select#1` `idx_c0`) */``.

## Scope

16 of the 32 kinds in `OptimizeHint` are generated through `indexesHint()`: `INDEX`, `GROUP_INDEX`, `JOIN_INDEX`, `INDEX_MERGE`, `MRR`, `ORDER_INDEX`, `SKIP_SCAN` together with their `NO_` counterparts, plus `NO_ICP` and `NO_RANGE_OPTIMIZATION`. In an `--oracle DQP` run on MySQL 8.0.34 (1376 oracle iterations), 22,016 of 44,032 hinted queries carried the malformed syntax and ran without affecting the plan.

The other 16 kinds go through `tablesHint()` and `semiHint()`, where the comma separates table names and is correct. They are unaffected.

## Verification

`TestMySQLDQP` passes before and after the change (301 s, exit code 0 in both cases). `mvn verify -DskipTests` passes. The reproduction above was checked on both MySQL 8.1.0 and 9.7.0.

### DIFF
--- a/src/sqlancer/mysql/gen/MySQLHintGenerator.java
+++ b/src/sqlancer/mysql/gen/MySQLHintGenerator.java
@@ -170,7 +170,7 @@ public class MySQLHintGenerator {
         MySQLTable table = Randomly.fromList(tables);
         List<MySQLIndex> allIndexes = table.getIndexes();
         sb.append(table.getName());
-        sb.append(", ");
+        sb.append(" ");
         if (allIndexes.isEmpty()) {
             sb.append("PRIMARY");
         } else {


mrigger:
LGTM, thanks!

PR #1350, opened 2026-07-29
Do not schedule disabled PostgreSQL tablespaces

splf-jojo:
## Summary

When PostgreSQL tablespace testing is disabled, `PostgresProvider.mapActions()`
now assigns zero executions to `CREATE_TABLESPACE`. Consequently,
`StatementExecutor` never selects that action and never calls
`PostgresTableSpaceGenerator.generate()`.

Following the review feedback, this moves the guard up to the scheduling layer,
rather than handling an already-selected action inside the generator.

This removes the previous generator-level `IgnoreMeException` approach. The
generator is now skipped entirely when tablespace testing is disabled.

## Testing

- `mvn -Dtest=sqlancer.postgres.TestPostgresProvider test`
  - `TestPostgresProvider`: 1 test passed, 0 failures, 0 errors.

### DIFF
--- a/src/sqlancer/postgres/PostgresProvider.java
+++ b/src/sqlancer/postgres/PostgresProvider.java
@@ -194,7 +194,7 @@ public class PostgresProvider extends SQLProviderAdapter<PostgresGlobalState, Po
             nrPerformed = r.getInteger(0, 2);
             break;
         case CREATE_TABLESPACE:
-            nrPerformed = r.getInteger(0, 2);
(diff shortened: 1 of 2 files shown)


mrigger:
Thanks for the PR! I think it might be better to solve this issue at a higher level. If tablespaces are disabled, I think that `PostgresTableSpaceGenerator.generate()` should never be called by the `StatementExecutor`. Could you check whether this would be possible?

mrigger (approved):
LGTM, thanks for the PR. Also great that you added the test case to ensure we won't have a regression. Will merge after the checks pass.

PR #1352, opened 2026-07-31
Add support for DECIMAL(M, D) columns in MySQL EET

tlmorgan24:
## Summary

The EET oracle's type inference previously required `FLOAT`/`DOUBLE`/`DECIMAL` columns to be created without an `(M, D)` precision/scale specifier, because its type-pinning `CAST` (used for the redundant branch of transformation rules No. 3–4) could only target the bare types.

This PR removes the exclusion for `DECIMAL` by tracking a column's exact `(M, D)` end-to-end and reproducing it as `CAST(... AS DECIMAL(M, D))`. `FLOAT`/`DOUBLE` remain excluded under EET. This is because `FLOAT(M, D)`/`DOUBLE(M, D)` is not a valid `CAST` target, so there is no way to reproduce such a column's exact type in a type-pinning cast (also note `FLOAT(M, D)`/`DOUBLE(M, D)` columns are deprecated anyway, but they still remain supported for all other oracles).

### Schema — track scale (`MySQLSchema.java`)
- Added a `scale` field (+ `getScale()`) to `MySQLColumn`, read from `information_schema.columns.NUMERIC_SCALE`, alongside the already-read `NUMERIC_PRECISION`.
- The `MySQLColumn` constructor and its single call site now thread `scale` through. `information_schema` always reports a concrete `(M, D)` for a `DECIMAL` column, defaulting to `(10, 0)` when the column was declared without one.

### Type domain — enrich `CastType` (`MySQLCastOperation.java`)
- `MySQLCastOperation.CastType` was changed from a plain `enum` into a small final value class. Its keyword kinds (`SIGNED`, `UNSIGNED`, `CHAR`, `FLOAT`, `DOUBLE`, `DECIMAL`) remain `public static final` singletons (so existing e.g. `== CastType.SIGNED` comparisons are unaffected), while `CastType.decimal(m, d)` produces `DECIMAL` instances carrying `(M, D)`.
- Added `getPrecision`/`getScale`, `equals`/`hashCode` (over kind + `(M, D)`), and a `toString` that renders `DECIMAL(M, D)`.

### Transformer (`MySQLEETTransformer.java`)
- `inferColumnType` maps a `DECIMAL` column to `CastType.decimal(column.getPrecision(), column.getScale())`.
- `commonType` compares with `.equals()` instead of `==`, so two `DECIMAL` subexpressions with matching `(M, D)` (distinct instances) share a type. If `(M, D)` differs, `commonType` conservatively returns `null`, as MySQL's precise aggregation rule in this case is not known.

### Table generation (`MySQLTableGenerator.java`)
- `DECIMAL` has now been reverted to the original behaviour of calling `optionallyAddPrecisionAndScale` (i.e. `(M, D)` is now allowed under EET).
- `FLOAT`/`DOUBLE` route through a new `optionallyAddFloatingPointPrecisionAndScale`, which omits `(M, D)` only while EET is active.

### DIFF
--- a/src/sqlancer/mysql/MySQLSchema.java
+++ b/src/sqlancer/mysql/MySQLSchema.java
@@ -57,6 +57,7 @@ public class MySQLSchema extends AbstractSchema<MySQLGlobalState, MySQLTable> {
 
         private final boolean isPrimaryKey;
         private final int precision;
+        private final int scale;
 
         public enum CollateSequence {
             NOCASE, RTRIM, BINARY;
@@ -66,14 +67,19 @@ public class MySQLSchema extends AbstractSchema<MySQLGlobalState, MySQLTable> {
             }
         }
 
-        public MySQLColumn(String name, MySQLDataType columnType, boolean isPrimaryKey, int precision) {
+        public MySQLColumn(String name, MySQLDataType columnType, boolean isPrimaryKey, int precision, int scale) {
             super(name, null, columnType);
             this.isPrimaryKey = isPrimaryKey;
             this.precision = precision;
+            this.scale = scale;
         }
 
         public int getPrecision() {
             return precision;
+        }
+
+        public int getScale() {
+            return scale;
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


mrigger (approved):
LGTM, thanks!

PR #1356, opened 2026-08-07
Implement EET DML oracle (initially for DELETE statements only)

tlmorgan24:
## Summary

The EET oracle previously only transformed `SELECT` queries and compared the two result sets. This PR adds a DML variant, `EETDMLOracle`, that transforms a DML statement's expressions and compares the two database states the original and transformed statements leave behind.

To observe the database state, a similar approach to the DQE oracle is used: add an auxiliary column that uniquely identifies each row, and run each statement inside a transaction that is rolled back. This way, the original and transformed statements can each be measured against the same starting state without permanently mutating the database.

Only `DELETE` is implemented in this PR, as it is the simplest (its only transformable site is the `WHERE` predicate, and row identity alone is a sufficient comparison surface). The generic infrastructure is designed so `UPDATE` and `INSERT` can be added later without rework. MySQL is included as a specific DBMS implementation.

The expression transformer (`EETTransformer` / `MySQLEETTransformer`) is statement-agnostic and remains unchanged.

### Dedicated new oracle (`common/oracle/EETDMLOracle.java`)

`check()`:
1. Pick a single target table from the schema and confine the generator's column pool to it.
2. Generate the `WHERE` predicate `p`; transform it in a boolean context (`transformer.transform(p, true)`).
3. Assemble the original and transformed `DELETE` strings via the generator.
4. `ADD COLUMN rowid`; then, inside a `try` whose `finally` drops the column, stamp every row once with a unique id.
5. For each `DELETE` (original, then transformed): `BEGIN`; run it; snapshot the surviving `rowid`s; `ROLLBACK` (rollback in a `finally`).
6. Compare the two surviving-`rowid` sets; on mismatch throw an `AssertionError` whose message embeds both statement strings.

Error handling mirrors the SELECT oracle: an expected DBMS error (on the `ExpectedErrors` allow-list) aborts the iteration via `IgnoreMeException`, while an unexpected error surfaces as a reported bug.

Test case reduction will be implemented at a later stage.

### New generator interface (`common/gen/EETDMLGenerator.java`)

A sibling of `EETGenerator`, supplying the DBMS-specific pieces the DML oracle needs.

- Abstract methods: `generateBooleanExpression()` (the predicate), `createTransformer()`, `setTablesAndColumns(...)`, `asString(E)` (render an expression), and `stampRowIdsStatement(T)` (assign every existing row a unique id).
- Default methods: `addRowIdColumnStatement` / `dropRowIdColumnStatement` / `selectRowIdsStatement` / `deleteStatement` / `beginTransactionStatement` / `rollbackTransactionStatement`.

### MySQL generator (`mysql/gen/MySQLExpressionGenerator.java`)

Now also implements `EETDMLGenerator<MySQLExpression, MySQLTable, MySQLColumn>`. It reuses the existing `generateBooleanExpression` / `createTransformer` / `setTablesAndColumns` and adds only `asString` and `stampRowIdsStatement`.

### Oracle registration (`mysql/MySQLOracleFactory.java`)

New `EET_DML` enum entry that constructs an `EETDMLOracle` with existing expected errors plus a couple of extra ones encountered.

### Table generation (`mysql/gen/MySQLTableGenerator.java`)

The `ENGINE` table option is forced to `InnoDB` if EET DML is being used (checked with `usesEETDML()` method). This is because it is the only transactional engine, and the test procedure relies on transactions.

## Follow-ups (to be integrated into same oracle)

- `UPDATE` support
- `INSERT` support
- Test case reduction, modelled on `EETReproducer` used for the SELECT oracle.

### DIFF
--- a/src/sqlancer/common/gen/EETDMLGenerator.java
+++ b/src/sqlancer/common/gen/EETDMLGenerator.java
@@ -77,7 +77,7 @@ public interface EETDMLGenerator<E extends Expression<C>, T extends AbstractTabl
 
     /**
      * SQL that assigns every existing row of {@code table} a distinct, stable identifier in the {@link #ROW_ID_COLUMN}
-     * column. DBMS-specific because it names the DBMS's UUID-generating function.
+     * column. For example, a 36-character UUID string.
      *
      * @param table
      *            the table whose rows are stamped
@@ -86,10 +86,19 @@ public interface EETDMLGenerator<E extends Expression<C>, T extends AbstractTabl
      */
     String stampRowIdsStatement(T table);
 
+    /**
+     * The SQL type of the auxiliary {@link #ROW_ID_COLUMN} column. It must be able to hold the identifiers that
+     * {@link #stampRowIdsStatement} produces, so it belongs with that statement as the other half of the row-id
+     * representation. For example, {@code VARCHAR(36)} would fit a 36-character UUID string.
+     *
+     * @return the column type
+     */
+    String rowIdColumnType();
+
     // --- Standard-SQL statements (override only where the DBMS's dialect differs) ---
 
     /**
-     * SQL that adds the auxiliary {@link #ROW_ID_COLUMN} column to {@code table}.
+     * SQL that adds the auxiliary {@link #ROW_ID_COLUMN} column to {@code table}, typed as {@link #rowIdColumnType}.
      *
      * @param table
      *            the table to add the column to
@@ -97,7 +106,7 @@ public interface EETDMLGenerator<E extends Expression<C>, T extends AbstractTabl
      * @return the SQL statement
      */
     default String addRowIdColumnStatement(T table) {
-        return "ALTER TABLE " + table.getName() + " ADD COLUMN " + ROW_ID_COLUMN + " VARCHAR(36)";
+        return "ALTER TABLE " + table.getName() + " ADD COLUMN " + ROW_ID_COLUMN + " " + rowIdColumnType();
     }
 
     /**
--- a/src/sqlancer/common/gen/EETDMLGenerator.java
+++ b/src/sqlancer/common/gen/EETDMLGenerator.java
@@ -1,5 +1,8 @@
 package sqlancer.common.gen;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import sqlancer.common.ast.newast.Expression;
 import sqlancer.common.oracle.EETTransformer;
 import sqlancer.common.schema.AbstractTable;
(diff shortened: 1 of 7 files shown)


src/sqlancer/common/gen/EETDMLGenerator.java
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * SQL that adds the auxiliary {@link #ROW_ID_COLUMN} column to {@code table}.
+     *
+     * @param table
+     *            the table to add the column to
+     *
+     * @return the SQL statement
+     */
+    default String addRowIdColumnStatement(T table) {
+        return "ALTER TABLE " + table.getName() + " ADD COLUMN " + ROW_ID_COLUMN + " VARCHAR(36)";
```
mrigger: Hmm, that part seems MySQL-specific. Can we make this code more general?

src/sqlancer/common/gen/EETDMLGenerator.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+        return statement;
+    }
+
+    /**
+     * SQL that starts a transaction, so a statement's effect can be observed and then undone.
+     *
+     * @return the SQL statement
+     */
+    default String beginTransactionStatement() {
```
mrigger, line 177: Just to check, are there any database systems that SQLancer supports that support transactions, but not `BEGIN` and `ROLLBACK`? If not, we can keep it simple and omit these two methods.

src/sqlancer/common/oracle/EETDMLOracle.java
```diff
@@ -0,0 +1,173 @@
...
+        this.state = state;
+        this.gen = gen;
+        this.transformer = gen.createTransformer();
+        this.errors = expectedErrors;
+    }
+
+    @Override
+    public void check() throws SQLException {
+        List<T> tables = state.getSchema().getDatabaseTables();
+        if (tables.isEmpty()) {
```
mrigger, line 75: Are there possible situations where `tables` is empty?

src/sqlancer/common/gen/EETDMLGenerator.java
tlmorgan24: Most do support those statements, so in most cases they will just use the default method, but there are one or two exceptions. E.g. HSQLDB and H2 use `START TRANSACTION` instead of `BEGIN`.

src/sqlancer/common/oracle/EETDMLOracle.java
tlmorgan24: Whether `tables` is empty depends on DBMS-specific code:
* The generateDatabase method is DBMS-specific, and from what I've seen, the DBMSs do ensure at least one table is created.
* If QPG is enabled, its mutateTables method may cause an attempted drop of the last remaining table in the schema. The drop method itself is DBMS-specific, and from what I've seen, the DBMS which support QPG do ensure they do not drop the last remaining table.

So, it seems like `tables` will never be empty at the moment, but that is only a consequence of DBMS-specific code. The non-emptiness of tables is not guaranteed by any shared/generic code, so if somebody adds extra DBMS support, they would risk breaking that guarantee unless they are careful. So, this emptiness check is guarding against that eventuality.

src/sqlancer/common/gen/EETDMLGenerator.java
tlmorgan24: I have now addressed this in the latest commit pushed, let me know if it matches what you had in mind

mrigger (approved):
LGTM, thanks!

mrigger:
By the way, as a minor suggestion, it would be fine to only summarize the high-level changes and rationale within a short paragraph or so. If the change summary is too detailed, it's easier to just read the code changes.

PR #1357, opened 2026-08-08
Extend the EET DML oracle to UPDATE support

tlmorgan24:
The EET DML oracle (`EETDMLOracle`) previously transformed only `DELETE` statements. This PR adds `UPDATE` to the same oracle. The key difference from `DELETE` is that `UPDATE` has two kinds of transformable expression (the `WHERE` predicate and each `SET` value expression) rather than one.

Comparing which rows are affected is no longer enough (two runs could change the same rows but write different values, which we would want to flag as a bug). The comparison is therefore upgraded to a full **post-image** (for each surviving row, its identifier together with every content column's value). This covers both `UPDATE` and `DELETE` statements, subsuming `DELETE`'s previous surviving-row-set comparison. It will also, with small additions, be able to cover `INSERT` statements, which will be added next.

Effort has been made to make the logs for reported bugs clear, showing the post-image discrepancy of the relevant rows. Test case reduction is left to a later PR.

### DIFF
--- a/src/sqlancer/common/oracle/EETDMLOracle.java
+++ b/src/sqlancer/common/oracle/EETDMLOracle.java
@@ -104,23 +104,11 @@ public class EETDMLOracle<E extends Expression<C>, S extends AbstractSchema<?, T
             orderByColumns = Randomly.subset(table.getColumns());
         }
 
-        String originalStatement;
-        String transformedStatement;
-        if (Randomly.getBoolean()) {
-            // UPDATE also transforms the written values: each SET value expression is transformed in a scalar context.
-            List<Map.Entry<C, E>> assignments = gen.generateSetAssignments();
-            List<Map.Entry<C, E>> transformedAssignments = new ArrayList<>();
-            for (Map.Entry<C, E> assignment : assignments) {
-                E transformedValue = transformer.transform(assignment.getValue(), false);
-                transformedAssignments.add(new AbstractMap.SimpleEntry<>(assignment.getKey(), transformedValue));
-            }
-            originalStatement = gen.updateStatement(table, assignments, predicate, orderByColumns, limit);
-            transformedStatement = gen.updateStatement(table, transformedAssignments, transformedPredicate,
-                    orderByColumns, limit);
-        } else {
-            originalStatement = gen.deleteStatement(table, predicate, orderByColumns, limit);
-            transformedStatement = gen.deleteStatement(table, transformedPredicate, orderByColumns, limit);
-        }
+        StatementPair statements = Randomly.getBoolean()
+                ? generateUpdateStatements(table, predicate, transformedPredicate, orderByColumns, limit)
+                : generateDeleteStatements(table, predicate, transformedPredicate, orderByColumns, limit);
+        String originalStatement = statements.original;
+        String transformedStatement = statements.transformed;
         generatedQueryString = originalStatement;
 
         int columnCount = gen.postImageColumns(table).size();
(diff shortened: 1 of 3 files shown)


src/sqlancer/common/oracle/EETDMLOracle.java
```diff
@@ -93,77 +103,192 @@ public void check() throws SQLException {
...
-        // Add the auxiliary column outside the try, then guard everything after it with the finally that drops it:
-        // the ALTER auto-commits (it is not undone by ROLLBACK), so a failure between adding and dropping would leak
-        // the column into the next iteration and cause cascading duplicate-column failures. The row-identity setup
-        // touches rows, so — like the DELETE itself — it can raise tolerated errors (e.g. functional-index maintenance
-        // truncation); such an error aborts the iteration (IgnoreMeException) rather than being reported as a bug.
+
+        String originalStatement;
+        String transformedStatement;
+        if (Randomly.getBoolean()) {
+            // UPDATE also transforms the written values: each SET value expression is transformed in a scalar context.
```
mrigger: I think it would be nice to have a method whose name would make the two options immediately clear. However, I assume that is not straightforward given the two return values?

src/sqlancer/common/oracle/EETDMLOracle.java
```diff
@@ -93,77 +103,192 @@ public void check() throws SQLException {
...
+                state.getOptions().canonicalizeSqlString());
+        SQLancerResultSet result = null;
+        try {
+            result = q.executeAndGet(state);
+            if (result == null) {
+                throw new IgnoreMeException();
+            }
+            while (result.next()) {
+                List<String> row = new ArrayList<>(columnCount);
+                for (int i = 1; i <= columnCount; i++) {
```
mrigger, line 268: Just to double-check, starting from `1` here is intended?

src/sqlancer/common/oracle/EETDMLOracle.java
```diff
@@ -93,77 +103,192 @@ public void check() throws SQLException {
...
+     * @throws SQLException
+     *             if cleanup fails (errors thrown elsewhere will always be rethrown as {@link IgnoreMeException} or
+     *             {@link AssertionError})
+     */
+    private List<List<String>> snapshotPostImage(String selectStatement, int columnCount) throws SQLException {
+        List<List<String>> rows = new ArrayList<>();
+        SQLQueryAdapter q = new SQLQueryAdapter(selectStatement, errors, true,
+                state.getOptions().canonicalizeSqlString());
+        SQLancerResultSet result = null;
+        try {
```
mrigger, line 261: We could subsequently think whether we can extract such low-level result handling logic to a more high-level API.

src/sqlancer/common/oracle/EETDMLOracle.java
```diff
@@ -93,77 +103,192 @@ public void check() throws SQLException {
...
+            shown++;
+        }
+        return message.toString();
+    }
+
+    // Indexes a post-image by its row identifier (the first column of each row)
+    private static Map<String, List<String>> indexByRowId(List<List<String>> image) {
+        Map<String, List<String>> byRowId = new LinkedHashMap<>();
+        for (List<String> row : image) {
+            byRowId.put(row.get(0), row);
```
mrigger: could we have a named constant, similar to the existing `ROW_ID_COLUMN`? This would make it clearer what part of the code relies on this assumption.

src/sqlancer/common/oracle/EETDMLOracle.java
tlmorgan24: Yes, this would be cleaner, I will take note of it for a future PR

src/sqlancer/common/oracle/EETDMLOracle.java
tlmorgan24: Yes, `result.getString(i)` uses JDBC `ResultSet` indexing, which is 1-based

src/sqlancer/common/oracle/EETDMLOracle.java
tlmorgan24: Addressed in latest push. Instead of using an additional named constant, I ended up designing it so that everything is derived from the existing ROW_ID_COLUMN

src/sqlancer/common/oracle/EETDMLOracle.java
tlmorgan24: Addressed in latest push. To get around the two-return-values issue, I created a simple `StatementPair` class, which the methods return instances of

mrigger (approved):
LGTM, thanks!

PR #1359, opened 2026-08-17
Add reproducer to EET DML

tlmorgan24:
Previously, a reproducer was only supported for the EET SELECT oracle; this PR adds support to the DML oracle too. This is to enable test case reduction when the DML oracle is used. Further improvements to the reduction are also planned for an upcoming PR.


mrigger (approved):
LGTM, thanks!

PR #1360, opened 2026-08-18
Implement query reduction for EET SELECT

tlmorgan24:
The EET oracle detects a bug by transforming a query's expressions into semantically equivalent ones and comparing the two result sets. The transformed query is built by randomly applying EET rules, each wrapping a subexpression in a larger but equivalent form. Every such application is one "transformation site". Until now, reduction only shrank the database setup statements. The transformed query itself was left at full size, which is a problem because EET's transformation can lead to very complex queries which bury the real cause of a bug. This PR adds a second reduction stage that shrinks the transformed query, while keeping it equivalent to the original so the oracle stays sound.

To achieve this, the transformed query is re-rendered with only a chosen subset of the transformation sites applied, and this choice is varied using a delta-debugging strategy. This feature required quite a lot of infrastructure in order to keep track of the applied transformations, reproduce them in a deterministic way, and log the reduction process clearly.

Later, this functionality will be extended to the EET DML oracle.

### DIFF
--- a/src/sqlancer/Main.java
+++ b/src/sqlancer/Main.java
@@ -270,6 +270,11 @@ public final class Main {
             this.reduceBugInformation = bugInformation;
         }
 
+        // for reducers that rewrite the failing queries themselves (e.g., TransformationReducer)
+        public void updateReducedBugInformation(String bugInformation) {
+            this.reduceBugInformation = bugInformation;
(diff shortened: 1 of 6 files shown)


mrigger (approved):
Thanks a lot!

PR #1361, opened 2026-08-22
Add per-site simplification to EET transformed query reduction

tlmorgan24:
The previous PR reduced EET's transformed queries by removing transformation sites. This adds a further reduction stage which simplifies the remaining sites. For example, `CASE WHEN random_true_expression THEN expr ELSE random_expression END` may be simplified to `CASE WHEN TRUE THEN expr ELSE random_expression END`, if the bug is still present in that simplified case.

This is currently applied only to the EET SELECT oracle, but will later be extended to DML.


mrigger (approved):
LGTM, thanks!

PR #1362, opened 2026-09-13
Extend transformed query reduction to DML

tlmorgan24:
The previously merged transformed query reduction (PR #1360) was for the EET SELECT oracle only. This extends the work to also benefit the EET DML oracle; the theoretical approach is the same (implementing the unchanged TransformationReproducer interface). The main difference is that a DML statement's transformable expressions vary by statement kind (e.g. UPDATE has SET values while DELETE does not), so much of the diff is to make the reproducer hold them in a fixed order with a reassembly function that rebuilds the statement string.

Other parts of the diff ensure the EET DML logs clearly lay out the reduction. A pre-existing bug is also solved, where any oracle supplying a reproducer would only report its findings when --use-reducer was enabled.


tlmorgan24:
By solving the bug where an oracle would not report its findings, some pre-existing defects unrelated to my work were revealed in the CI tests. After fixing these too, the current PR is trying to accomplish too many things at once, so I am closing it and splitting it into two separate PRs.

PR #1363, opened 2026-09-13
Extend EET transformed query reduction to DML

tlmorgan24:
The previously merged transformed query reduction (PR #1360) was for the EET SELECT oracle only. This extends the work to also benefit the EET DML oracle; the theoretical approach is the same (implementing the unchanged TransformationReproducer interface). The main difference is that a DML statement's transformable expressions vary by statement kind (e.g. UPDATE has SET values while DELETE does not), so much of the diff is to make the reproducer hold them in a fixed order with a reassembly function that rebuilds the statement string. Other parts of the diff ensure the EET DML logs clearly lay out the reduction.


tlmorgan24:
Independently of my changes, Yugabyte seem to have released a new Docker image which introduces a bug, revealed by the Yugabyte CI because it dynamically fetches the ':latest' image. Further investigating this bug is out of scope for this PR, so in keeping with the approach for most other DBMSs, I have pinned the CI to a specific image instead

mrigger (approved):
LGTM, thanks!

PR #1364, opened 2026-09-14
Fix bug where some oracles were not reporting their findings

tlmorgan24:
This PR solves a bug where any oracle supplying a reproducer would only report its findings when --use-reducer was enabled. This bug has been present since 2022 (commit 625204e0), and I noticed it as I have been extending reproducer support to more oracles.

This meant the CI was not running correctly for some oracles (because the CI runs without --use-reducer). By solving this, the CI now runs correctly, which has surfaced some previously hidden defects. With the help of Claude, I then went on to fix those newly surfaced bugs for the affected DBMSs so the CI passes again:

- `HSQLDBExpressionGenerator.generateUnoptimizedQueryString` cleared the WHERE clause and projected `COUNT(*)` without re-projecting the predicate, making the unoptimised side count every row of the table. It now projects the predicate per row.
- `CockroachDBExpressionGenerator.generateOptimizedQueryString` projected a `COUNT(*)` column without using a GROUP BY, which returns a single row with the count. `countRows` was then called on this, which counts the number of rows, and therefore always counted one row instead of the actual underlying rows. `CockroachDBExpressionGenerator.generateOptimizedQueryString` now projects `*` instead of `COUNT(*)`, so `countRows` counts the actual rows.
- `YSQLExpressionGenerator.generateOptimizedQueryString` set the WHERE clause only in its non-aggregate branch, so the aggregate branch counted every row while the unoptimized side applied the predicate; the WHERE clause is now set for both. Finally, YSQLErrors was missing "character number must be positive" as an expected error (which is already listed for Postgres).


mrigger (approved):
LGTM! Sounds like quite a serious bug. Great that you have found and fixed it!

PR #1365, opened 2026-09-18
Add support for Apache Cloudberry

roseduan:
Apache Cloudberry is a PostgreSQL-derived MPP database, so the provider extends PostgresProvider and reuses PostgreSQL's expression generators and oracles. Cloudberry is a transparent MPP system: a single connection to the coordinator is enough, so there is no worker-node registration and no extension to create. createDatabase is inherited, wrapped only in a retry for an intermittent MPP-specific failure (see the comment on the override).

The Cloudberry-specific parts are the distribution policy, assigned per table via ALTER TABLE ... SET DISTRIBUTED with a distribution key chosen to be a subset of any PRIMARY KEY or UNIQUE constraint, and leaf partitions for the declaratively partitioned tables the inherited generator emits. Storage formats (heap, ao_row, ao_column, pax) are exercised through the inherited USING clause.

CloudberryCommon collects the errors Cloudberry returns for statements that are valid in PostgreSQL but violate an MPP restriction. CloudberryBugs follows the TiDBBugs pattern, gating each work-around for an unfixed target-DB bug behind a flag so it is easy to find and remove once the bug is fixed; each flag's comment records the scope that was actually verified against a live cluster.

Two inherited actions are not generated. CREATE STATISTICS can crash the coordinator, which a whitelisted error message cannot handle. CREATE VIEW builds its body with PostgresRandomQueryGenerator, which readily emits DISTINCT ON or LIMIT with no matching ORDER BY; such a view is re-evaluated per reference, and on an MPP cluster the row it yields can legitimately differ between evaluations, which breaks every oracle that compares two evaluations of the same data.

This PR covers database generation plus the three oracles that need no Cloudberry-specific oracle implementation: NoREC, TLPWhere and the fuzzer. The remaining oracles, and a CI job, follow in separate PRs.

Verified against a 3-segment cluster built from Cloudberry main (3.0.0-devel): mvn verify -DskipTests=true is clean, and NoREC, TLPWhere and the fuzzer each ran 4 threads x 140-200s across seeds 0-10 with no thread deaths and no oracle reports.


mrigger:
@roseduan, thanks a lot for the PR. This looks great! I'm currently on vacation until September 27. I'll try to take a look asap.

PR #1367, opened 2026-09-20
Add join support to EET DML

tlmorgan24:
This PR adds join support to EET DML, complementing the existing join support that the EET SELECT oracle has.

The added support is as follows:

- `DELETE` - joins fully supported
- `UPDATE` - joins supported, but the `SET` values are restricted to only the target table's columns. This is because some DBMSs update a target row once even when it matches several joined rows, taking the value from an unspecified one of them. A written value referencing a joined column could therefore differ between the two executions for reasons that are not a bug. Restricting the writes to the target table makes the choice of row deterministic.
- `INSERT .. SELECT` - remains single-table, as joining the `SELECT` source would allow one source row to produce several inserted rows, which the derived row identifier would no longer be able to tell apart.


mrigger (approved):
LGTM, thanks!

